PR #19, opened 2019-10-21
Checksummed addresses when creating light payment

rorolopetegui:
Partner Address was getting stored as bytes sometimes.
Now Partner and light client addresses are getting stored as checksummed addresses.


marcosmartinez7 (approved):
lgtm

PR #28, opened 2019-11-08
RLP-472 resolved

rorolopetegui:
When getting messages it was showing all from other clients. Filter to check if its the same light client added in SQL query.


marcosmartinez7 (approved):
utACK

PR #31, opened 2019-11-25
Parse identifiers and amounts in incoming Message

debrayperalta:
The LC SDK has a limitiation due to JS integer handling big values can lead to truncation, and therefore problems in the protocol.

This method parses according to the message the identifiers or amounts to int, because the LC SDK has to handle them as strings in order not to lose precision.


marcosmartinez7 (approved):
lgtm

PR #34, opened 2019-11-30
Returning apikey when sending onboarding (Already registered)

rorolopetegui:


marcosmartinez7 (dismissed):
Please modify the response object in order to have the same structure that the registration situation

marcosmartinez7 (approved):
lgtm

PR #36, opened 2019-12-01
Parse message amounts

debrayperalta:
This PR makes the message parser on LC payment messages to parse amount numbers, in order to manage bigger amounts from JS


marcosmartinez7 (approved):
lgtm

PR #37, opened 2019-12-01
Separated the V1 from Hub and Full Node

rorolopetegui:
-Separated URL endpoints for Hub, and Full mode
-Created array with common endpoints for both


marcosmartinez7 (approved):
lgtm

PR #49, opened 2019-12-26
RLP 520/error creating payments insufficient balance

rorolopetegui:
When making a light payment and having no balance, it was making an assert that was causing the hub  to fall. 
-Now it raises a InsufficientFounds Exceptions and shows a message.
-Checks the balance in channel before sending the LT, and compares it with the amount desired to be payed.


marcosmartinez7 (changes requested):
I think the assertion is ok since for the channel.py service is a precondition to have funds when invoking it. 

If the assertion fails then we have a bug, i think thats is the general idea with the assertions and we should keep it like that. 

In order to fix the issue we need to identify: 

- Why in normal payments the assertion always met? Is there any validation that we are skipping on light payments when invoking the channel.py service?

marcosmartinez7:
The `light-client ` branch was deleted and replaced by `develop`. 

Please move the PR to that branch.

PR #53, opened 2020-01-10
Rlp 582/fix test codebase

rorolopetegui:
-Fixing test codebase
-Fixed circular dependency in Node.py. In some cases using this class, it was getting a circular dependency


raiden/tests/unit/test_raiden_event_handler.py
```diff
@@ -20,9 +20,10 @@ def test_handle_contract_send_channelunlock_already_unlocked():
 
     Regression test for https://github.com/raiden-network/raiden/issues/3152
     """
-    channel_identifier = 1
+    channel_identifier = 2
```
marcosmartinez7: Is this identifier arbitrary?

raiden/tests/unit/transfer/test_channel.py
```diff
@@ -83,6 +83,7 @@ def make_unlock(unlock_end, partner_end):
         partner_onchain_locksroot=merkleroot(channel_state.partner_state.merkletree),
         block_number=1,
         block_hash=make_block_hash(),
+        participant1=""
```
marcosmartinez7: This does not make sense at all, a participant must be an hexadecimal address

raiden/tests/unit/transfer/test_channel.py
```diff
@@ -148,6 +149,7 @@ def make_unlock(unlock_end, partner_end):
...
```
marcosmartinez7: Must be a rsk address

raiden/tests/utils/factories.py
```diff
@@ -804,37 +804,89 @@ def __init__(
...
-        return self.channels[index].our_state.address
+        return self.get_sub_channel(index).our_state.address
 
     def partner_address(self, index: int) -> Address:
-        return self.channels[index].partner_state.address
+        return self.get_sub_channel(index).partner_state.address
 
-    def get_route(self, channel_index: int) -> RouteState:
-        return make_route_from_channel(self.channels[channel_index])
+    def get_route(self, key: str, channel_key: str) -> RouteState:
```
marcosmartinez7: The parameters names are confusing. The `key `parameter represents the address of the node, so it should be named `node_address `or something. The `channel_key` is the channel number, so `channel_identifier `would make more sense since is how is named on the node

raiden/tests/utils/factories.py
```diff
@@ -804,37 +804,89 @@ def __init__(
...
+        for key in self.channels.keys():
+            for channel_key in self.channels[key].keys():
+                if counter in args:
+                    ret_routes.append(self.get_route(key, channel_key))
+                    counter += 1
+                else:
+                    counter += 1
+        return ret_routes
+
+    def get_route_by_index(self, *args) -> RouteState:
```
marcosmartinez7: It does not makes sense to pass *args here, args is for variable arguments, and at this case you only need to receive an index, an int. 

Also, there are ways to avoid the nested loop, you can access to for channel in self.channels.values(): to iterate over the values instead of keys.

marcosmartinez7 (changes requested):

raiden/tests/unit/test_channelstate.py
```diff
@@ -1371,7 +1376,12 @@ def test_action_close_must_change_the_channel_state():
     channel_state = create_channel_from_models(our_model1, partner_model1, partner_key1)
 
     block_number = 10
-    state_change = ActionChannelClose(canonical_identifier=channel_state.canonical_identifier)
+    state_change = ActionChannelClose(
+        canonical_identifier=channel_state.canonical_identifier,
+        participant1=our_model1.participant_address,
+        participant2=partner_model1.participant_address,
+        signed_close_tx=""
```
marcosmartinez7, line 1383: I think this should be none on this case, since there is no signed close tx in a non-light channel

Can we just assign it as none or we must modifiy the core code?

raiden/tests/unit/test_channelstate.py
```diff
@@ -1405,7 +1415,12 @@ def test_update_must_be_called_if_close_lost_race():
     assert is_valid, msg
 
     block_number = 10
-    state_change = ActionChannelClose(canonical_identifier=channel_state.canonical_identifier)
+    state_change = ActionChannelClose(
+        canonical_identifier=channel_state.canonical_identifier,
+        participant1=our_model1.participant_address,
+        participant2=partner_model1.participant_address,
+        signed_close_tx=""
```
marcosmartinez7, line 1422: Same as previous comment

raiden/tests/unit/test_channelstate.py
```diff
@@ -1433,7 +1448,12 @@ def test_update_transfer():
     channel_state = create_channel_from_models(our_model1, partner_model1, partner_key1)
 
     block_number = 10
-    state_change = ActionChannelClose(canonical_identifier=channel_state.canonical_identifier)
+    state_change = ActionChannelClose(
+        canonical_identifier=channel_state.canonical_identifier,
+        participant1=our_model1.participant_address,
+        participant2=partner_model1.participant_address,
+        signed_close_tx=""
```
marcosmartinez7, line 1455: Same as previous comment

raiden/tests/utils/factories.py
```diff
@@ -804,37 +804,89 @@ def __init__(
...
+            for channel_key in self.channels[key].keys():
+                routes.append(self.get_route(key, channel_key))
+        return routes
+
+    def get_routes_by_index(self, indexes: list) -> List[RouteState]:
+        counter = 0
+        ret_routes = []
+        for key in self.channels.keys():
+            for channel_key in self.channels[key].keys():
+                if counter in indexes:
```
marcosmartinez7, line 842: Remove else clause and just increment the counter anyway.

marcosmartinez7 (changes requested):

rorolopetegui:
-We separate this PR in two different ones, one for Unit Tests and another for Integration:
PR Numbers are: 
Integration: #63 
Unit testing: #64

PR #55, opened 2020-01-15
Handle new channel return when duplicated was returning none instead of previous state

rorolopetegui:
The return statement was badly indented, it was breaking some tests for duplicated channels


marcosmartinez7 (approved):
Lgtm

PR #56, opened 2020-01-16
Fix for handle_transferrefundcancelroute was trying to get channel_st…

rorolopetegui:
This has changes for handle_transferrefundcancelroute function that was trying to get channel_state by channel_identifier, new structure of channels now needs to get by address first

### DIFF
--- a/raiden/tests/utils/factories.py
+++ b/raiden/tests/utils/factories.py
@@ -133,7 +133,7 @@ def make_uint256() -> int:
 
 
 def make_channel_identifier() -> ChannelID:
-    return ChannelID(make_uint256())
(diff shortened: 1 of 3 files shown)


raiden/tests/utils/factories.py
```diff
@@ -133,7 +133,7 @@ def make_uint256() -> int:
 
 
 def make_channel_identifier() -> ChannelID:
-    return ChannelID(make_uint256())
+    return ChannelID(make_uint64())
```
marcosmartinez7: I dont think is relevant but why did you change this?

marcosmartinez7 (changes requested):

PR #59, opened 2020-01-20
RLP-642/Fix-Refund-Transfers

rorolopetegui:
This fixes the refund transfers, to make a refund transfer you need to make the next scenario:

```
        10 <- 10 B  2 -> 10
A ->                                <- D
        10 <- 10 C  10 -> 10
```
The balances of the channels are in the arrows of B and C nodes.
First of all, when trying a mediated transfer between A and D, the first route that transfer will try is A-B-C, when B-C fails, it will generated a refund from B to A, and then a mediated from A-C-D will happen. This is what we're trying to fix.
We find the next issues when trying this transfer:
-Node B will send a refund to A, and A is not performing a re-route to C (Fixed, and a mediated between A-D is perfomed now)
-B sends a refund message (matrix), and when receiving a Processed message, isn't removing the Refund from queue. This will generate a Lockexpired message that will remain on the queue
-When performing the first issue, now fixed, you cant make another mediated with refund. Only the first time it'll work, then it has a problem with the nonce expected.

-Fixes refundtransfer and some tests that was related to.

Related issue in raiden [here](https://github.com/raiden-network/raiden/commit/1cc9eeccbdbb761cd68c6b8324a4c3a854516bb8). 

> The fix involves splitting existing ReceiveTransferRefundCancelRoute into two different
> state changes: `ReceiveTransferCancelRoute` and `ActionTransferReroute`. The first is
> always handled, while the re-routing is only done for transfers that are not atomic swaps.
> 
> The re-route handling now works by creating a copy of the existing initiator task (keyed
> by the new secrethash) which can then go through the rest of the flow, while the original
> route is left there until its lock is expired.


marcosmartinez7 (dismissed):
Lgtm

marcosmartinez7 (commented):
Lets wait to test refund transfers manually first

raiden/message_handler.py
```diff
@@ -28,12 +28,12 @@
     ReceiveSecretRequest,
     ReceiveSecretReveal,
     ReceiveTransferRefund,
-    ReceiveTransferRefundCancelRoute,
-    ReceiveSecretRequestLight, ReceiveSecretRevealLight)
+    ActionTransferReroute,
```
marcosmartinez7, line 29: Just to be clear, this event represents the refund received plus try another route? The side effects are the same as the old `ReceiveTransferRefundCancelRoute` ?

raiden/message_handler.py
```diff
@@ -141,10 +141,10 @@ def handle_message_lockexpired(raiden: RaidenService, message: LockExpired) -> N
 
     @staticmethod
     def handle_message_refundtransfer(raiden: RaidenService, message: RefundTransfer) -> None:
-        token_network_address = message.token_network_address
-        from_transfer = lockedtransfersigned_from_message(message)
         chain_state = views.state_from_raiden(raiden)
+        from_transfer = lockedtransfersigned_from_message(message)
```
marcosmartinez7: Is there any reason to move these lines?

raiden/raiden_event_handler.py
```diff
@@ -331,16 +331,16 @@ def handle_paymentsentfailed(
         payment_status = raiden.targets_to_identifiers_to_statuses[target].pop(
             payment_identifier, None
         )
-
-        handle_receive_events_with_payments(raiden.wal.storage,
-                                            payment_status.payment_hash_invoice,
-                                            'raiden.transfer.events.EventPaymentSentFailed',
-                                            payment_identifier)
+        if payment_status:
```
marcosmartinez7, line 336: Why is this needed? If payment status is not present then we have a bug here.

raiden/storage/serialization/cache.py
```diff
@@ -0,0 +1,67 @@
+from typing import Any, Callable, Dict
```
marcosmartinez7: What is the purpose of this module and why it was added?

marcosmartinez7:
General things to fix

- The title of the PR is not related with the changes itself. RLP-595 is not the refund transfer issue. 
- The PR cannot be merged, please resolve conflicts first.
- The LGTM tool cannot act if the conflicts are not resolved.

raiden/message_handler.py
marcosmartinez7: It would be nice to explain that on the PR description or comments, just to know the reasons behind the refactor.

marcosmartinez7: Lets point to the raiden related issue too

rorolopetegui: [Here's the issue fix in raiden](https://github.com/raiden-network/raiden/commit/1cc9eeccbdbb761cd68c6b8324a4c3a854516bb8) they sepparate the functionality in two parts, ActionTransferReroute and ReceiveTransferCancelRoute.
I'll edit the PR description to make a link to this fix

raiden/message_handler.py
rorolopetegui: Just to make the code more raiden like. So the code is more "human readable"/like to compare with raiden.

raiden/raiden_event_handler.py
rorolopetegui: handle_receive_events_with_payments is making use of payment_status. So the condition will always be true. 
In the case that payment_status if None the function handle_receive_events_with_payments will make the node to fall. In the case that we use to fix the refund, a LockedExpired is making the node fall because of this line.

raiden/storage/serialization/cache.py
rorolopetegui: Deleted in next commit. This was part of a major change that i tryied to do, but was not worth it.

raiden/raiden_event_handler.py
marcosmartinez7: I understand. I think we should fix this in a future. Basically we have a function that is invoked every time that a payment fail, and there are some situations where the `payment_status` exists and in others dont. We should split this into different functions.

JONAF2103: Agree we should do that

raiden/tests/utils/factories.py
```diff
@@ -1024,6 +1024,7 @@ def make_transfers_pair(
 
         pair = MediationPairState(received_transfer, lockedtransfer_event.recipient, sent_transfer)
         transfers_pairs.append(pair)
+        payer_index += 1
```
JONAF2103, line 1024: Is this necessary? you are inside a range in a for, i mean is this ok? I think this could jump and skip indexes, the for starts like this ```for payer_index in range(number_of_channels - 1):``` lets suppose that in the first loop payer_index = 0, then with this sentence you will put that payer_index = 1 before ending the loop 0, then the loop 1 will increase that value and instead of 1 you will have 2, is this possible or i'm just wrong?

marcosmartinez7: This is confusing for me. If the idea is to access to two contiguous positions, why not just use only one index? 

`channel_state=channels.get_sub_channel(payer_index +1)`

rorolopetegui: As you guys say, the idea of this make_transfers_pair is to get access to two contiguos sub_channels.
I changed this cause the other way that i was implementing was more clearly than this one.

marcosmartinez7 (approved):
Lgtm

JONAF2103 (approved):

marcosmartinez7 (changes requested):
Please fix the two lgtm warnings.

marcosmartinez7 (approved):

PR #60, opened 2020-01-20
Fix for handle_batch_unlock

rorolopetegui:
It was looking into channelidentifiers_to_channels only by channel_identifier, new structure needs node_address to get later the channel_identifier.


marcosmartinez7 (approved):
Lgtm

PR #61, opened 2020-01-20
Perform check when trying to get channel from channelidentifiers_to_channels

rorolopetegui:
A test was failing cause it was trying to get over a None return, "None.get(channel_identifier)"


raiden/transfer/mediated_transfer/initiator_manager.py
```diff
@@ -296,7 +296,9 @@ def handle_transferrefundcancelroute(
         return TransitionResult(payment_state, list())
 
     channel_identifier = initiator_state.channel_identifier
-    channel_state = channelidentifiers_to_channels.get(channel_identifier)
+    channel_state = channelidentifiers_to_channels.get(initiator_state.transfer.initiator)
+    if channel_state:
+        channel_state = channel_state.get(channel_identifier)
 
     if not channel_state:
```
marcosmartinez7: This should be a basic `else if ` statement.

marcosmartinez7 (changes requested):
Thre is still a if-else sentence to fix

marcosmartinez7 (approved):
Lgtm

PR #66, opened 2020-02-03
Change in function that was returning badly

rorolopetegui:
-Change for list_channelstate_for_tokennetwork that was returning a list of list channelstate instead of a single list channel_state itself


marcosmartinez7 (approved):
Lgtm

PR #68, opened 2020-02-05
When sending big amount values from LC it throws insert errors on SQLite

rorolopetegui:
-This fix a problem that was making Hub-Mode crashing when trying a payment between a LC and a Partner grater than 10 tokens, when trying to insert a tuple into SQLite.
-Now the Amount is a Text field, and gets parsed when the object Light_client_payment is instanciated


raiden/lightclient/models/light_client_payment.py
```diff
@@ -34,7 +34,7 @@ def __init__(
         self.partner_address = to_checksum_address(partner_address)
         self.is_lc_initiator = is_lc_initiator
         self.token_network_id = to_checksum_address(token_network_id)
-        self.amount = amount
+        self.amount = int(amount)
```
marcosmartinez7: Why is the amount changed to be received as `string` if it will be casted to `int ` and it was already received as an `int`?

marcosmartinez7 (changes requested):

raiden/lightclient/models/light_client_payment.py
marcosmartinez7: It does make no sense to receive as a string since all the instances of LightClientPayment receives an int as an amount. 

Regardless that the change does not make sense, if you change the type received on the constructor you need to modify all the instanciations to receive a string.

marcosmartinez7 (changes requested):

marcosmartinez7:
Please use a more representative PR title, the bug was not only with transfers of 10 tokens.

raiden/lightclient/models/light_client_payment.py
rorolopetegui: You're right, as we talk in the daily, i'll change this for a better operation in the code.

marcosmartinez7 (approved):
utACK

PR #70, opened 2020-02-07
Missing parameter in wait_for_settle

rorolopetegui:
This PR fixes a missed parameter (Partner_addresses) in wait_for_settle function. 
-This function is used in token tab when trying to "leave network" of a token.
Now it closes the channels as expected


marcosmartinez7 (approved):
utACK

PR #71, opened 2020-02-11
RLP-639/Contract without code

rorolopetegui:
This fixes when a Token has no code, maybe because the address runs out of gas, or similar problems that makes the token get deployed without any code.
-When trying to get the code of the token, it was returning 0x00. Added a condition for that returning code.
-Now handling exception AddressWithoutCode and gets logged into console.

### DIFF
--- a/raiden/blockchain_events_handler.py
+++ b/raiden/blockchain_events_handler.py
@@ -53,8 +53,6 @@ def handle_tokennetwork_new(raiden: "RaidenService", event: Event):
     token_network_address = args["token_network_address"]
     token_address = typing.TokenAddress(args["token_address"])
     block_hash = data["block_hash"]
-    # from raiden.network.rpc.client import check_address_has_code
-    # check_address_has_code(raiden.chain.client, token_network_address, 'TokenNetwork')
     try:
         token_network_proxy = raiden.chain.token_network(token_network_address)
         raiden.blockchain_events.add_token_network_listener(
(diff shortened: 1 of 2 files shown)


raiden/blockchain_events_handler.py
```diff
@@ -52,27 +53,30 @@ def handle_tokennetwork_new(raiden: "RaidenService", event: Event):
     token_network_address = args["token_network_address"]
     token_address = typing.TokenAddress(args["token_address"])
     block_hash = data["block_hash"]
+    # from raiden.network.rpc.client import check_address_has_code
+    # check_address_has_code(raiden.chain.client, token_network_address, 'TokenNetwork')
+    try:
```
marcosmartinez7, line 56: I dont understand the changes. 

The method that raises the `AddressWithoutCode` is `check_address_has_code` right? 

And that code is commented and is not inside the `try-catch `block

marcosmartinez7 (changes requested):

marcosmartinez7:
Please modify the title with the jira issue id

raiden/blockchain_events_handler.py
rorolopetegui: The function check_address_has_code is called when trying to get the token_network_proxy as a check... if it raises the AddressWithoutCode exception, the try-catch block will handle it, and not break the node

rorolopetegui: I'll delete the comments, was just for debugging

marcosmartinez7: Ok got it !

marcosmartinez7 (approved):
unACK

PR #72, opened 2020-02-12
Light payment with not enough balance

rorolopetegui:
This fixes an error that happens when trying to create a light payment with not enough balance. This was making the HUB to crash.
-Changed the logic of create_light_client_payment function, now it creates a LT object, and saves it, similar to normal payments.


marcosmartinez7 (changes requested):
What do you think about reestructuring this in order to do not have two code blocks that instantiate the LockedTransfer, both here and on the `create_sendlockedtransfer_event ` function?

rorolopetegui:
> What do you think about reestructuring this in order to do not have two code blocks that instantiate the LockedTransfer, both here and on the `create_sendlockedtransfer_event ` function?

Ok, i like it... Let me implement it.

marcosmartinez7 (approved):
unACK

PR #73, opened 2020-02-17
Set Total deposit change logic

rorolopetegui:
Get Channel total deposit before sending the new deposit. It was trying to deposit with the actual balance of the channel, and the balance includes payments. This was making wrong operations, and getting errors.


marcosmartinez7 (approved):
LGTM

marcosmartinez7 (commented):
LGTM but we need to rebuild the webapp probably according to https://github.com/rsksmart/lumino-web/pull/5

marcosmartinez7 (approved):
lgtm

PR #75, opened 2020-02-24
Rlp 665/apply pylint to the lumino node 1

JONAF2103:
Fixing warnings for pylint

* Fixing unused variables
* Fixing unused parameters
* Fixing unspecified parameters

### DIFF
--- a/raiden/network/proxies/secret_registry.py
+++ b/raiden/network/proxies/secret_registry.py
@@ -256,7 +256,7 @@ class SecretRegistry:
             # assert/revert, and the account should always be funded
             self.proxy.jsonrpc_client.check_for_insufficient_eth(
                 transaction_name="registerSecretBatch",
-                _address=self.node_address,
+                address=self.node_address,
                 transaction_executed=True,
                 required_gas=gas_limit,
                 block_identifier=checking_block,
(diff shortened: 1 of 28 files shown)


raiden/network/proxies/secret_registry.py
```diff
@@ -256,7 +256,7 @@ def register_secret_batch(self, secrets: List[Secret]):
             # assert/revert, and the account should always be funded
             self.proxy.jsonrpc_client.check_for_insufficient_eth(
                 transaction_name="registerSecretBatch",
-                address=self.node_address,
+                _address=self.node_address,
```
marcosmartinez7: We should be using the `_address` parameter instead of:

`our_address = to_checksum_address(self.address)`

The method uses the JSONRPC client address, but we should be able to pass an address parameter to check also for LC addresses.

raiden/transfer/views.py
```diff
@@ -571,7 +571,7 @@ def list_channelstate_for_tokennetwork_lumino(
     chain_state: ChainState,
     payment_network_id: PaymentNetworkID,
     token_addresses_split,
-    node_address
+    _node_address
```
marcosmartinez7: We should remove this parameter, its not used and it makes no sense since the method returns the tokens. Thats not related with a node address in particular.

marcosmartinez7 (changes requested):

marcosmartinez7 (approved):
utACK

PR #79, opened 2020-03-02
RLP-670/Fix-Refund-Transfers

rorolopetegui:
This fixes the refund transfers, to make a refund transfer you need to make the next scenario:
```
        10 <- 10 B  2 -> 10
A ->                                <- D
        10 <- 10 C  10 -> 10
```

The balances of the channels are in the arrows of B and C nodes. 
First of all, when trying a mediated transfer between A and D, the first route that transfer will try is A-B-C, when B-C fails, it will generated a refund from B to A, and then a mediated from A-C-D will happen. This is what we're trying to fix.
We find the next issues when trying this transfer:
-Node B will send a refund to A, and A is not performing a re-route to C (Fixed, and a mediated between A-D is perfomed now)
-B sends a refund message (matrix), and when receiving a Processed message, isn't removing the Refund from queue. This will generate a Lockexpired message that will remain on the queue
-When performing the first issue, now fixed, you cant make another mediated with refund. Only the first time it'll work, then it has a problem with the nonce expected.


rorolopetegui:
Moved to [PR 59](https://github.com/rsksmart/lumino/pull/59)

PR #78, opened 2020-02-27
RLP-406 Each client has a different password for each homeserver

JONAF2103:
* Adding a new field to the client table **current_server_name** to save the current logged in server
* Adding logic to delete a LC when the current server is not available anymore to force a new onboarding

PR #80, opened 2020-03-03
RLP-406 Each client has a different password for each homeserver

JONAF2103:
* This commit fixes the data integrity of the client table, now we can delete clients without having a foreign key error, this solves an error introduced by [this](https://github.com/rsksmart/lumino/pull/78) PR 

* Also fixes a log syntax problem


marcosmartinez7 (approved):
Lgtm

PR #81, opened 2020-03-09
RLP-574 Delete/Update old non closing balance proof

JONAF2103:
* Adding try catch to manage the way we update the non closing balance proof to delete it when it fails or it's successful

* Adding the delete non closing balance proof method to delete the balance proof by channel id

* Checking if the balance proof exists on the db before sending that again if it's in the state of the event

### DIFF
--- a/raiden/network/proxies/token_network.py
+++ b/raiden/network/proxies/token_network.py
@@ -1742,7 +1742,7 @@ class TokenNetwork:
         else:
             log.info("updateNonClosingBalanceProof successful", **log_details)
         finally:
-            # if an error occurred we need to delete the non closing balance proof from the database
+            # we need to delete the non closing balance proof from the database to avoid re-send it
             log.info("Deleting non closing balance proof for channel id " + str(channel_identifier), **log_details)
             raiden.wal.storage.delete_light_client_non_closing_balance_proof(channel_id=channel_identifier)
 
--- a/raiden/network/proxies/token_network.py
+++ b/raiden/network/proxies/token_network.py
@@ -1437,6 +1437,7 @@ class TokenNetwork:
         closing_signature: Signature,
         non_closing_signature: Signature,
         given_block_identifier: BlockSpecification,
+        raiden: "RaidenService"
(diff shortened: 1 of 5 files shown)


raiden/network/proxies/token_network.py
```diff
@@ -1566,175 +1567,184 @@ def update_transfer_light(
...
-            raise RaidenUnrecoverableError("update_transfer failed for an unknown reason")
+                if not partner_details.is_closer:
+                    raise RaidenUnrecoverableError(
+                        "update_transfer cannot be sent if the partner did not close the channel"
+                    )
 
-        log.info("updateNonClosingBalanceProof successful", **log_details)
+                raise RaidenUnrecoverableError("update_transfer failed for an unknown reason")
+        except Exception:
+            log.info("An error occurred trying to update the non closing balance proof for channel id " +
```
marcosmartinez7, line 1739: Why dont we just catch the `RaidenUnrecoverableError `and `RaidenRecoverableError` here? 

I think is more natural to make other errors just fail at runtime, any other exception that is not one of those two should not be an expected exception

marcosmartinez7 (changes requested):

raiden/network/proxies/token_network.py
JONAF2103: Because the first call is to the smart contract, if anything else happens that throws an error then you will need to repeat the action, so suppose that everything is fine with the smart contract call but later something fails, then you will force to call the smart contract again, so i wanted to avoid that but it's your call, if you want we can only catch those kind of errors

marcosmartinez7: Thats a good reason.

Still seems too wide for me to catch any exception, but since we dont want to lose gas innecesary and i dont have any better option in mind right now, we can go with this implementation.

JONAF2103: Ok cool, i mean i was thinking on the gas basically so i can't think in another way to avoid spending gas in this case.

PR #83, opened 2020-03-16
RLP-656/Added default value for signed_tx

rorolopetegui:
This fixes a bug that was happen when trying to close a channel, it was sending a signed_tx attr and it was not needed. Only needed for LC.

Fix: Assigned default value for signed_tx in ActionChannelClose, when it is a non-lightclient.


marcosmartinez7 (approved):
utACK

PR #84, opened 2020-03-19
Feature/RLP-688/Mediated-transfer-LC

rorolopetegui:
Fixes mediated transfer between a LC and a Lumino node.
Errors, bugs and features that was not letting the HUB make the mediated transfer, here's the description:

- LC was not able to make a mediated transfer
**Problem**: At the creation of the locked transfer message (First message of the protocol), it was checking if the LC had a channel with the given partner
**Solution**: Now if it can get the channel with the partner, it gets the first channel and tries a mediated transfer
- LC was not able to receive a mediated transfer
**Problem**: Some messages of the protocol was trying to be sended to wrong addresses
**Solution**: Changed variables that was confusing the addresses, and sending the message wrongly.

### DIFF
--- a/raiden/api/python.py
+++ b/raiden/api/python.py
@@ -1696,6 +1696,7 @@ class RaidenAPI:
         amount: typing.TokenAmount,
         secrethash: typing.SecretHash
     ) -> HubResponseMessage:
+
         channel_state = views.get_channelstate_for(
             views.state_from_raiden(self.raiden),
             registry_address,
@@ -1703,6 +1704,13 @@ class RaidenAPI:
             creator_address,
             partner_address,
         )
+        if not channel_state:
+            channel_state = views.get_channelstate_for(
+                views.state_from_raiden(self.raiden),
+                registry_address,
+                token_address,
+                creator_address,
+            )
         if channel_state:
             chain_state = views.state_from_raiden(self.raiden)
 
--- a/raiden/api/python.py
+++ b/raiden/api/python.py
@@ -1704,13 +1704,15 @@ class RaidenAPI:
             creator_address,
             partner_address,
         )
+        #If we dont have a channel with the partner we need to get a channel to try a mediated transfer
         if not channel_state:
-            channel_state = views.get_channelstate_for(
+            channel_state = views.get_first_channelstate(
                 views.state_from_raiden(self.raiden),
                 registry_address,
                 token_address,
                 creator_address,
             )
+
         if channel_state:
             chain_state = views.state_from_raiden(self.raiden)
 
--- a/raiden/api/python.py
+++ b/raiden/api/python.py
@@ -1750,7 +1750,7 @@ class RaidenAPI:
                                                     message=locked_transfer, is_signed=False)
             return HubResponseMessage(lcpm_id, LightClientProtocolMessageType.PaymentSuccessful, payment_hub_message)
         else:
-            raise ChannelNotFound("Channel with given partner address doesnt exists")
+            raise ChannelNotFound("Your light client has 0 channels oppened")
 
     def validate_light_client(self, api_key: str):
         """
--- a/raiden/api/python.py
+++ b/raiden/api/python.py
@@ -16,7 +16,7 @@ from eth_utils import is_binary_address, to_checksum_address, to_canonical_addre
 from ecies import encrypt
 
 import raiden.blockchain.events as blockchain_events
-from raiden import waiting
+from raiden import waiting, routing
 from raiden.api.validations.api_error_builder import ApiErrorBuilder
 from raiden.api.validations.channel_validator import ChannelValidator
 from raiden.constants import (
@@ -95,7 +95,7 @@ from raiden.utils.typing import (
     TokenNetworkAddress,
     TokenNetworkID,
     Tuple,
-    SignedTransaction)
+    SignedTransaction, InitiatorAddress)
 
 from raiden.rns_constants import RNS_ADDRESS_ZERO
 from raiden.utils.rns import is_rns_address
@@ -1706,12 +1706,28 @@ class RaidenAPI:
         )
         #If we dont have a channel with the partner we need to get a channel to try a mediated transfer
         if not channel_state:
-            channel_state = views.get_first_channelstate(
-                views.state_from_raiden(self.raiden),
-                registry_address,
-                token_address,
-                creator_address,
+            token_network_id = views.get_token_network_by_token_address(
+                views.state_from_raiden(self.raiden), registry_address, token_address
             )
+            routes, _ = routing.get_best_routes(
+                chain_state=views.state_from_raiden(self.raiden),
+                token_network_id=token_network_id.address,
+                one_to_n_address=self.raiden.default_one_to_n_address,
+                from_address=InitiatorAddress(creator_address),
+                to_address=partner_address,
+                amount=amount,
+                previous_address=None,
+                config=self.raiden.config,
+                privkey=self.raiden.privkey,
+            )
+            if len(routes) > 0:
+                channel_state = views.get_channelstate_for(
+                    views.state_from_raiden(self.raiden),
+                    registry_address,
+                    token_address,
+                    creator_address,
+                    routes[0].node_address,
+                )
 
         if channel_state:
             chain_state = views.state_from_raiden(self.raiden)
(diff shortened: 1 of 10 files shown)


raiden/api/python.py
```diff
@@ -1696,13 +1696,21 @@ def create_light_client_payment(
...
     ) -> HubResponseMessage:
+
         channel_state = views.get_channelstate_for(
             views.state_from_raiden(self.raiden),
             registry_address,
             token_address,
             creator_address,
             partner_address,
         )
+        if not channel_state:
```
marcosmartinez7, line 1708: Could you please explain why is this necessary?

raiden/transfer/node.py
```diff
@@ -951,7 +951,7 @@ def handle_state_change(
         iteration = handle_init_target(chain_state, state_change, storage, chain_state.our_address)
     elif type(state_change) == ActionInitTargetLight:
         assert isinstance(state_change, ActionInitTargetLight), MYPY_ANNOTATION
-        iteration = handle_init_target(chain_state, state_change, storage, state_change.transfer.initiator)
+        iteration = handle_init_target(chain_state, state_change, storage, state_change.transfer.target)
```
marcosmartinez7, line 965: This is at least confusing, but i suspect that it will break light payments reception:

- `ActionInitTargetLight ` is the state that is generated on a light client payment reception on a single hop payment. That is a feature that is currently working and tested. Making this change could affect that.
- `handle_init_target `receives an initiator address, so at least is the parameter name confusing.
- `handle_init_target` invokes `subdispatch_targettask` by `initiator address`, so its also confusing.

raiden/transfer/views.py
```diff
@@ -292,20 +292,32 @@ def get_channelstate_for(
     if token_network and creator_address in token_network.channelidentifiers_to_channels or \
         token_network and partner_address in token_network.channelidentifiers_to_channels:
         channels = []
-        for channel_id in token_network.partneraddresses_to_channelidentifiers[partner_address]:
+        if partner_address:
```
marcosmartinez7: Why we need to do this differentiation?

marcosmartinez7 (changes requested):

raiden/api/python.py
```diff
@@ -1696,13 +1696,21 @@ def create_light_client_payment(
...
+            channel_state = views.get_channelstate_for(
+                views.state_from_raiden(self.raiden),
+                registry_address,
+                token_address,
+                creator_address,
+            )
         if channel_state:
```
JONAF2103, line 1732: I think this could be redundant. If you see you already checked if channel state hasn't a value before and if it doesn't then you assign a value so it should have a value at this point, maybe that "if" is not necessary.

raiden/api/python.py
rorolopetegui: In the first "try", he'll check if he has a channel with the partner, that if is necessary for mediated transfers. It will return the first channel that the node has, and try mediate the payment.

raiden/transfer/views.py
rorolopetegui: The normal behavior of this function is to search by partner_address (parameter in the function), i adapt this function to return a channel_state if you dont send the partner_address, so it'll return the first channel for the node. That "if" is necessary so we can make the  search correctly.

marcosmartinez7: The responsability of this function is (or should be) "find the channel between `creator `and `partner`"

Having said that, It makes no sense for me to have optional parameters.

On this part of the code:


```
 if creator_address in token_network.channelidentifiers_to_channels:
          channel = token_network.channelidentifiers_to_channels[creator_address].get(channel_id)

 if channel is None and partner_address in token_network.channelidentifiers_to_channels:
        # Check if partner address had a open channel, can be a hub node.
          channel = token_network.channelidentifiers_to_channels[partner_address].get(channel_id)
           address_to_get_channel_state = partner_address
```


there are present the two scenarios, or its a hub channel (creator is the hub node) or its a light channel (creator is not the hub, basically is not the raiden.address)

i think the comment should said:

`                # Check if partner address had a open channel, can be a light client channel.`

Returning to the discussion, why you need to modify the responsability of this function in order to return the first channel for a creator_address?

If the idea is "i need to get the first channel to try the first route for a mediated transfer" some things comes to my mind

1. We must not overload a function to support a particular case
2. We can create another function to do that, `get_first_possible_route`
3. Why are we reimplementing this? isnt just the behaviour of the `try_new_route ` method?

raiden/api/python.py
marcosmartinez7: I think thats related with the modification of that method. But, i dont see a need for a "retry" on a method invocation. The method should be invoked with the right parameters and it should handle the situation.

marcosmartinez7 (changes requested):

JONAF2103 (changes requested):

raiden/transfer/node.py
rorolopetegui: As we spoke yesterday for this case. 
In the ActionInitTarget, the handle_init_target function gets chain_state.our_address that corresponds as it says the node itself address. This address is used to get a channel_state, from here we can see that in Light case, we get the initiator address.
The function that is using this addresses is get_channelstate_by_canonical_identifier_and_address, and if you send an address from a partner it will first try to get the LC address with that address and a channel identifier, and then take a look at channelidentifiers_to_channels with the lc_address obtained. We take the decision of letting this change so the two cases ActionInitTargetLight and ActionInitTarget are going to behave the same, we can send the initiator as it was, but this will cause confuse situations like this.

marcosmartinez7: Agree.

We still have to modify the mentioned methods to receive a more representative parameter name for the address instead of `initiator`, but i agree with the change. 

Also i want to make mention that since we know the `channel_id ` and a member of the channel (therefore we can use `get_channelstate_by_canonical_identifier_and_address` ), using `initiator `or `target `has the same effect.

raiden/api/python.py
rorolopetegui: Please re-review this, i created a function to get the first channel_state in case that channel_state is none

raiden/transfer/views.py
rorolopetegui: Please re-review this, i created a function to get the first channel_state in case that channel_state is none

raiden/api/python.py
rorolopetegui: Please re-review this, i created a function to get the first channel_state in case that channel_state is none

raiden/api/python.py
```diff
@@ -1696,13 +1696,39 @@ def create_light_client_payment(
...
+                config=self.raiden.config,
+                privkey=self.raiden.privkey,
+            )
+            if len(routes) > 0:
+                channel_state = views.get_channelstate_for(
+                    views.state_from_raiden(self.raiden),
+                    registry_address,
+                    token_address,
+                    creator_address,
+                    routes[0].node_address,
```
marcosmartinez7, line 1729: Here we are returning the first route to the light client, to make it signing the transaction with that `channel_identifier`. I have a question about that: what happens if the routes are different when sending the signed locked transfer? 

When the LC sends the signed LT, the `ActionInitInitiatorLight ` state change is processed, and that state change receives routes, that are being calculated again at that point.

Basically, `initiate_payment ` triggers again the route calculation

raiden/api/python.py
```diff
@@ -1740,7 +1766,7 @@ def create_light_client_payment(
                                                     message=locked_transfer, is_signed=False)
             return HubResponseMessage(lcpm_id, LightClientProtocolMessageType.PaymentSuccessful, payment_hub_message)
         else:
-            raise ChannelNotFound("Channel with given partner address doesnt exists")
+            raise ChannelNotFound("Your light client has 0 channels oppened")
```
marcosmartinez7, line 1769: This is not true after this routing modification.

Node A channels: B,C,D,E. Node E and D have a channel with F.

Node A sends a payment to node F, the first try to get a direct channel will fail, the routing one too (because there is no balance to route from A to F i.e). Therefore channel_state will be none, but the light client has B,C,D,E channels, not zero.

marcosmartinez7 (changes requested):

marcosmartinez7:
I will merge this into a collaborator draft pr

PR #85, opened 2020-03-19
feature/RLP-690/add-routing-mode

marcosmartinez7:
**Description**

This is the first PR related with mediation fees.  It includes all the code related with with the `routing-mode` but **only for internal routing**

Any flag that is not exctrictly related with the `private` routing mode, where nodes does not interact with the Path Finding Services, will not be included on this PR.

### DIFF
--- a/raiden/app.py
+++ b/raiden/app.py
@@ -1,7 +1,7 @@
 import structlog
 from eth_utils import to_checksum_address
 
-from raiden.constants import DISCOVERY_DEFAULT_ROOM, PATH_FINDING_BROADCASTING_ROOM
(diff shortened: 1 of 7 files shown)


rorolopetegui:
Can you update the title of the PR with the issue number? So that way all the PRs have the same convention that we spoken before.
Maybe you can explain in this PR a little more about Routing mode (In this case internal), or just putting a link to understand what this is fixing

rorolopetegui (changes requested):
Can you update the title of the PR with the issue number? So that way all the PRs have the same convention that we spoken before.
Maybe you can explain in this PR a little more about Routing mode (In this case internal), or just putting a link to understand what this is fixing

rorolopetegui (approved):
uACK

PR #87, opened 2020-03-23
Feature/internal-routing-mediation-fees

marcosmartinez7:
**Description**

Includes the features for the first mediation fees MVP: mediation fees with internal routing for normal nodes (excluding light client payments)

**Task list for completion:** 

- Update lumino contracts for pathfinding services support.
- Modify payment method for normal nodes to include fees on LT 
- Modify the rest of the messages that involves fees amounts and the messages handlers (full payment protocol should work) 
- Modify internal routing to set estimated fees
- Update CLI only with the internal routing flags 
- Add the related code with those flags
- Update capacity on deposit
- Split fees into channel in and channel out
- Create test cases for fees

### DIFF
--- a/raiden/ui/cli.py
+++ b/raiden/ui/cli.py
@@ -233,11 +233,12 @@ def options(func):
                 "--routing-mode",
                 help=(
                     "Specify the routing mode to be used.\n"
-                    '"basic": use local routing\n'
                     '"pfs": use the path finding service\n'
+                    '"local": use local routing, but send updates to the PFS\n'
+                    '"private": use local routing and don\'t send updates to the PFS\n'
                 ),
                 type=EnumChoiceType(RoutingMode),
-                default=RoutingMode.BASIC.value,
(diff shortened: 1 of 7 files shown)


marcosmartinez7:
> What's the default value for this? since this is an optional value should we have a default value or in this case is not necessary? also a good practice is to have the optional values at the end of the param list.

Is not neccesary on this case since that it would be an error to start up the node without a `SecretRegistry ` address

raiden/ui/cli.py
```diff
@@ -632,6 +633,7 @@ def print_step(description: str, error: bool = False) -> None:
         port = next(free_port_generator)
 
         args["api_address"] = "localhost:" + str(port)
+        args["routing_mode"] = RoutingMode.PRIVATE
```
rorolopetegui, line 636: This is for testing purposes only? Or we're setting this value always in Private for some reason?

PR #88, opened 2020-03-27
bugfix/RLP-693/Add-function-for-lockexpired

rorolopetegui:
This fixes when a node that is a receiver from a payment and that payment expires, it was not handling that event. 
**Solution**: handle_lock_expired function was missed, searched in the git history and re-added now.


marcosmartinez7 (approved):
utACK

JONAF2103 (approved):
Looks good to me

PR #89, opened 2020-03-27
Bugfix/rlp 680/lc capacity

JONAF2103:
This PR is to fix a bug that we have for LC's connections.

* With this we fix the total amount of LC's that can be connected at the same time on a HUB node to make operations.


marcosmartinez7 (approved):
LGTM

rorolopetegui (approved):
lgtm

PR #94, opened 2020-05-27
Fix rpc, settlement integration tests

marcosmartinez7:
This PR fixes some test for: 

- raiden/tests/integration/api/test_pythonapi_regression.py
- raiden/tests/integration/long_running/test_settlement.py
- raiden/tests/integration/long_running/test_token_networks.py
- all the rpc integration tests
- raiden/tests/integration/test_balance_proof_check.py
- raiden/tests/unit/test_tokennetwork.py

Major modifications:
- Updated sol version for tests contracts
- Modified unlock event to get the right parameter from event data


raiden/blockchain_events_handler.py
```diff
@@ -475,7 +475,7 @@ def handle_channel_batch_unlock(raiden: "RaidenService", event: Event):
         transaction_hash=transaction_hash,
         canonical_identifier=canonical_identifier,
         participant=args["participant"],
-        partner=args["partner"],
+        partner=partner,
```
rorolopetegui: Partner here can be 2 things,  depending on which participant has the raiden.address, why this was using args["partner"] instead of this var?
Maybe the args["partner"] contemplates some other use cases that we're not seeing here.
I'll not mark this as a change for this PR cause maybe it's ok this change, but i only want to know why this change is needed.

raiden/tests/integration/long_running/test_settlement.py
```diff
@@ -76,33 +78,30 @@ def run_test_settle_is_automatically_called(raiden_network, token_addresses):
...
         app0.raiden.alarm.sleep_time,
-        app1.raiden.address
+        app0.raiden.address
     )
 
-    channel_state = views.get_channelstate_for_close_channel(
-        views.state_from_raiden(app0.raiden), registry_address, token_address, app0.raiden.address, app1.raiden.address,
-        channel_identifier
+    channel_state = views.get_channelstate_for(
+        views.state_from_raiden(app0.raiden), registry_address, token_address, app1.raiden.address
```
rorolopetegui, line 93: Can we re'review this? This function needs more arguments: 
```
def get_channelstate_for_close_channel(
    chain_state: ChainState,
    token_address: TokenAddress,
    creator_address: Address,
    partner_address: Address,
    channel_id_to_check: ChannelID = None
) -> Optional[NettingChannelState]:
```

raiden/tests/integration/long_running/test_settlement.py
```diff
@@ -351,78 +342,64 @@ def run_test_batch_unlock(
...
     timeout = 30 if blockchain_type == "parity" else 10
-    with gevent.Timeout(seconds=30, exception=AssertionError(msg)):
-        # Wait for both nodes (Bob and Alice) to see the on-chain unlock
-        wait_for_batch_unlock(
-            app=alice_app,
-            token_network_id=token_network_identifier,
-            participant=bob_address,
-            partner=alice_address,
-        )
+    with gevent.Timeout(timeout):
```
rorolopetegui, line 394: Nice Catch here

raiden/tests/utils/smartcontracts.py
```diff
@@ -116,18 +115,13 @@ def get_test_contract(name):
...
-    contract_path = os.path.abspath(
-        os.path.join(os.path.dirname(__file__), "..", "smart_contracts", f"{name}.sol")
-    )
-    contracts = compile_files_cwd([contract_path])
-    contract_key = os.path.basename(contract_path) + ":" + name
-
-    contract_proxy, receipt = deploy_client.deploy_single_contract(
-        contract_name=name, contract=contracts[contract_key]
+def deploy_rpc_test_contract(deploy_client, name):
+    contract_path, contracts = get_test_contract(f"{name}.sol")
```
rorolopetegui, line 119: This is good! More readable

raiden/tests/utils/smartcontracts.py
```diff
@@ -116,18 +115,13 @@ def get_test_contract(name):
...
+    contract_proxy, _ = deploy_client.deploy_solidity_contract(
+        name, contracts, libraries=dict(), constructor_parameters=None, contract_path=contract_path
     )
 
-    return contract_proxy, receipt
+    return contract_proxy
```
rorolopetegui, line 124: Just to take note, this receipt is not used in any other places?

rorolopetegui (changes requested):
Lgtm, just some minor changes needed

rorolopetegui:
Can we add a description to this PR? Some thing like: "Tests of settlement was not working", or add a little desc to this?

raiden/blockchain_events_handler.py
marcosmartinez7: Youre right, and since we need to implement the unlock for lc, i will rollback this change and check the tests results.

raiden/tests/utils/smartcontracts.py
marcosmartinez7: No, because the rest of the code does not assume this function returns a tuple. The addition of receipt was a work in progress that we didnt continue with.

raiden/tests/integration/long_running/test_settlement.py
marcosmartinez7: This function is get_channel_state_for, the only missing param is the partner address: 

```
def get_channelstate_for(
    chain_state: ChainState,
    token_address: TokenAddress,
    creator_address: Address = None,
    partner_address: Address = None
) -> Optional[NettingChannelState]:
```

and in this case is not needed, since the app1 address is the creator of the channel.

rorolopetegui: Ups! You're right! Mb here!

rorolopetegui (approved):
lgtm

PR #97, opened 2020-07-07
Bugfix/rlp 733/deposit light invalid previous deposit

marcosmartinez7:
This PR resolves the following bug:

- Partner or light client opens a channel
- The partner deposits on the channel, 0.1 tokens for example
- The light client now deposit some tokens, the hub returns that it must be greater than the previous total deposit (0.1)

The solution:

Since when the payment channel proxy is created, we cannot know if one of the participants is a light client or not (it can be both also), we swap the participants of the payment channel in order to set `participant1=light client address` always.

This way, the deposit preconditions, that are tested against `participant1 ` channel state, are always done against the light client side when it request a new deposit

### DIFF
--- a/raiden/api/python.py
+++ b/raiden/api/python.py
@@ -633,6 +633,8 @@ class RaidenAPI:
             signed_approval_tx=signed_approval_tx,
             signed_deposit_tx=signed_deposit_tx
         )
+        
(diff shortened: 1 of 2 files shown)


JONAF2103 (approved):
LGTM

PR #103, opened 2020-08-17
Bugfix/rlp-700/mediated-transfer-invoicing

mauricioirace:
Changes:
- Inject storage into mediated transfer


.gitignore
```diff
@@ -94,3 +94,5 @@ keystores
 
 # raiden logs
 raiden/raiden-*
+createEnv
```
marcosmartinez7: I dont think this scripts are part of the project.

raiden/api/webui/static/endpointConfig.js
```diff
@@ -1,5 +1,5 @@
-const backendUrl='http://localhost:5000'; 
-const nodeAddress = '0xa691a629E473D3dd19C24fe97F6b22A9Ee48F873'; 
+const backendUrl='http://localhost:5001'; 
+const nodeAddress = '0x4693688986f02569E2b4842e4D676C46a86c0388';
```
marcosmartinez7: This file must be on the gitignore

marcosmartinez7 (changes requested):
Some changes requested, also, you have commited a log file

raiden/api/webui/static/endpointConfig.js
mauricioirace: I agree, the problem is that when you try to start the node without that file, it crashes (It updates the file, but it doesn't create it if it doesn exist).

In fact, It seeme it happens here (/raiden/api/rest.py):

```
def _set_ui_endpoint(self):
        # Overrides the backend url in the ui bundle
        with open(self.flask_app.root_path + '/webui/static/endpointConfig.js') as f:
            lines = f.readlines()
```

I think changing the mode from 'r' to something else (maybe 'w+') in order to generate the file or replace it from scratch, would allow us to ignore the file without breaking anything

marcosmartinez7 (approved):
utack

PR #107, opened 2020-09-02
RLP-755 Refactor LC messages

JONAF2103:
In this PR we are:

* Removing light_client_payment -> client db relationship
* Adding light_client_protocol_message -> client db relationship
* Changing all calls that use light_client_payment -> client
* Modifiying StoreMessageEvent to store light client address and changing all the calls to fill that value correctly

This affects the payments flow and LC messaging logic

### DIFF
--- a/raiden/api/python.py
+++ b/raiden/api/python.py
@@ -1715,7 +1715,7 @@ class RaidenAPI:
 
             # Create the light_client_payment
             is_lc_initiator = 1
-            payment = LightClientPayment(creator_address, partner_address,
+            payment = LightClientPayment(partner_address,
                                          is_lc_initiator, channel_state.token_network_identifier,
                                          amount,
                                          str(date.today()),
--- a/raiden/api/python.py
+++ b/raiden/api/python.py
@@ -1729,6 +1729,7 @@ class RaidenAPI:
                 locked_transfer,
                 False,
                 payment.payment_id,
+                creator_address,
                 order,
                 LightClientProtocolMessageType.PaymentSuccessful,
                 self.raiden.wal
--- a/raiden/lightclient/handlers/light_client_message_handler.py
+++ b/raiden/lightclient/handlers/light_client_message_handler.py
@@ -34,6 +34,7 @@ def build_light_client_protocol_message(identifier: int, message: Message, signe
         message_type,
         unsigned_msg,
         signed_msg,
+        None,
         light_client_address
     )
 
--- a/raiden/lightclient/handlers/light_client_message_handler.py
+++ b/raiden/lightclient/handlers/light_client_message_handler.py
@@ -13,11 +13,13 @@ from raiden.messages import Message, LockedTransfer, SecretRequest, RevealSecret
     LockExpired
 from raiden.storage.sqlite import SerializedSQLiteStorage
 from raiden.storage.wal import WriteAheadLog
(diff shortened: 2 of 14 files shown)


raiden/api/python.py
```diff
@@ -1715,7 +1715,7 @@ def create_light_client_payment(
 
             # Create the light_client_payment
             is_lc_initiator = 1
-            payment = LightClientPayment(creator_address, partner_address,
+            payment = LightClientPayment(partner_address,
```
marcosmartinez7, line 1718: Why the `creator_address `param was removed?

JONAF2103: Because in the database it was directly mapped to the LC address and we are removing that from the table so it's not needed anymore.

raiden/lightclient/handlers/light_client_message_handler.py
```diff
@@ -43,10 +45,11 @@ class LightClientMessageHandler:
     def store_light_client_protocol_message(cls, identifier: int, message: Message, signed: bool, payment_id: int,
                                             order: int, message_type: LightClientProtocolMessageType,
                                             wal: WriteAheadLog):
+        payment = wal.storage.get_light_client_payment(payment_id)
         return wal.storage.write_light_client_protocol_message(
             message,
             build_light_client_protocol_message(identifier, message, signed,
-                                                payment_id, order, message_type)
+                                                payment_id, order, message_type, payment.light_client_address)
```
mortelli: it was my understanding that we could have LC messages without a payment involved.

so if we don't have a payment, how do we set the `light_client_address` in cases other than this one?

mauricioirace: I have the same question.

Also, wouldn't it fail in such case? (because payment is None)

JONAF2103: Yes but the query for payments is using a JOIN to get the information about the client, we don't want to remove that since we don't know all the places where it's being used so i keep the light_client_address inside the query for payements to keep it working and to keep the simplicity when you query a payment to get the client who did that payment, so basically with the payment id you can get the client.

JONAF2103: Long story short, no it will never be None since if you have a payment id you have a client who did that payment and the query for the payment will provide that information.

mauricioirace: I understand.

But what do we do when we need to store a message that is not directly associated with a payment (e.g. settlement, register_secret)?

JONAF2103: You put None on that field since it's optional, the relationship is like this:

Payment -> Light Client Protocol Message -> Client

In this case i use the relation between Payment and Light Client Protocol Message to get the Client. If you have a Payment you always have messages but not always that you have a message it's for a payment.

JONAF2103: You were right about the error, a fix it's already implemented.

mauricioirace (approved):
utACK

marcosmartinez7 (approved):
Lgtm in general.

Not 100% sure it will work on mediated payments. 

The construction of `StoreMessageEvent` receives the lc address from diferent attributes, that on mediated transfers maybe are not ok.

I will test it after merge with mediated light branch

PR #108, opened 2020-09-03
RLP-747 Fixing lumino empty token list crash

JONAF2103:
This PR fixes the crash on lumino node when the chain doesn't have any token deployed.

### DIFF
--- a/raiden/api/python.py
+++ b/raiden/api/python.py
@@ -916,8 +916,7 @@ class RaidenAPI:
         if registry_address and not is_binary_address(registry_address):
             raise InvalidAddress('Expected binary address format for registry in get_channel_list')
 
-        token_addresses_split = token_addresses.split(",") \
-            if token_addresses is not None and token_addresses.__len__() > 0 else list()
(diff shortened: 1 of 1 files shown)


marcosmartinez7 (approved):

mauricioirace (approved):
LGTM

raiden/api/python.py
```diff
@@ -916,7 +916,8 @@ def get_channel_list_for_tokens(
         if registry_address and not is_binary_address(registry_address):
             raise InvalidAddress('Expected binary address format for registry in get_channel_list')
 
-        token_addresses_split = token_addresses.split(",")
+        token_addresses_split = token_addresses.split(",") \
+            if token_addresses is not None and token_addresses.__len__() > 0 else list()
```
mortelli: a couple of things here:
1. `if token_addresses is not None` can be replaced with`if token_addresses`.
2. we should always choose not to call internal functions if it can be avoided. unless there is a good reason not to, do `len(token_addresses)` instead.

mortelli (changes requested):

mortelli (approved):

PR #111, opened 2020-09-08
Feature/rlp 628/unlock light

mauricioirace:
Changelog:
- Added unlock message
- Added unlock endpoint


raiden/messages.py
```diff
@@ -1831,6 +1831,21 @@ def verify_request_monitoring(
             and recover(reward_proof_data, self.reward_proof_signature) == requesting_address
         )
 
+class UnlockRequest(Message):
```
marcosmartinez7: Lets rename this to `UnlockLightRequest`, just to know that this class is used on light client scenarios.

raiden/raiden_event_handler.py
```diff
@@ -598,6 +604,148 @@ def handle_contract_send_channelunlock(
                     partner=our_address,
                 )
 
+
+    @staticmethod
+    def handle_contract_send_channelunlock_light(
```
marcosmartinez7, line 551: This looks good but the code is duplicated. 

The only difference is the `store_light_client_protocol_message`, so, i think we can generalise `handle_contract_send_channelunlock_light `and `handle_contract_send_channelunlock`

raiden/transfer/events.py
```diff
@@ -359,6 +359,73 @@ def from_dict(cls, data: Dict[str, Any]) -> "ContractSendChannelBatchUnlock":
...
+            pex(self.token_network_identifier),
+            self.channel_identifier,
+            pex(self.participant),
+            pex(self.triggered_by_block_hash),
+        )
+
+    def __eq__(self, other: Any) -> bool:
+        return (
+            super().__eq__(other)
+            and isinstance(other, ContractSendChannelBatchUnlock)
```
marcosmartinez7: Be careful with this, you're comparing with another class

raiden/transfer/events.py
```diff
@@ -359,6 +359,73 @@ def from_dict(cls, data: Dict[str, Any]) -> "ContractSendChannelBatchUnlock":
...
+        result = {
+            "canonical_identifier": self.canonical_identifier.to_dict(),
+            "participant": to_checksum_address(self.participant),
+            "triggered_by_block_hash": serialize_bytes(self.triggered_by_block_hash),
+        }
+
+        return result
+
+    @classmethod
+    def from_dict(cls, data: Dict[str, Any]) -> "ContractSendChannelBatchUnlock":
```
marcosmartinez7: same here

marcosmartinez7 (changes requested):

raiden/raiden_event_handler.py
mauricioirace: I agree, and I think I have an idea of how to to do it.

I'll generalize in next commit.

raiden/api/python.py
```diff
@@ -1769,3 +1770,8 @@ def validate_light_client(self, api_key: str):
                 status_code=HTTPStatus.FORBIDDEN,
                 log=log
             )
+
+    def unlock_payment_light(self, signed_tx: typing.SignedTransaction, token_address: typing.TokenAddress):
+        registry: TokenNetworkRegistry = self.raiden.default_registry
+        token_network: TokenNetwork = self.raiden.chain.token_network(registry.get_token_network(token_address))
+        token_network.proxy.broadcast_signed_transaction_and_wait(signed_tx)
```
JONAF2103: Probably you will need to add a new line here

raiden/api/rest.py
```diff
@@ -86,7 +86,8 @@
     LightClientMatrixCredentialsBuildResource,
     LightClientResource,
     PaymentLightResource,
-    CreatePaymentLightResource, WatchtowerResource, LightClientMessageResource)
+    CreatePaymentLightResource, WatchtowerResource, LightClientMessageResource, UnlockedPaymentLightResource,
```
JONAF2103: Maybe we can put this as the other imports instead of all in a single line

raiden/api/rest.py
```diff
@@ -228,6 +229,10 @@
         "/light_channels/<hexaddress:token_address>/<hexaddress:creator_address>/<hexaddress:partner_address>",
         LightChannelsResourceByTokenAndPartnerAddress
     ),
+    (
+        "/unlocked_payments/<hexaddress:token_address>",
```
JONAF2103: Maybe this could be something like ```/payments_light/unlocked``` since it's a payments related feature

raiden/messages.py
```diff
@@ -1831,6 +1831,34 @@ def verify_request_monitoring(
...
+
+    def to_dict(self):
+        return {
+            "type": self.__class__.__name__,
+            "token_address": to_normalized_address(self.token_address),
+            "channel_identifier": self.channel_identifier,
+            "receiver": to_normalized_address(self.receiver),
+            "sender": to_normalized_address(self.sender)
+        }
+
```
JONAF2103, line 1876: I don't see a from_dict implementation for this object, you should add it

raiden/network/rpc/smartcontract_proxy.py
```diff
@@ -11,8 +12,9 @@
 from raiden.exceptions import (
     InsufficientFunds,
     ReplacementTransactionUnderpriced,
-    TransactionAlreadyPending,
+    TransactionAlreadyPending, RaidenRecoverableError, RawTransactionFailed,
```
JONAF2103: Same here we should keep consistency on the imports

raiden/raiden_event_handler.py
```diff
@@ -90,6 +232,31 @@ def unlock(
...
+def unlock_light(raiden: "RaidenService",
+                 chain_state,
+                 channel_unlock_event,
+                 participant: Address,
+                 partner: Address):
+    canonical_identifier: CanonicalIdentifier = channel_unlock_event.canonical_identifier
+
+    token_network: TokenNetwork = views.get_token_network_by_identifier(chain_state, canonical_identifier.token_network_address)
+
+    LightClientMessageHandler.store_light_client_protocol_message(
```
JONAF2103: Here we should check for duplicated messages before adding it to the database. I have created a new function to check this on the branch ```feature/RLP-748/settlement_light``` maybe you can use it to check for duplicated messages and if the message it's already on database you should skip this. 

When you can have duplicated messages?
When you have a lumino node that is down for some reason and when you start it again it recreates all the state events again. Then you have this code running again and adding a new message with the same data.

JONAF2103 (commented):
LGTM In general, i think some minor changes are required but it looks promising

raiden/messages.py
mauricioirace: Yes. I also forgot the pack/unpack. They are also needed right?

JONAF2103: Yes that's true

raiden/api/python.py
mortelli: we shouldn't need static type checks or hints _at all_ at this level.

the issue stemming from the fact that the type for the `self.raiden` attribute cannot be inferred should be fixed by adding expected types in the `RaidenAPI` constructor, instead of here.

raiden/api/rest.py
```diff
@@ -228,6 +236,10 @@
         "/light_channels/<hexaddress:token_address>/<hexaddress:creator_address>/<hexaddress:partner_address>",
         LightChannelsResourceByTokenAndPartnerAddress
     ),
+    (
+        "/payments_light/unlocked/<hexaddress:token_address>",
+        UnlockedPaymentLightResource
+    ),
     ("/payments_light", PaymentLightResource),
     ("/light_client_messages", LightClientMessageResource, "Message polling"),
     ("/payments_light/create", CreatePaymentLightResource, "create_payment"),
```
mortelli: i know you didn't add all these, but take a look at the lack of structure here:
- `payments_lights`/`unlocked`/`addr`
- `payments_light`
- `light_client_messages`
- `payments_light`/`create`

i see 2 issues here.

first: is there a reason why it is _unlocked_ and not _unlock_? (taking the conjugation for _create_ as reference)

secondly, ordering. for this, we can do something like:
1. `payments_light`
2. `payments_light`/`create`
3. `payments_lights`/`unlocked`/`addr`
4. `light_client_messages`

(`4` could go first, but i would keep the other 3 in that order).

i think this makes it more readable.

raiden/api/rest.py
```diff
@@ -2218,3 +2230,12 @@ def create_light_client_payment(
             return ApiErrorBuilder.build_and_log_error(errors=str(e), status_code=HTTPStatus.NOT_FOUND, log=log)
         except UnhandledLightClient as e:
             return ApiErrorBuilder.build_and_log_error(errors=str(e), status_code=HTTPStatus.FORBIDDEN, log=log)
+
+    def post_unlocked_payment_light(self, signed_tx: typing.SignedTransaction, token_address: typing.TokenAddress):
+        try:
+            self.raiden_api.unlock_payment_light(signed_tx, token_address)
+            return api_response(result=dict(), status_code=HTTPStatus.NO_CONTENT)
```
mortelli, line 2212: could you please explain why we are returning nothing here in the case of success?

raiden/lightclient/handlers/light_client_message_handler.py
```diff
@@ -42,6 +42,16 @@ def build_light_client_protocol_message(identifier: int, message: Message, signe
...
+    @classmethod
+    def is_message_already_stored(cls,
+                                  light_client_address: AddressHex,
+                                  message_type: LightClientProtocolMessageType,
+                                  unsigned_message: Message,
+                                  wal: WriteAheadLog):
+        return message_type and light_client_address and unsigned_message \
+               and wal.storage.is_message_already_stored(light_client_address,
+                                                         message_type.value,
+                                                         unsigned_message) is not None
```
mortelli: are you sure we need the `is not None` at the end of this line?

raiden/transfer/channel.py
```diff
@@ -26,7 +26,7 @@
     EventInvalidReceivedTransferRefund,
     EventInvalidReceivedUnlock,
     SendProcessed,
-    ContractSendChannelUpdateTransferLight)
+    ContractSendChannelUpdateTransferLight, ContractSendChannelBatchUnlockLight)
```
mortelli: i'd add this to line `30`, for consistency's sake

raiden/api/v1/resources.py
```diff
@@ -151,6 +151,14 @@ def get(self, **kwargs):
         )
 
 
+class UnlockedPaymentLightResource(BaseResource):
+    post_schema = UnlockedPaymentLightPutSchema()
+
+    @use_kwargs(post_schema)
+    def post(self, signed_tx: typing.SignedTransaction, **kwargs):
+        return self.rest_api.post_unlocked_payment_light(signed_tx, **kwargs)
```
mortelli: it's weird that we are passing keyword arguments to the `rest_api` method, when on the other side it expects _only_ a signed transaction plus a token address (and nothing else).

but this problem is too big to be fixed in this PR, so i'd leave as is.

raiden/raiden_event_handler.py
```diff
@@ -29,7 +31,7 @@
     EventPaymentSentFailed,
     EventPaymentSentSuccess,
     SendProcessed,
-    ContractSendChannelUpdateTransferLight)
+    ContractSendChannelUpdateTransferLight, ContractSendChannelBatchUnlockLight)
```
mortelli: i'd also place this one on line `35`

raiden/raiden_event_handler.py
```diff
@@ -73,13 +75,153 @@
...
+    unlock
+):
+    canonical_identifier = channel_unlock_event.canonical_identifier
+    token_network_identifier = channel_unlock_event.canonical_identifier.token_network_address
+    channel_identifier = channel_unlock_event.canonical_identifier.channel_identifier
+    participant = channel_unlock_event.participant
+
+    assert raiden.wal, "The Raiden Service must be initialize to handle events"
+
+    payment_channel: PaymentChannel = raiden.chain.payment_channel(
```
mortelli: we're not using this, are we?

raiden/raiden_event_handler.py
```diff
@@ -73,13 +75,153 @@
...
+    channel_unlock_event,
+    our_address,
+    unlock
+):
+    canonical_identifier = channel_unlock_event.canonical_identifier
+    token_network_identifier = channel_unlock_event.canonical_identifier.token_network_address
+    channel_identifier = channel_unlock_event.canonical_identifier.channel_identifier
+    participant = channel_unlock_event.participant
+
+    assert raiden.wal, "The Raiden Service must be initialize to handle events"
```
mortelli: typo in `initialized`

raiden/raiden_event_handler.py
```diff
@@ -73,13 +75,153 @@
...
     raiden: "RaidenService",
-    payment_channel: PaymentChannel,
-    end_state: NettingChannelEndState,
-    participant: Address,
-    partner: Address,
-) -> None:
+    chain_state: ChainState,
+    channel_unlock_event,
+    our_address,
+    unlock
```
mortelli: can we add expected types to this signature?

raiden/raiden_event_handler.py
```diff
@@ -73,13 +75,153 @@
...
+    )
+
+    if not channel_state:
+        # channel was cleaned up already due to an unlock
+        raise RaidenUnrecoverableError(
+            f"Failed to find channel state with partner:"
+            f"{to_checksum_address(participant)}, token_network:pex(token_network_identifier)"
+        )
+
+    our_address = channel_state.our_state.address
```
mortelli: this is a parameter, but we are setting it again here.

can you please explain why this is the case?

mortelli (commented):
it is still the case for me that i don't understand the business logic well enough (yet) to make corrections in that sense. the `handle_unlock` function seems suspiciously lengthy though, although it might still be correct.

going through the code, there are _a lot_ of design issues which are made apparent by looking at the diff.

however, the comments and corrections i've left were made leaving these design issues aside, as it is a bigger problem which will need to be tackled later down the road.

raiden/api/rest.py
mauricioirace: 204 No Content means the request was successfully processed, but we are not returning any content

raiden/api/rest.py
marcosmartinez7: for me `payments_light/unlock` seems correct

raiden/api/rest.py
mortelli: you're totally right, my bad

raiden/lightclient/handlers/light_client_message_handler.py
```diff
@@ -42,6 +42,16 @@ def build_light_client_protocol_message(identifier: int, message: Message, signe
 class LightClientMessageHandler:
     log = structlog.get_logger(__name__)  # pylint: disable=invalid-name
 
+    @classmethod
+    def is_message_already_stored(cls,
```
marcosmartinez7, line 46: Why is this method needed? also.. this change is added on the settlement light pr,can u explain why?

mauricioirace: that's what we talked about in slack

raiden/messages.py
mauricioirace: this is outdated

mortelli (approved):
untested LGTM

raiden/transfer/unlock.py
```diff
@@ -0,0 +1,141 @@
...
+from raiden.transfer.channel import get_batch_unlock_gain
+from raiden.transfer.identifiers import CanonicalIdentifier
+from raiden.transfer.state import ChainState, NettingChannelState, NettingChannelEndState
+from raiden.transfer.utils import get_state_change_with_balance_proof_by_locksroot, \
+    get_event_with_balance_proof_by_locksroot
+from raiden.transfer.views import get_channelstate_by_token_network_and_partner
+from raiden.utils import Address
+from raiden.utils.typing import TokenNetworkID
+
+def get_channel_state(
```
marcosmartinez7, line 17: There are already some functions to retrieve chanenl state on views, why is this function needed?

mauricioirace: This function was part of handle_batch_unlock function. I just encapsulated it into another function in order to make it shorter and cleaner in the light version, without changing the idea behind the code. I like too keep the functions as shorter as possible, grouping them in smaller functions if necessary.

Anyway, if there is a function that do exactly this (getting a channel_state or 'die' trying) I'll cyhange it by it.

If there's a method

raiden/transfer/unlock.py
```diff
@@ -0,0 +1,141 @@
...
+
+def should_search_state_changes(
+    channel_state: NettingChannelState,
+) -> bool:
+    partner_locksroot = channel_state.partner_state.onchain_locksroot
+    # we want to unlock, because there are unlocked/unclaimed locks
+    return partner_locksroot != EMPTY_HASH
+
+
+def get_partner_state_for_unlock(
```
mauricioirace, line 63: Also, this could be removed, as we are not restoring the chanel state in the light version

raiden/transfer/unlock.py
```diff
@@ -0,0 +1,141 @@
...
+    return None
+
+
+def get_our_state_for_unlock(
+    raiden: "RaidenService",
+    canonical_identifier: CanonicalIdentifier,
+    participant: Address,
+    channel_state: NettingChannelState
+) -> Optional[NettingChannelEndState]:
+    if should_search_events(channel_state):
```
mauricioirace, line 109: Same here

JONAF2103:
Is this duplicated or related to this? https://github.com/rsksmart/lumino/pull/126

marcosmartinez7:
@mauricioirace can we close this pr and continue on https://github.com/rsksmart/lumino/pull/126?

PR #112, opened 2020-09-08
RLP-765: add remove light client transport functionality

mortelli:
This PR adds a method to the `NodeTransport` class to remove light client transport layers.

It is unused as of now, but it's a one-liner that might come in handy later.

Closes [RLP-765](https://jirainfuy.atlassian.net/browse/RLP-765).


marcosmartinez7 (approved):
lgtm

PR #113, opened 2020-09-08
RLP-766: address normalization

mortelli:
This PR normalizes the use of the `address` field in the `MatrixTransport` and `MatrixLightClientTransport` classes, in terms of complying with the `TransportLayer` abstraction and removing redundancy for this field.

More specifically:
- `address` is now a field for the `MatrixTransport` constructor (and therefore of the `MatrixLightClientTransport` constructor as well) instead of being hidden inside the `config` dict.
- `TransportLayer.address` method is now decorated with `@property`.
- all matrix transport layer references to `._address` and `get_address` were replaced with `TransportLayer.address` calls.
- `get_address` method is removed, since it is redundant.
- `config["transport"]["matrix"]["address"]` is also removed, since it is no longer used.

Closes [RLP-766](https://jirainfuy.atlassian.net/browse/RLP-766).

### DIFF
--- a/raiden/api/rest.py
+++ b/raiden/api/rest.py
@@ -2072,11 +2072,11 @@ class RestAPI:
         if light_client is not None and light_client["result_code"] == 200:
 
             matrix_light_client_transport_instance = get_matrix_light_client_instance(
-                self.raiden_api.raiden.config["transport"]["matrix"],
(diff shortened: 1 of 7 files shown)


marcosmartinez7 (approved):
lgtm

PR #110, opened 2020-09-07
RLP-750: transport generalization

mortelli:
This PR adds the initial implementation for the **transport layer abstraction**, as described in the [Transport Research issue](https://jirainfuy.atlassian.net/browse/RLP-749). Presentations on this matter can be found [here](https://github.com/anarancio/lumino-docs/tree/master/presentations/transport).

This PR combines the result of multiple others (which build on this one) in order to carry out the implementation and solve the design problems previously laid out in the research documentation. The full list of PRs is [here](https://github.com/rsksmart/lumino/pulls?q=is%3Apr+label%3Atransport+is%3Aclosed).

~Tests are not being updated right now, so they will break. This will be tackled through a [separate issue](https://jirainfuy.atlassian.net/browse/RLP-775).~ Test suite is now functional; results can be found [here](https://github.com/anarancio/lumino-docs/tree/master/unit-testing/results/feature/RLP-750_transport-generalization).

After this PR is merged, the RIF Comms transport layer implementation can be added to the `rif-comms` branch. After this implementation is completed, `rif-comms` can be merged into `develop` and eventually `master`.

PR #114, opened 2020-09-08
RLP-769: queue identifier obfuscation

mortelli:
This PR iterates on the message sending function of transport layer abstraction, with the aim of obfuscating the Matrix-specific queue identifier parameter from its signature.

### details
- the `send_async` method is now called `send_message`, and its parameters have changed so that it receives a _transport message_ and a _recipient_.
  - _transport message_ is a wrapper class (located in `transport.components`) which contains an embedded raiden message, plus whatever other parameters that need to be passed down to the transport layer implementations (see **design** for more on this).
  - the _recipient_ is a regular address.
- wrapping and unwrapping methods have been added to more easily go to and from a _transport message_ to the components (like queue identifiers) that both lumino and matrix understand and are already working with.
- due to being in the `transport` package, the word _transport_ was removed from the class names in that package.
  - outside of it, however, named imports have been used to avoid confusion with generic or already-present names.
- code formatting and import optimizations has been applied to all of the changed files.

### design
the original idea was to have queue identifier completely absent from any of the generic transport layer code. 

the main reason for why this is not implemented is that, unfortunately, the `QueueIdentifier` type is not _just_ part of the matrix transport layer, but it's actually part of the upper layers as well. it turns out that raiden/lumino events and messages already contain queue identifiers in them, so it's not feasible to keep all matrix-specific code beneath the generic transport layer and also achieve this within the time frame that we have.

since:
1. queue identifiers are—for the foreseeable future—here to stay,
2. it's not expected for rif communications to have such a field directly as a part of its message sending signature,

the easiest solution in this case is to create a container class for messages at the generic transport layer level, which include the queue identifier in its params. 

then, this _transport message_ class knows how to build itself from the components that come from the upper layers, i.e. the queue identifier and the raiden message itself; and then the matrix transport layer knows how to put this transport message apart and work with its composing fields (rather than a new class) so that it maintains its current working logic.

eventually, the rif comms transport layer will know how to unwrap this class as well in order to process the message in its own manner.

in consequence of this:
- it is not needed to apply any modifications to the matrix code, except packing/unpacking this container class right before transport logic needs to be executed.
- rif communications can ignore this parameter if it turns out it's not needed for its transport layer code.
- we could eventually add additional data to the _transport message_ params, if rif communication needs it. 

the evident drawback is, of course, having information in the transport layer which might be specific to just _one_ implementation, and passing it around needlessly. 

however, we cannot afford neither altering the raiden/lumino business logic at the message and event levels, nor changing the way the matrix transport implementation works right now.

additionally, this drawback is somewhat mitigated by wrapping and unwrapping the transport messages in just a handful of places, so that unneeded information does not go further than it needs to.

---

closes [RLP-769](https://jirainfuy.atlassian.net/browse/RLP-769).


raiden/raiden_service.py
```diff
@@ -922,17 +922,6 @@ def _initialize_messages_queues(self, chain_state: ChainState):
         for queue_identifier, _event_queue in events_queues.items():
             self.start_health_check_for(queue_identifier.recipient)
 
-            # for event in event_queue:
```
mortelli, line 924: not sure why this was commented out in the first place...

if anyone knows, please fill me in on this

transport/abstraction.py
```diff
@@ -2,16 +2,22 @@
...
 from raiden.message_handler import MessageHandler
-from raiden.messages import Message
 from raiden.raiden_service import RaidenService
-from raiden.transfer.identifiers import QueueIdentifier
 from raiden.utils import Address
+from transport.components import Message
 
 
-class TransportLayer(ABC):
+class Layer(ABC):
```
JONAF2103: Maybe Layer is too generic as a name, because if you think this you could ask, why a type called Layer have send_message as a member? Maybe you can use something a little more specific here like CommunicationsLayer or something else.

JONAF2103 (commented):
Looks good so far

transport/abstraction.py
mortelli: it's already in the transport package, so it's actually `transport.abstractions.Layer`

mortelli: also, i made sure to used named imports elsewhere to avoid confusion

raiden/network/transport/matrix/transport.py
```diff
@@ -321,6 +321,14 @@ def _http_retry_delay() -> Iterable[float]:
 
         self._message_handler: Optional[MessageHandler] = None
 
+    @classmethod
+    def unwrap(cls, transport_message: TransportMessage) -> (Message, QueueIdentifier):
```
marcosmartinez7: the first tought that came to my mind was this "unwrap what? a transport?" 

i think unrwap_message looks better

mortelli: isn't it implicit in the parameter name?

marcosmartinez7: I also think that since we have the wrap method on TransportMessage we should have this unwrap method there too

A TransportMessage knows how to unwrap itself, i dont think is a transport responsability

raiden/network/transport/matrix/transport.py
```diff
@@ -498,33 +506,32 @@ def start_health_check(self, node_address):
             # representing the target node
             self._address_mgr.refresh_address_presence(node_address)
 
-    def send_async(self, queue_identifier: QueueIdentifier, message: Message):
+    def send_message(self, message: TransportMessage, recipient: Address):
```
marcosmartinez7, line 494: neat

raiden/raiden_service.py
marcosmartinez7: there is already a task for that commented code, we can remove it (RLP 344)

transport/components.py
```diff
@@ -0,0 +1,35 @@
+from raiden.messages import Message as RaidenMessage
+from raiden.transfer.identifiers import QueueIdentifier
+from raiden.utils import Address
+
+
+class Params:
```
marcosmartinez7: Personally i dont like the Params classname. 

We are at transport/components, therefore this say to me "the parameters is a component of a transport", but, the parameters are a component of the transport message

I think MessageParams is better.

transport/components.py
```diff
@@ -0,0 +1,35 @@
...
+        """
+        The QueueIdentifier coming from the (upper) Lumino business logic layer.
+        """
+
+
+class Message:
+    """
+    Message is a wrapper class which embeds a Raiden Message, plus any optional transport layer params.
+    """
+
```
marcosmartinez7: i would move the unwrap method here

raiden/network/transport/matrix/transport.py
mortelli: you're right. i've moved this to the `transport.components.Message` class

marcosmartinez7: yeah, both unwrap and unwrap_message are ok to me, not sure about the location of this method

transport/components.py
mortelli: done

raiden/network/transport/matrix/transport.py
mortelli: if the message knows how to wrap and unwrap itself, then i think it should be a method of the message class

transport/components.py
mortelli: i agree, fixed

raiden/network/transport/matrix/transport.py
```diff
@@ -498,33 +498,32 @@ def start_health_check(self, node_address):
...
-                "Do not use send_async for {} messages".format(message.__class__.__name__)
+                "Do not use send_message for {} messages".format(raiden_message.__class__.__name__)
             )
 
         self.log.info(
-            "Send async",
-            receiver_address=pex(receiver_address),
-            message=message,
+            "Send message",
+            receiver_address=pex(recipient),
```
marcosmartinez7: i will probably change the logger parameter too to `recipient`

mortelli: done

raiden/network/transport/matrix/transport.py
marcosmartinez7: agree

marcosmartinez7 (approved):
Lgtm

mortelli:
putting this on hold until we make sure #110 is working ok with what we have so far

mortelli:
> putting this on hold until we make sure #110 is working ok with what we have so far

this has now [been verified](https://github.com/rsksmart/lumino/pull/110#issuecomment-702364526).

as a result this PR is no longer on hold; as soon as i re-run the test suite plus the basic manual testing scenarios, it will be ready to be merged.

mortelli:
[as done in](https://github.com/rsksmart/lumino/pull/110#issuecomment-702364526) PR #110, i've manually verified basic operations in the following scenarios:
- `FN` ←→ `FN`
- `FN` ←→ `FN` ←→ `FN`
- `LC` ←→ `FN`
- `LC` ←→ `FN` ←→ `FN`

i've also made sure the test suite results match [the latest ones for the destination branch](https://github.com/anarancio/lumino-docs/blob/master/unit-testing/results/transport-generalization/01-10-2020/summary.txt).

with these 2 things now done, this PR is ready to be merged.

PR #115, opened 2020-09-08
Feature/rlp 748/settlement light

JONAF2103:
This PR introduces the ability to execute the settlement process from light clients.

Here we are:

* Adding new state change to store a message for the LC
* Adding new message for the LC as a payload to be signed
* Adding the handling of the new state to store the message
* Adding new endpoint to handle the settlement light with the signed message

Now a light client will be able to settle a channel after the channel is closed and ready for settle.

### DIFF
--- a/raiden/api/python.py
+++ b/raiden/api/python.py
@@ -51,7 +51,7 @@ from raiden.lightclient.models.light_client_payment import LightClientPayment, L
 from raiden.lightclient.models.light_client_protocol_message import LightClientProtocolMessageType
 
 from raiden.messages import RequestMonitoring, LockedTransfer, RevealSecret, Unlock, Delivered, SecretRequest, \
-    Processed, LockExpired
+    Processed, LockExpired, SettlementRequest
 from raiden.settings import DEFAULT_RETRY_TIMEOUT, DEVELOPMENT_CONTRACT_VERSION
 
 from raiden.transfer import architecture, views
@@ -455,6 +455,11 @@ class RaidenAPI:
         )
         assert channel_state, f"channel {channel_state} is gone"
         return channel_state.identifier
+
+    def send_settlement_light(self, message: Dict, signed_tx: typing.SignedTransaction):
+        request = SettlementRequest.from_dict(message)
+        token_network = self.raiden.chain.token_network(request.channel_network_identifier)
+        token_network.send_settlement_light(signed_tx)
 
     def channel_open(
         self,
--- a/raiden/api/rest.py
+++ b/raiden/api/rest.py
@@ -86,7 +86,7 @@ from raiden.api.v1.resources import (
     LightClientMatrixCredentialsBuildResource,
     LightClientResource,
     PaymentLightResource,
-    CreatePaymentLightResource, WatchtowerResource, LightClientMessageResource)
+    CreatePaymentLightResource, WatchtowerResource, LightClientMessageResource, SettlementLightResource)
 
 from raiden.constants import GENESIS_BLOCK_NUMBER, UINT256_MAX, Environment, EMPTY_PAYMENT_HASH_INVOICE
 
@@ -231,6 +231,9 @@ URLS_HUB_V1 = [
     ("/payments_light", PaymentLightResource),
     ("/light_client_messages", LightClientMessageResource, "Message polling"),
     ("/payments_light/create", CreatePaymentLightResource, "create_payment"),
+    (
+        '/payments_light/settlement', SettlementLightResource,
+    ),
     ("/watchtower", WatchtowerResource),
     (
         '/light_clients/matrix/credentials',
@@ -2218,3 +2221,17 @@ class RestAPI:
             return ApiErrorBuilder.build_and_log_error(errors=str(e), status_code=HTTPStatus.NOT_FOUND, log=log)
         except UnhandledLightClient as e:
             return ApiErrorBuilder.build_and_log_error(errors=str(e), status_code=HTTPStatus.FORBIDDEN, log=log)
+
+    def receive_light_client_signed_settlement(self, message: Dict, signed_tx: typing.SignedTransaction):
+        headers = request.headers
+        api_key = headers.get("x-api-key")
+        if not api_key:
+            return ApiErrorBuilder.build_and_log_error(errors="Missing api_key auth header",
+                                                       status_code=HTTPStatus.BAD_REQUEST, log=log)
+        light_client = LightClientService.get_by_api_key(api_key, self.raiden_api.raiden.wal)
+        if not light_client:
+            return ApiErrorBuilder.build_and_log_error(
+                errors="There is no light client associated with the api key provided",
+                status_code=HTTPStatus.FORBIDDEN, log=log)
+
+        self.raiden_api.send_settlement_light(message, signed_tx)
--- a/raiden/api/v1/resources.py
+++ b/raiden/api/v1/resources.py
@@ -27,7 +27,7 @@ from raiden.api.v1.encoding import (
     LightClientMatrixCredentialsBuildSchema,
     PaymentLightPutSchema,
     CreatePaymentLightPostSchema,
-    WatchtowerPutResource, LightClientMessageGetSchema)
+    WatchtowerPutResource, LightClientMessageGetSchema, SettlementLightSchema)
 from raiden.messages import  Unlock
 
 from raiden.utils import typing
@@ -301,6 +301,19 @@ class CreatePaymentLightResource(BaseResource):
             amount=amount,
             secrethash=secrethash
         )
+
+
+class SettlementLightResource(BaseResource):
+    put_schema = SettlementLightSchema
+
+    @use_kwargs(put_schema, locations=("json",))
+    def put(self,
+            message: Dict,
+            signed_tx: str):
+        """
+        put a settlement signed tx to be send by the hub
+        """
+        return self.rest_api.receive_light_client_signed_settlement(message, signed_tx)
 
 
 class LightClientMessageResource(BaseResource):
--- a/raiden/messages.py
+++ b/raiden/messages.py
@@ -6,7 +6,7 @@ from eth_utils import (
     decode_hex,
     encode_hex,
     to_canonical_address,
-    to_normalized_address,
+    to_normalized_address, to_checksum_address,
 )
 
 from raiden.constants import UINT64_MAX, UINT256_MAX, EMPTY_PAYMENT_HASH_INVOICE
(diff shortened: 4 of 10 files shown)


raiden/messages.py
```diff
@@ -1832,6 +1832,89 @@ def verify_request_monitoring(
         )
 
 
+class SettlementRequest(Message):
```
marcosmartinez7, line 1835: Lets rename this to `SettlementLightRequest`, in order to know that this class is related with light clients

raiden/api/python.py
```diff
@@ -456,6 +456,11 @@ def channel_open_light(
         assert channel_state, f"channel {channel_state} is gone"
         return channel_state.identifier
 
+    def send_settlement_light(self, message: Dict, signed_tx: typing.SignedTransaction):
+        request = SettlementRequest.from_dict(message)
+        token_network = self.raiden.chain.token_network(request.channel_network_identifier)
+        token_network.send_settlement_light(signed_tx)
```
marcosmartinez7, line 462: We need to catch exceptions here

raiden/api/python.py
```diff
@@ -456,6 +456,11 @@ def channel_open_light(
...
+
```
mortelli, line 463: maybe it would be useful to have a `return` here (we'd need to add one to `token_network.send_settlement_light` as well)

raiden/api/rest.py
```diff
@@ -231,6 +231,9 @@
     ("/payments_light", PaymentLightResource),
     ("/light_client_messages", LightClientMessageResource, "Message polling"),
     ("/payments_light/create", CreatePaymentLightResource, "create_payment"),
+    (
+        '/payments_light/settlement', SettlementLightResource,
+    ),
```
mortelli, line 236: we can condense this to 1 line

raiden/api/rest.py
```diff
@@ -231,6 +231,9 @@
     ("/payments_light", PaymentLightResource),
     ("/light_client_messages", LightClientMessageResource, "Message polling"),
     ("/payments_light/create", CreatePaymentLightResource, "create_payment"),
+    (
+        '/payments_light/settlement', SettlementLightResource,
```
marcosmartinez7, line 235: I dont think this is the correct resource since settlement isnt related with payments at all. 

Take a look at `_close_light ` rest method, its invoked from `patch_channel_light` and the resource is `LightChannelsResourceByTokenAndPartnerAddress`

raiden/api/rest.py
```diff
@@ -2218,3 +2221,17 @@ def create_light_client_payment(
             return ApiErrorBuilder.build_and_log_error(errors=str(e), status_code=HTTPStatus.NOT_FOUND, log=log)
         except UnhandledLightClient as e:
             return ApiErrorBuilder.build_and_log_error(errors=str(e), status_code=HTTPStatus.FORBIDDEN, log=log)
+
+    def receive_light_client_signed_settlement(self, message: Dict, signed_tx: typing.SignedTransaction):
+        headers = request.headers
+        api_key = headers.get("x-api-key")
```
mortelli, line 2227:
```suggestion
        api_key = request.headers.get("x-api-key")
```

mortelli, line 2227: we could also extract `x-api-key` to a constant since it is used 5 times in `rest.py`

raiden/messages.py
marcosmartinez7: `/light_channels` seems better as a path

raiden/network/proxies/payment_channel.py
```diff
@@ -283,6 +284,8 @@ def unlock(
 
     def settle(
         self,
+        raiden: "RaidenService",
+        channel_settle_event: ContractSendChannelSettle,
```
marcosmartinez7, line 288: why we need to modify the `settle `function and receive the `channel_settle_event`?

raiden/network/proxies/token_network.py
```diff
@@ -2158,6 +2176,9 @@ def settle(
         partner_locksroot: Locksroot,
         given_block_identifier: BlockSpecification,
     ):
+
+        channel_identifier = channel_settle_event.channel_identifier
```
marcosmartinez7, line 2180: why the settle function must be modified? this is related with the previous comment

raiden/api/rest.py
```diff
@@ -2218,3 +2221,17 @@ def create_light_client_payment(
...
+        if not api_key:
+            return ApiErrorBuilder.build_and_log_error(errors="Missing api_key auth header",
+                                                       status_code=HTTPStatus.BAD_REQUEST, log=log)
+        light_client = LightClientService.get_by_api_key(api_key, self.raiden_api.raiden.wal)
+        if not light_client:
+            return ApiErrorBuilder.build_and_log_error(
+                errors="There is no light client associated with the api key provided",
+                status_code=HTTPStatus.FORBIDDEN, log=log)
+
+        self.raiden_api.send_settlement_light(message, signed_tx)
```
mortelli, line 2237: same comment about the `return` here

raiden/api/v1/resources.py
```diff
@@ -303,6 +303,19 @@ def post(
...
 
+class SettlementLightResource(BaseResource):
+    put_schema = SettlementLightSchema
+
+    @use_kwargs(put_schema, locations=("json",))
+    def put(self,
+            message: Dict,
+            signed_tx: str):
+        """
+        put a settlement signed tx to be send by the hub
```
mortelli, line 314: `send` → `sent`

raiden/network/proxies/token_network.py
```diff
@@ -2213,47 +2239,66 @@ def settle(
             # gas will stop us from sending a transaction that will fail
             pass
 
-        with self.channel_operations_lock[partner]:
-            error_prefix = "Call to settle will fail"
-            gas_limit = self.proxy.estimate_gas(
-                checking_block, "settleChannel", channel_identifier=channel_identifier, **kwargs
-            )
+        if self_or_lc_address == self.node_address:
```
marcosmartinez7, line 2242: So we have the settle function, both for light clients and full nodes? i think is better to just receive a flag if the settlement that we want to do is for a light client or not to avoid this:

```

        self_or_lc_address = channel_settle_event.channel_state.our_state.address if \
            channel_settle_event.channel_state.our_state.address != partner else \
            channel_settle_event.channel_state.partner_state.address
```

Or just have a settle_light, exactly as close_light on token_network proxy. Its cleaner for me

raiden/network/proxies/token_network.py
```diff
@@ -2213,47 +2239,66 @@ def settle(
...
+                )
+                error_msg = f"{error_prefix}. {msg}"
+                log.critical(error_msg, **log_details)
+                raise RaidenUnrecoverableError(error_msg)
+
+            log.info("settle successful", **log_details)
+        else:
+            log.info("Is light client, storing new state to handle message for settlement.", **log_details)
+            # is LC, we create a new state to store a message for the LC
+            store_settlement_message = ActionStoreSettlementMessage(
```
marcosmartinez7, line 2288: I dont think this action of storing a message is a state change, i think is more an event "store this on the db" (we have the StoreMessageEvent for that)

Also is not a great idea to invoke the handle and track state change from a token network proxy. This class is used to invoke onchain operations with TokenNetwork contract, not to generate new state changes

marcosmartinez7 (changes requested):

raiden/network/proxies/token_network.py
```diff
@@ -248,6 +250,21 @@ def new_netting_channel_light(self, creator: Address, partner: Address, signed_t
...
+            transaction_hash = self.proxy.broadcast_signed_transaction(signed_tx)
+            self.client.poll(transaction_hash)
+            receipt_or_none = check_transaction_threw(self.client, transaction_hash)
+            if receipt_or_none:
+                raise RaidenRecoverableError("trying to settle channel failed")
+        except HTTPError as e:
+            log.warning("settle channel failed: transaction malformed", ex=e)
+            raise RawTransactionFailed("Light Client raw transaction malformed")
+        except Exception as e:
+            log.warning("settle channel failed", ex=e)
```
mortelli, line 265: shouldn't we be logging at `error` level here? (instead of `warning`)

raiden/network/proxies/token_network.py
```diff
@@ -2180,20 +2201,25 @@ def settle(
 
         # The second participant transferred + locked amount must be higher
         our_bp_is_larger = our_maximum > partner_maximum
+
+        self_or_lc_address = channel_settle_event.channel_state.our_state.address if \
```
mortelli, line 2205: this should be named `participant2address`

raiden/network/proxies/token_network.py
mortelli, line 2242: same here

raiden/network/proxies/token_network.py
```diff
@@ -2213,47 +2239,66 @@ def settle(
...
 
-        with self.channel_operations_lock[partner]:
-            error_prefix = "Call to settle will fail"
-            gas_limit = self.proxy.estimate_gas(
-                checking_block, "settleChannel", channel_identifier=channel_identifier, **kwargs
-            )
+        if self_or_lc_address == self.node_address:
+            # is not lc
+            with self.channel_operations_lock[partner]:
+                error_prefix = "Call to settle will fail"
```
mortelli, line 2245: we can do either uppercase or lowercase on prefixes, but let's be consistent on this

raiden/network/proxies/token_network.py
```diff
@@ -2213,47 +2239,66 @@ def settle(
...
+                    receipt_or_none = check_transaction_threw(self.client, transaction_hash)
 
-        transaction_executed = gas_limit is not None
-        if not transaction_executed or receipt_or_none:
-            if transaction_executed:
-                block = receipt_or_none["blockNumber"]
-            else:
-                block = checking_block
+            transaction_executed = gas_limit is not None
+            if not transaction_executed or receipt_or_none:
```
mortelli, line 2261: it's a bit hard to understand semantically what _if not transaction_executed or receipt_or_none_ means.

the problem here is the name of the `receipt_or_none` field. please name it to something more descriptive or declare and use a boolean variable to make this `if` more readable

raiden/raiden_event_handler.py
```diff
@@ -603,104 +603,102 @@ def handle_contract_send_channelunlock(
...
         raiden: "RaidenService", channel_settle_event: ContractSendChannelSettle
     ):
-        # only make settlements if our node is a participant, dont do it for lc
-        if channel_settle_event.channel_state.our_state.address == raiden.address or channel_settle_event.channel_state.partner_state.address == raiden.address:
-            assert raiden.wal, "The Raiden Service must be initialize to handle events"
-            canonical_identifier = CanonicalIdentifier(
-                chain_identifier=raiden.chain.network_id,
-                token_network_address=channel_settle_event.token_network_identifier,
-                channel_identifier=channel_settle_event.channel_identifier,
+        assert raiden.wal, "The Raiden Service must be initialize to handle events"
```
mortelli, line 606: are you sure that by checking `wal != None` we can assert that the raiden service is initialized?

(also: typo on `initialized`)

mortelli (commented):
some style/programming comments based on a quick glance through the code

raiden/network/proxies/token_network.py
mortelli: i agree here, it jumped out to me that we now need to pass a `raiden`-typed parameter to the `settle` function.

let's please make sure we absolutely cannot do without this

PR #117, opened 2020-09-09
Feature/rlp 748/settlement light

JONAF2103:
This PR introduces the ability to execute the settlement process from light clients.

Here we are:

* Adding the new event to handle for LC's
* Refactoring store_light_client_protocol_message to put payment_id as optional at the end of the arguments since it's an optional parameter on the database
* Adding the event handler for the ContractSendChannelSettleLight event
* Refactoring the method handle_contract_send_channelsettle to encapsulate the logic in an internal method to reuse that logic for light settlement, basically we use the same parameters to handle the operation
* Adding SettlementParameters object to encapsulate the parameters generated by process_data_and_get_settlement_parameters
* Adding a new type of message to store for LC's
* Adding logic to patch_light_channel to handle settle with a signed settle transaction
* Adding handle method for that logic
* Adding handle method for settle in payment channel proxy
* Adding handle method for settle in token network proxy
* Adding view and method to validate requests for settle

Now a light client will be able to settle a channel after the channel is closed and ready for settle.

Test results:

```
============================= test session starts ==============================
platform linux -- Python 3.7.3, pytest-5.2.0, py-1.9.0, pluggy-0.13.1
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/home/jonathan/workspace/rsk/lumino-settlement/.hypothesis/examples')
rootdir: /home/jonathan/workspace/rsk/lumino-settlement, inifile: setup.cfg
plugins: random-0.2, select-0.1.2, hypothesis-3.88.3, timeout-1.3.3, cov-2.5.1
timeout: 540.0s
timeout method: signal
timeout func_only: False
collected 659 items

raiden/tests/test_message_delivered.py .                                 [  0%]
raiden/tests/fuzz/test_state_changes.py FF.FFFFs                         [  1%]
raiden/tests/integration/test_balance_proof_check.py ..                  [  1%]
raiden/tests/integration/test_blockchainservice.py .                     [  1%]
raiden/tests/integration/test_echo_node.py sss                           [  2%]
raiden/tests/integration/test_pythonapi_base.py .....s......             [  4%]
raiden/tests/integration/test_raidenservice.py .X                        [  4%]
raiden/tests/integration/test_recovery.py sss                            [  4%]
raiden/tests/integration/test_regression.py .F..                         [  5%]
raiden/tests/integration/test_regression_parity.py s                     [  5%]
raiden/tests/integration/test_send_queued_messages.py ..                 [  5%]
raiden/tests/integration/api/test_pythonapi.py .....s......              [  7%]
raiden/tests/integration/api/test_pythonapi_regression.py .              [  7%]
raiden/tests/integration/api/test_restapi.py ...x....................... [ 11%]
..........                                                               [ 13%]
raiden/tests/integration/cli/test_cli_development.py ss                  [ 13%]
raiden/tests/integration/cli/test_cli_production.py sssssss              [ 14%]
raiden/tests/integration/long_running/test_integration_events.py .....   [ 15%]
raiden/tests/integration/long_running/test_settlement.py F.F..xx..       [ 16%]
raiden/tests/integration/long_running/test_stress.py s                   [ 16%]
raiden/tests/integration/long_running/test_token_networks.py F           [ 17%]
raiden/tests/integration/network/proxies/test_discovery.py ..            [ 17%]
raiden/tests/integration/network/proxies/test_payment_channel.py .       [ 17%]
raiden/tests/integration/network/proxies/test_secret_registry.py ....    [ 18%]
raiden/tests/integration/network/proxies/test_service_registry.py ..     [ 18%]
raiden/tests/integration/network/proxies/test_token.py .                 [ 18%]
raiden/tests/integration/network/proxies/test_token_network.py .....     [ 19%]
raiden/tests/integration/network/proxies/test_token_network_registry.py . [ 19%]
                                                                         [ 19%]
raiden/tests/integration/network/transport/test_matrix_transport.py .... [ 20%]
.......F....ss.................x..                                       [ 25%]
raiden/tests/integration/network/transport/test_udp.py ss.               [ 25%]
raiden/tests/integration/rpc/test_client.py .                            [ 25%]
raiden/tests/integration/rpc/assumptions/test_parity_rpc_assumptions.py s [ 26%]
                                                                         [ 26%]
raiden/tests/integration/rpc/assumptions/test_rpc_call_assumptions.py .. [ 26%]
..                                                                       [ 26%]
raiden/tests/integration/rpc/assumptions/test_rpc_filters_assumptions.py . [ 26%]
.                                                                        [ 27%]
raiden/tests/integration/rpc/assumptions/test_rpc_gas_estimation_assumptions.py . [ 27%]
.                                                                        [ 27%]
raiden/tests/integration/rpc/assumptions/test_rpc_gas_price_assumptions.py . [ 27%]
.                                                                        [ 27%]
raiden/tests/integration/rpc/assumptions/test_rpc_transaction_assumptions.py . [ 27%]
..                                                                       [ 28%]
raiden/tests/integration/transfer/test_mediatedtransfer.py ....sss       [ 29%]
raiden/tests/integration/transfer/test_mediatedtransfer_events.py .      [ 29%]
raiden/tests/integration/transfer/test_mediatedtransfer_invalid.py ..... [ 30%]
.                                                                        [ 30%]
raiden/tests/integration/transfer/test_refund_invalid.py .               [ 30%]
raiden/tests/integration/transfer/test_refundtransfer.py ....            [ 30%]
raiden/tests/unit/test_accounts.py ......                                [ 31%]
raiden/tests/unit/test_binary_encoding.py .............................. [ 36%]
......................                                                   [ 39%]
raiden/tests/unit/test_case_utils.py ......                              [ 40%]
raiden/tests/unit/test_channelstate.py .............................     [ 45%]
raiden/tests/unit/test_cli.py .............                              [ 47%]
raiden/tests/unit/test_dict_encoding.py ................................ [ 51%]
................                                                         [ 54%]
raiden/tests/unit/test_light_client_messages.py .........                [ 55%]
raiden/tests/unit/test_lightclient_onboarding.py ........                [ 56%]
raiden/tests/unit/test_logging.py ...................................... [ 62%]
.                                                                        [ 62%]
raiden/tests/unit/test_matrix_presence.py .....                          [ 63%]
raiden/tests/unit/test_matrix_transport_unit.py ......                   [ 64%]
raiden/tests/unit/test_message_locked_transfer.py ....                   [ 65%]
raiden/tests/unit/test_messages.py .....                                 [ 65%]
raiden/tests/unit/test_mock.py .                                         [ 66%]
raiden/tests/unit/test_mtree.py .....                                    [ 66%]
raiden/tests/unit/test_namedbuffer.py .....                              [ 67%]
raiden/tests/unit/test_node_queue.py ...                                 [ 67%]
raiden/tests/unit/test_notifyingqueue.py ...                             [ 68%]
raiden/tests/unit/test_operators.py ...                                  [ 68%]
raiden/tests/unit/test_pfs_integration.py ...................            [ 71%]
raiden/tests/unit/test_raiden_event_handler.py .                         [ 71%]
raiden/tests/unit/test_rpc.py .                                          [ 72%]
raiden/tests/unit/test_serialization.py ........                         [ 73%]
raiden/tests/unit/test_signature_lc.py .                                 [ 73%]
raiden/tests/unit/test_sqlite.py .....                                   [ 74%]
raiden/tests/unit/test_startup.py ..................                     [ 76%]
raiden/tests/unit/test_state.py ..                                       [ 77%]
raiden/tests/unit/test_tokennetwork.py ...F....                          [ 78%]
raiden/tests/unit/test_udp_transport.py ssss                             [ 79%]
raiden/tests/unit/test_upgrade.py .....                                  [ 79%]
raiden/tests/unit/test_version_check.py ...                              [ 80%]
raiden/tests/unit/test_wal.py .......                                    [ 81%]
raiden/tests/unit/api/test_api.py ....                                   [ 81%]
raiden/tests/unit/api/test_api_events.py ...                             [ 82%]
raiden/tests/unit/storage/test_storage.py ....                           [ 83%]
raiden/tests/unit/storage/test_versions.py ..                            [ 83%]
raiden/tests/unit/storage/migrations/test_v16_to_v17.py .                [ 83%]
raiden/tests/unit/storage/migrations/test_v17_to_v18.py .                [ 83%]
raiden/tests/unit/storage/migrations/test_v18_to_v19.py .                [ 83%]
raiden/tests/unit/storage/migrations/test_v19_to_v20.py .                [ 83%]
raiden/tests/unit/storage/migrations/test_v20_to_v21.py .                [ 84%]
raiden/tests/unit/storage/migrations/test_v21_to_v22.py .                [ 84%]
raiden/tests/unit/transfer/test_channel.py ....                          [ 84%]
raiden/tests/unit/transfer/test_node.py .                                [ 84%]
raiden/tests/unit/transfer/test_utils.py .....                           [ 85%]
raiden/tests/unit/transfer/mediated_transfer/test_events.py .            [ 85%]
raiden/tests/unit/transfer/mediated_transfer/test_initiatorstate.py .... [ 86%]
.........................                                                [ 90%]
raiden/tests/unit/transfer/mediated_transfer/test_mediatorstate.py ..... [ 91%]
F.....F.......F.........F.......F.....                                   [ 96%]
raiden/tests/unit/transfer/mediated_transfer/test_mediatorstate_regression.py . [ 96%]
F.FFF.                                                                   [ 97%]
raiden/tests/unit/transfer/mediated_transfer/test_targetstate.py ....... [ 98%]
......x                                                                  [100%]
= 21 failed, 601 passed, 32 skipped, 5 xfailed, 1 xpassed, 9 warnings in 9527.27s (2:38:47) =
```

Here the entire log for testing:

[tests.log](https://github.com/rsksmart/lumino/files/5269264/tests.log)

### DIFF
--- a/raiden/api/objects.py
+++ b/raiden/api/objects.py
@@ -1,3 +1,6 @@
+from raiden.utils.typing import TokenAmount, Locksroot, BlockHash
+
+
 class FlatList(list):
     """
     This class inherits from list and has the same interface as a list-type.
@@ -56,3 +59,22 @@ class DashboardGeneralItem:
     quantity = 0
     event_type_code = 0
     event_type_class_name = ""
+
+
+class SettlementParameters:
+    """" Object to store settlement parameters for internal use """
+    def __init__(self,
+                 transferred_amount: TokenAmount,
+                 locked_amount: TokenAmount,
+                 locksroot: Locksroot,
+                 partner_transferred_amount: TokenAmount,
+                 partner_locked_amount: TokenAmount,
+                 partner_locksroot: Locksroot,
+                 block_identifier: BlockHash):
+        self.transferred_amount = transferred_amount,
+        self.locked_amount = locked_amount,
+        self.locksroot = locksroot,
+        self.partner_transferred_amount = partner_transferred_amount,
+        self.partner_locked_amount = partner_locked_amount,
+        self.partner_locksroot = partner_locksroot,
+        self.block_identifier = block_identifier
--- a/raiden/api/objects.py
+++ b/raiden/api/objects.py
@@ -71,10 +71,10 @@ class SettlementParameters:
                  partner_locked_amount: TokenAmount,
                  partner_locksroot: Locksroot,
                  block_identifier: BlockHash):
-        self.transferred_amount = transferred_amount,
-        self.locked_amount = locked_amount,
-        self.locksroot = locksroot,
-        self.partner_transferred_amount = partner_transferred_amount,
-        self.partner_locked_amount = partner_locked_amount,
-        self.partner_locksroot = partner_locksroot,
+        self.transferred_amount = transferred_amount
+        self.locked_amount = locked_amount
+        self.locksroot = locksroot
+        self.partner_transferred_amount = partner_transferred_amount
+        self.partner_locked_amount = partner_locked_amount
+        self.partner_locksroot = partner_locksroot
         self.block_identifier = block_identifier
--- a/raiden/api/python.py
+++ b/raiden/api/python.py
@@ -590,12 +590,14 @@ class RaidenAPI:
         Race condition, this can fail if channel was settled externally.
         """
 
+        chain_state = views.state_from_raiden(self.raiden)
+
         channels_to_settle = ChannelValidator.validate_and_get_channels_to_settle(
             token_address=token_address,
             creator_address=creator_address,
             partner_address=partner_address,
             registry_address=registry_address,
-            raiden=self.raiden)
+            chain_state=chain_state)
 
         # get the channel to settle
         channel_iterator = filter(lambda channel:
--- a/raiden/api/python.py
+++ b/raiden/api/python.py
@@ -70,7 +70,7 @@ from raiden.transfer.state import (
     TransferTask,
     ChainState)
 
-from raiden.transfer.state_change import ActionChannelClose, ActionChannelSettle
+from raiden.transfer.state_change import ActionChannelClose
 from raiden.utils import pex, typing
 from raiden.utils.gas_reserve import has_enough_gas_reserve
 from raiden.utils.typing import (
--- a/raiden/api/python.py
+++ b/raiden/api/python.py
@@ -70,7 +70,7 @@ from raiden.transfer.state import (
     TransferTask,
     ChainState)
 
-from raiden.transfer.state_change import ActionChannelClose
+from raiden.transfer.state_change import ActionChannelClose, ActionChannelSettle
 from raiden.utils import pex, typing
 from raiden.utils.gas_reserve import has_enough_gas_reserve
 from raiden.utils.typing import (
@@ -574,6 +574,55 @@ class RaidenAPI:
             )
             raise InsufficientFunds(msg)
 
+    def settle_light(
+        self,
+        registry_address: PaymentNetworkID,
+        token_address: TokenAddress,
+        creator_address: Address,
+        partner_address: Address,
+        signed_settle_tx: typing.SignedTransaction = None,
+    ):
+        """Settle a channel opened with `partner_address` for the given
+        `token_address`.
+        Race condition, this can fail if channel was settled externally.
+        """
+
+        channels_to_settle = ChannelValidator.can_settle_channel(
+            token_address=token_address,
+            partner_addresses=[partner_address],
+            registry_address=registry_address,
+            raiden=self.raiden)
+
+        channel_ids = [channel_state.identifier for channel_state in channels_to_settle]
+
+        chain_state = views.state_from_raiden(self.raiden)
+
+        channel_state = views.get_channelstate_for(
+            chain_state=chain_state,
+            payment_network_id=registry_address,
+            token_address=token_address,
+            creator_address=creator_address,
+            partner_address=partner_address,
+        )
+
+        channel_proxy = self.raiden.chain.payment_channel(
+            canonical_identifier=channel_state.canonical_identifier
+        )
+
+        channel_proxy.settle_channel_light(
+            block_identifier=views.state_from_raiden(self.raiden).block_hash,
+            signed_settle_tx=signed_settle_tx
+        )
+
+        waiting.wait_for_settle(
+            raiden=self.raiden,
+            payment_network_id=registry_address,
+            token_address=token_address,
+            channel_ids=channel_ids,
+            retry_timeout=DEFAULT_RETRY_TIMEOUT,
+            partner_addresses=[partner_address]
+        )
+
     def set_total_channel_deposit_light(
         self,
         registry_address: PaymentNetworkID,
@@ -745,6 +794,27 @@ class RaidenAPI:
             partner_addresses=[partner_address],
             retry_timeout=retry_timeout,
             signed_close_tx=signed_close_tx
+        )
+
+    def channel_settle_light(
+        self,
+        registry_address: PaymentNetworkID,
+        token_address: TokenAddress,
+        creator_address: Address,
+        partner_address: Address,
+        signed_settle_tx: typing.SignedTransaction = None
+    ):
+        """
+            Settle a channel opened with `partner_address` for the given
+            `token_address`.
+            Race condition, this can fail if channel was settled externally.
+        """
+        self.settle_light(
+            registry_address=registry_address,
+            token_address=token_address,
+            creator_address=creator_address,
+            partner_address=partner_address,
+            signed_settle_tx=signed_settle_tx
         )
 
     def channel_batch_close_light(
@@ -1728,11 +1798,11 @@ class RaidenAPI:
                 locked_transfer.message_identifier,
                 locked_transfer,
                 False,
-                payment.payment_id,
                 creator_address,
                 order,
                 LightClientProtocolMessageType.PaymentSuccessful,
-                self.raiden.wal
+                self.raiden.wal,
+                payment.payment_id
             )
             payment_hub_message = PaymentHubMessage(payment_id=payment.payment_id,
                                                     message_order=order,
--- a/raiden/api/python.py
+++ b/raiden/api/python.py
@@ -582,6 +582,7 @@ class RaidenAPI:
         token_address: TokenAddress,
         creator_address: Address,
         partner_address: Address,
+        channel_identifier: typing.ChannelID,
         signed_settle_tx: typing.SignedTransaction = None,
     ):
         """Settle a channel opened with `partner_address` for the given
@@ -591,24 +592,24 @@ class RaidenAPI:
 
         channels_to_settle = ChannelValidator.validate_and_get_channels_to_settle(
             token_address=token_address,
+            creator_address=creator_address,
             partner_address=partner_address,
             registry_address=registry_address,
             raiden=self.raiden)
 
-        channel_ids = [channel_state.identifier for channel_state in channels_to_settle]
+        # get the channel to settle
+        channel_iterator = filter(lambda channel:
+                                  channel.our_state.address == creator_address
+                                  and channel.partner_state.address == partner_address
+                                  and channel.token_address == token_address
+                                  and channel.identifier == channel_identifier, channels_to_settle)
 
-        chain_state = views.state_from_raiden(self.raiden)
+        channel_list = list(channel_iterator)
 
-        channel_state = views.get_channelstate_for(
-            chain_state=chain_state,
-            payment_network_id=registry_address,
-            token_address=token_address,
-            creator_address=creator_address,
-            partner_address=partner_address,
-        )
-
-        if channel_state.identifier not in channel_ids:
+        if len(channel_list) <= 0:
             raise RaidenRecoverableError("Failed trying to settle a channel that's not in waiting_for_settle state")
+
+        channel_state = channel_list[0]
 
         channel_proxy = self.raiden.chain.payment_channel(
             canonical_identifier=channel_state.canonical_identifier
@@ -623,7 +624,7 @@ class RaidenAPI:
             raiden=self.raiden,
             payment_network_id=registry_address,
             token_address=token_address,
-            channel_ids=channel_ids,
+            channel_ids=[channel_state.identifier],
             retry_timeout=DEFAULT_RETRY_TIMEOUT,
             partner_addresses=[partner_address]
         )
@@ -809,6 +810,7 @@ class RaidenAPI:
         token_address: TokenAddress,
         creator_address: Address,
         partner_address: Address,
+        channel_identifier: typing.ChannelID,
         signed_settle_tx: typing.SignedTransaction = None
     ):
         """
@@ -821,6 +823,7 @@ class RaidenAPI:
             token_address=token_address,
             creator_address=creator_address,
             partner_address=partner_address,
+            channel_identifier=channel_identifier,
             signed_settle_tx=signed_settle_tx
         )
 
--- a/raiden/api/python.py
+++ b/raiden/api/python.py
@@ -618,7 +618,7 @@ class RaidenAPI:
         )
 
         channel_proxy.settle_channel_light(
-            block_identifier=views.state_from_raiden(self.raiden).block_hash,
+            block_identifier=chain_state.block_hash,
             signed_settle_tx=signed_settle_tx
         )
 
--- a/raiden/api/python.py
+++ b/raiden/api/python.py
@@ -589,7 +589,7 @@ class RaidenAPI:
 
         channels_to_settle = ChannelValidator.validate_and_get_channels_to_settle(
             token_address=token_address,
-            partner_addresses=[partner_address],
+            partner_address=partner_address,
             registry_address=registry_address,
             raiden=self.raiden)
 
@@ -625,6 +625,8 @@ class RaidenAPI:
             retry_timeout=DEFAULT_RETRY_TIMEOUT,
             partner_addresses=[partner_address]
         )
+
+        return channel_state
 
     def set_total_channel_deposit_light(
         self,
@@ -812,7 +814,7 @@ class RaidenAPI:
             `token_address`.
             Race condition, this can fail if channel was settled externally.
         """
-        self.settle_light(
+        return self.settle_light(
             registry_address=registry_address,
             token_address=token_address,
             creator_address=creator_address,
--- a/raiden/api/python.py
+++ b/raiden/api/python.py
@@ -609,7 +609,22 @@ class RaidenAPI:
         channel_list = list(channel_iterator)
 
         if not channel_list:
-            raise RaidenRecoverableError("Failed trying to settle a channel that's not in waiting_for_settle state")
+            # we check if the channel is already settled, in that case we raise an exception indicating that
+            settled_channels = views.get_channelstate_settled(
+                chain_state=chain_state,
+                payment_network_id=registry_address,
+                token_address=token_address
+            )
+            filtered_settled_channels_iterator = filter(lambda channel:
+                                                        channel.our_state.address == creator_address
+                                                        and channel.partner_state.address == partner_address
+                                                        and channel.token_address == token_address
+                                                        and channel.identifier == channel_identifier, settled_channels)
+            filtered_settled_channels_list = list(filtered_settled_channels_iterator)
+            if filtered_settled_channels_list:
+                raise RaidenRecoverableError("Failed trying to settle a channel that's already settled")
+            else:
+                raise RaidenRecoverableError("Failed trying to settle a channel that's not in waiting_for_settle state")
 
         channel_state = channel_list[0]
 
--- a/raiden/api/python.py
+++ b/raiden/api/python.py
@@ -608,7 +608,7 @@ class RaidenAPI:
 
         channel_list = list(channel_iterator)
 
-        if len(channel_list) <= 0:
+        if not channel_list:
             raise RaidenRecoverableError("Failed trying to settle a channel that's not in waiting_for_settle state")
 
         channel_state = channel_list[0]
--- a/raiden/api/python.py
+++ b/raiden/api/python.py
@@ -587,7 +587,7 @@ class RaidenAPI:
         Race condition, this can fail if channel was settled externally.
         """
 
-        channels_to_settle = ChannelValidator.can_settle_channel(
+        channels_to_settle = ChannelValidator.validate_and_get_channels_to_settle(
             token_address=token_address,
             partner_addresses=[partner_address],
             registry_address=registry_address,
--- a/raiden/api/python.py
+++ b/raiden/api/python.py
@@ -605,6 +605,9 @@ class RaidenAPI:
             partner_address=partner_address,
         )
 
+        if channel_state.identifier not in channel_ids:
+            raise RaidenRecoverableError("Failed trying to settle a channel that's not in waiting_for_settle state")
+
         channel_proxy = self.raiden.chain.payment_channel(
             canonical_identifier=channel_state.canonical_identifier
         )
--- a/raiden/api/rest.py
+++ b/raiden/api/rest.py
@@ -120,7 +120,8 @@ from raiden.transfer.events import (
     EventPaymentSentFailed,
     EventPaymentSentSuccess,
 )
-from raiden.transfer.state import CHANNEL_STATE_CLOSED, CHANNEL_STATE_OPENED, NettingChannelState
+from raiden.transfer.state import CHANNEL_STATE_CLOSED, CHANNEL_STATE_OPENED, NettingChannelState, \
+    CHANNEL_STATE_SETTLING
 from raiden.utils import (
     create_default_identifier,
     optional_address_to_string,
@@ -1633,6 +1634,37 @@ class RestAPI:
         # return api_response(result=result.data)
         return None
 
+    def _settle_light(self,
+                      registry_address: typing.PaymentNetworkID,
+                      channel_state: NettingChannelState,
+                      signed_settle_tx: typing.SignedTransaction):
+        """ This operation validates and send the settlement signed transaction for a LC """
+        log.debug(
+            "Settling light channel",
+            node=pex(self.raiden_api.address),
+            registry_address=to_checksum_address(registry_address),
+            channel_identifier=channel_state.identifier
+        )
+
+        self.validate_channel_status_for_settle(channel_state)
+
+        try:
+            self.raiden_api.channel_settle_light(
+                registry_address=registry_address,
+                token_address=channel_state.token_address,
+                creator_address=channel_state.our_state.address,
+                partner_address=channel_state.partner_state.address,
+                signed_settle_tx=signed_settle_tx
+            )
+        except InsufficientFunds as e:
+            return api_error(errors=str(e), status_code=HTTPStatus.PAYMENT_REQUIRED)
+        except RawTransactionFailed as e:
+            return ApiErrorBuilder.build_and_log_error(errors=str(e), status_code=HTTPStatus.BAD_REQUEST, log=log)
+        except RaidenRecoverableError as e:
+            return ApiErrorBuilder.build_and_log_error(errors=str(e), status_code=HTTPStatus.BAD_REQUEST, log=log)
+
+        return self.update_channel_state(registry_address, channel_state)
+
     def _deposit_light(
         self,
         registry_address: typing.PaymentNetworkID,
@@ -1778,6 +1810,15 @@ class RestAPI:
         return self.update_channel_state(registry_address, channel_state)
 
     @staticmethod
+    def validate_channel_status_for_settle(channel_state):
+        if channel.get_status(channel_state) != CHANNEL_STATE_SETTLING:
+            return api_error(
+                errors="Attempted to settle a channel that is not in waiting_for_settle state",
+                status_code=HTTPStatus.CONFLICT,
+            )
+        return None
+
+    @staticmethod
     def validate_channel_status(channel_state):
         if channel.get_status(channel_state) != CHANNEL_STATE_OPENED:
             return api_error(
@@ -1807,9 +1848,9 @@ class RestAPI:
         signed_approval_tx: typing.SignedTransaction,
         signed_deposit_tx: typing.SignedTransaction,
         signed_close_tx: typing.SignedTransaction,
+        signed_settle_tx: typing.SignedTransaction,
         total_deposit: typing.TokenAmount = None,
-        state: str = None,
-
+        state: str = None
     ):
         log.debug(
             "Patching light channel",
@@ -1865,6 +1906,8 @@ class RestAPI:
 
         elif state == CHANNEL_STATE_CLOSED:
             result = self._close_light(registry_address, channel_state, signed_close_tx)
+        elif state == CHANNEL_STATE_SETTLING:
+            result = self._settle_light(registry_address, channel_state, signed_settle_tx)
         else:  # should never happen, channel_state is validated in the schema
             result = api_error(
                 errors="Provided invalid channel state {}".format(state),
--- a/raiden/api/rest.py
+++ b/raiden/api/rest.py
@@ -121,8 +121,7 @@ from raiden.transfer.events import (
     EventPaymentSentFailed,
     EventPaymentSentSuccess,
 )
-from raiden.transfer.state import CHANNEL_STATE_CLOSED, CHANNEL_STATE_OPENED, NettingChannelState, \
-    CHANNEL_STATE_SETTLING
+from raiden.transfer.state import CHANNEL_STATE_CLOSED, CHANNEL_STATE_OPENED, NettingChannelState
 from raiden.utils import (
     create_default_identifier,
     optional_address_to_string,
--- a/raiden/api/rest.py
+++ b/raiden/api/rest.py
@@ -86,7 +86,8 @@ from raiden.api.v1.resources import (
     LightClientMatrixCredentialsBuildResource,
     LightClientResource,
     PaymentLightResource,
-    CreatePaymentLightResource, WatchtowerResource, LightClientMessageResource)
+    CreatePaymentLightResource, WatchtowerResource, LightClientMessageResource,
+    SettlementLightResourceByTokenAndPartnerAddress)
 
 from raiden.constants import GENESIS_BLOCK_NUMBER, UINT256_MAX, Environment, EMPTY_PAYMENT_HASH_INVOICE
 
@@ -228,6 +229,10 @@ URLS_HUB_V1 = [
     (
         "/light_channels/<hexaddress:token_address>/<hexaddress:creator_address>/<hexaddress:partner_address>",
         LightChannelsResourceByTokenAndPartnerAddress
+    ),
+    (
+        "/light_channels/<hexaddress:token_address>/<hexaddress:creator_address>/<hexaddress:partner_address>/settle",
+        SettlementLightResourceByTokenAndPartnerAddress
     ),
     ("/payments_light", PaymentLightResource),
     ("/light_client_messages", LightClientMessageResource, "Message polling"),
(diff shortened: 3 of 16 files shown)


raiden/api/python.py
```diff
@@ -1728,11 +1801,11 @@ def create_light_client_payment(
                 locked_transfer.message_identifier,
                 locked_transfer,
                 False,
-                payment.payment_id,
                 creator_address,
                 order,
                 LightClientProtocolMessageType.PaymentSuccessful,
-                self.raiden.wal
+                self.raiden.wal,
+                payment.payment_id
```
marcosmartinez7, line 1878: i dont understand the addition of payment.payment_id parameter and how is this related with settlement.

raiden/lightclient/handlers/light_client_message_handler.py
```diff
@@ -43,9 +43,10 @@ class LightClientMessageHandler:
     log = structlog.get_logger(__name__)  # pylint: disable=invalid-name
 
     @classmethod
-    def store_light_client_protocol_message(cls, identifier: int, message: Message, signed: bool, payment_id: int,
+    def store_light_client_protocol_message(cls, identifier: int, message: Message, signed: bool,
```
marcosmartinez7, line 46: same here, this changes are apearing here because this branch is behind develop?

raiden/api/python.py
JONAF2103: It's not, but we forgot to do that in the refactor of the database. Basically we had the payment id defined as an integer in the middle and when you put None for those cases where you have to do that since is not a payment related message, it shows you a warning, so to solve this you need to put payment_id as a default parameter = None, but to do that you need to move it to the end. I can remove this change if it's a problem, i just wanted to keep the code clean that's all.

raiden/lightclient/handlers/light_client_message_handler.py
JONAF2103: Same as above

raiden/api/python.py
mauricioirace: It makes sense to me.

raiden/api/objects.py
```diff
@@ -56,3 +59,22 @@ class DashboardGeneralItem:
     quantity = 0
     event_type_code = 0
     event_type_class_name = ""
+
+
+class SettlementParameters:
```
mauricioirace, line 64: Is this class used somewhere?

raiden/api/validations/channel_validator.py
```diff
@@ -118,3 +118,31 @@ def can_close_channel(token_address,
         )
 
         return channels_to_close
+
+    @staticmethod
+    def can_settle_channel(token_address,
```
mauricioirace: I don't like the idea of a function called can_X that doesn't return a boolean. Maybe get_channels_to_settle or something like would make more sense in my opinion.

raiden/messages.py
```diff
@@ -1832,6 +1832,78 @@ def verify_request_monitoring(
...
+        packed.participant1_locked_amount = self.participant1_locked_amount
+        packed.participant1_locksroot = self.participant1_locksroot
+        packed.participant2 = self.participant2
+        packed.participant2_transferred_amount = self.participant2_transferred_amount
+        packed.participant2_locked_amount = self.participant2_locked_amount
+        packed.participant2_locksroot = self.participant2_locksroot
+
+    def to_dict(self) -> Dict[str, Any]:
+        return {
+            "channel_identifier": self.channel_identifier,
```
mauricioirace, line 1882: "type" property  is missing

raiden/network/proxies/token_network.py
```diff
@@ -2255,6 +2255,66 @@ def settle(
 
         log.info("settle successful", **log_details)
 
+    def settle_light(
```
mauricioirace, line 2244: Couldn't settle and settle_light be generalized? They have the same base structure.

raiden/transfer/views.py
```diff
@@ -638,6 +638,41 @@ def filter_channels_by_partneraddress(
...
+    # for partner in partner_addresses:
+
+    for node_address in token_network.partneraddresses_to_channelidentifiers.keys():
+        if node_address in token_network.channelidentifiers_to_channels:
+            channels = token_network.channelidentifiers_to_channels[node_address]
+            if channels is not None:
+                for _channelId, channel in channels.items():
+                    for partner_address in partner_addresses:
+                        if channel.partner_state.address == partner_address:
+                            if channel.close_transaction and channel.close_transaction.result == 'success':
```
mauricioirace: I think this could be  refactorized in order to prevent such amount of identation levels.

mauricioirace (changes requested):
Some minor changes I would make

raiden/api/rest.py
```diff
@@ -1633,6 +1634,37 @@ def initiate_payment_light(
...
+                token_address=channel_state.token_address,
+                creator_address=channel_state.our_state.address,
+                partner_address=channel_state.partner_state.address,
+                signed_settle_tx=signed_settle_tx
+            )
+        except InsufficientFunds as e:
+            return api_error(errors=str(e), status_code=HTTPStatus.PAYMENT_REQUIRED)
+        except RawTransactionFailed as e:
+            return ApiErrorBuilder.build_and_log_error(errors=str(e), status_code=HTTPStatus.BAD_REQUEST, log=log)
+        except RaidenRecoverableError as e:
```
mauricioirace:
```suggestion
        except (RawTransactionFailed, RaidenRecoverableError) as e:
```

raiden/api/objects.py
JONAF2103: Yes to wrap the result parameters of the function called process_data_and_get_settlement_parameters in RaidenEventHandler, that function is used to generate the data for settlement.

raiden/api/validations/channel_validator.py
JONAF2103: I know, but this function validates and also returns the channels, basically is a code that was there for close too, i think we can change that name to something like validate_and_get_channels_to_settle instead.

raiden/messages.py
JONAF2103: What do you mean, i get lost here

raiden/transfer/views.py
JONAF2103: Probably but we have it in other places so i didn't spend too much time thinking about it

raiden/network/proxies/token_network.py
JONAF2103: You mean the error handling maybe? the other parts are different.

raiden/api/rest.py
JONAF2103: I extracted this from the current settle function, i will take a look at it

raiden/messages.py
mauricioirace: The other messages have this line in their to_dict implementation: `type": self.__class__.__name__,`. If i didn't missunderstand its meaning, that propety is the one used later by the light client to decide what kind of message it is, and what should it do in response to it.

raiden/network/proxies/token_network.py
mauricioirace: yes. Anyway, it's true it may be a bit too much time expensive right now.

raiden/messages.py
JONAF2103: Well good catch there, but is not used by the Light Client, it's used by the node only to verify that the message is correctly deserialized and serialized from database. The LC uses the LightClientProtocolMessageType to handle the message. Anyway i will add that

raiden/network/proxies/token_network.py
JONAF2103: Not at all, i will try to simplify this code and make it more general

raiden/transfer/views.py
JONAF2103: We can change that into this:

```python
def filter_channels_to_settle_by_partner_address(
    chain_state: ChainState,
    token_address: TokenAddress,
    partner_addresses: List[Address],
) -> List[NettingChannelState]:
    token_network = get_token_network_by_token_address(
        chain_state, payment_network_id, token_address
    )

    result: List[NettingChannelState] = []
    if not token_network:
        return result

    node_addresses = filter(lambda node_address: node_address in token_network.channelidentifiers_to_channels,
                            token_network.partneraddresses_to_channelidentifiers.keys())
    channels = filter(lambda channels: channels is not None,
                      map(lambda node_address: token_network.channelidentifiers_to_channels[node_address],
                          node_addresses))

    channels_result = map(lambda partner_address: filter(lambda _channelId, channel: channel.partner_state.address == partner_address and
                                                         channel.close_transaction and
                                                         channel.close_transaction.result == 'success',
                             channels.items()), partner_addresses)

    states = filter_channels_by_status(channels_result, [CHANNEL_STATE_SETTLING])
    # If multiple channel states are found, return the last one.
    if states:
        result.append(states[-1])

    return result
```
Tell me what do you think about it, maybe we can analize to change that into lambda expresions like this

raiden/api/python.py
```diff
@@ -574,6 +574,58 @@ def _set_total_deposit_preconditions(
...
+        self,
+        registry_address: PaymentNetworkID,
+        token_address: TokenAddress,
+        creator_address: Address,
+        partner_address: Address,
+        signed_settle_tx: typing.SignedTransaction = None,
+    ):
+        """Settle a channel opened with `partner_address` for the given
+        `token_address`.
+        Race condition, this can fail if channel was settled externally.
```
mortelli, line 590: are we doing anything about this race condition? what's our policy on this?

raiden/api/python.py
```diff
@@ -574,6 +574,58 @@ def _set_total_deposit_preconditions(
...
+
+        if channel_state.identifier not in channel_ids:
+            raise RaidenRecoverableError("Failed trying to settle a channel that's not in waiting_for_settle state")
+
+        channel_proxy = self.raiden.chain.payment_channel(
+            canonical_identifier=channel_state.canonical_identifier
+        )
+
+        channel_proxy.settle_channel_light(
+            block_identifier=views.state_from_raiden(self.raiden).block_hash,
```
mortelli: by this point, can this value differ from the one obtained on line `598`?

raiden/api/python.py
```diff
@@ -574,6 +574,58 @@ def _set_total_deposit_preconditions(
...
+            signed_settle_tx=signed_settle_tx
+        )
+
+        waiting.wait_for_settle(
+            raiden=self.raiden,
+            payment_network_id=registry_address,
+            token_address=token_address,
+            channel_ids=channel_ids,
+            retry_timeout=DEFAULT_RETRY_TIMEOUT,
```
mortelli, line 645: couldn't this ever have a different value other than the default? (a param, maybe?)

raiden/api/python.py
```diff
@@ -747,6 +799,27 @@ def channel_close_light(
             signed_close_tx=signed_close_tx
         )
 
+    def channel_settle_light(
```
mortelli, line 824: what is the point of this function, if it just calls `self.settle_light`?

raiden/api/rest.py
```diff
@@ -1777,6 +1809,15 @@ def _close(
 
         return self.update_channel_state(registry_address, channel_state)
 
+    @staticmethod
+    def validate_channel_status_for_settle(channel_state):
```
mortelli: we could add type hints here, no?

raiden/api/rest.py
```diff
@@ -1633,6 +1634,37 @@ def initiate_payment_light(
...
+                      signed_settle_tx: typing.SignedTransaction):
+        """ This operation validates and send the settlement signed transaction for a LC """
+        log.debug(
+            "Settling light channel",
+            node=pex(self.raiden_api.address),
+            registry_address=to_checksum_address(registry_address),
+            channel_identifier=channel_state.identifier
+        )
+
+        self.validate_channel_status_for_settle(channel_state)
```
mortelli: you're not doing anything with the result here

raiden/api/rest.py
```diff
@@ -1777,6 +1809,15 @@ def _close(
 
         return self.update_channel_state(registry_address, channel_state)
 
+    @staticmethod
+    def validate_channel_status_for_settle(channel_state):
+        if channel.get_status(channel_state) != CHANNEL_STATE_SETTLING:
+            return api_error(
+                errors="Attempted to settle a channel that is not in waiting_for_settle state",
```
mortelli: you can use the `CHANNEL_STATE_SETTLING` const here

raiden/api/python.py
```diff
@@ -574,6 +574,58 @@ def _set_total_deposit_preconditions(
...
+        channel_state = views.get_channelstate_for(
+            chain_state=chain_state,
+            payment_network_id=registry_address,
+            token_address=token_address,
+            creator_address=creator_address,
+            partner_address=partner_address,
+        )
+
+        if channel_state.identifier not in channel_ids:
+            raise RaidenRecoverableError("Failed trying to settle a channel that's not in waiting_for_settle state")
```
mortelli: i am somewhat confused by this block of code, considering the existence of the `validate_channel_status_for_settle` function.

any thoughts on this?

raiden/api/validations/channel_validator.py
```diff
@@ -118,3 +118,31 @@ def can_close_channel(token_address,
...
+    def validate_and_get_channels_to_settle(token_address,
+                                            partner_addresses,
+                                            registry_address,
+                                            raiden):
+
+        if not is_binary_address(token_address):
+            raise InvalidAddress("Expected binary address format for token in channel settle")
+
+        if not all(map(is_binary_address, partner_addresses)):
+            raise InvalidAddress("Expected binary address format for partner in channel settle")
```
mortelli: very nice use of `all` and `map` here!

i would just update the error string to use words in plural, since we are inspecting all received partner addresses

raiden/api/validations/channel_validator.py
```diff
@@ -118,3 +118,31 @@ def can_close_channel(token_address,
         )
 
         return channels_to_close
+
+    @staticmethod
+    def validate_and_get_channels_to_settle(token_address,
```
mortelli: it would be nice to have type hints here also

raiden/api/validations/channel_validator.py
```diff
@@ -118,3 +118,31 @@ def can_close_channel(token_address,
...
+                                            partner_addresses,
+                                            registry_address,
+                                            raiden):
```
mortelli: let's not pass an entire `raiden` variable here if we can avoid it; i think we can get away with using a `ChainState`-typed param instead.

raiden/api/validations/channel_validator.py
```diff
@@ -118,3 +118,31 @@ def can_close_channel(token_address,
...
+
+        chain_state = views.state_from_raiden(raiden)
+        channels_to_settle = views.filter_channels_to_settle_by_partner_address(
+            chain_state=chain_state,
+            payment_network_id=registry_address,
+            token_address=token_address,
+            partner_addresses=partner_addresses,
+        )
+
+        return channels_to_settle
```
mortelli:
```suggestion
        return views.filter_channels_to_settle_by_partner_address(
            chain_state=chain_state,
            payment_network_id=registry_address,
            token_address=token_address,
            partner_addresses=partner_addresses,
        )
```

raiden/lightclient/handlers/light_client_message_handler.py
```diff
@@ -97,6 +98,18 @@ def is_light_client_protocol_message_already_stored(cls, payment_id: int,
...
+    def is_message_already_stored(cls,
+                                  light_client_address: AddressHex,
+                                  message_type: LightClientProtocolMessageType,
+                                  unsigned_message: Message,
+                                  wal: WriteAheadLog):
+
+        return message_type and light_client_address and unsigned_message \
+                            and wal.storage.is_message_already_stored(light_client_address,
+                                                                      message_type.value,
+                                                                      unsigned_message) is not None
```
mortelli: do we need the `is not None` here? why/why not?

raiden/network/proxies/token_network.py
```diff
@@ -2227,19 +2227,83 @@ def settle(
...
-            if transaction_executed:
-                block = receipt_or_none["blockNumber"]
+        self.validate_settlement_transaction_error(channel_identifier,
+                                                   checking_block,
+                                                   error_prefix,
+                                                   log_details,
+                                                   partner,
+                                                   transaction_error,
+                                                   False,
+                                                   gas_limit)
```
mortelli: minor remark: named parameters here would help with readability, i think

raiden/network/proxies/token_network.py
```diff
@@ -2227,19 +2227,83 @@ def settle(
...
+    def validate_settlement_transaction_error(self,
+                                              channel_identifier,
+                                              checking_block,
+                                              error_prefix,
+                                              log_details,
+                                              partner,
+                                              transaction_error,
+                                              is_light,
+                                              gas_limit=None):
+        transaction_executed = is_light or gas_limit is not None
```
mortelli: i need some help understanding this line :thinking: 

how can we be sure that a transaction was executed just due to it being LC-related? same question with the gas limit.

i appreciate any light shed on this

raiden/network/proxies/token_network.py
```diff
@@ -2227,19 +2227,83 @@ def settle(
...
+        if not transaction_executed or transaction_error:
```
mortelli: be very careful with operator precedence here! `not` is evaluated before `or`; make sure this condition is doing what you want it to

raiden/transfer/channel.py
```diff
@@ -26,7 +26,7 @@
     EventInvalidReceivedTransferRefund,
     EventInvalidReceivedUnlock,
     SendProcessed,
-    ContractSendChannelUpdateTransferLight)
+    ContractSendChannelUpdateTransferLight, ContractSendChannelSettleLight)
```
mortelli: i'd opt for adding this to line `30` instead, for consistency's sake

raiden/transfer/channel.py
```diff
@@ -1825,12 +1825,22 @@ def handle_block(
...
             )
-            event = ContractSendChannelSettle(
-                canonical_identifier=channel_state.canonical_identifier,
-                triggered_by_block_hash=state_change.block_hash,
-                channel_state= channel_state
-            )
-            events.append(event)
+
+            if channel_state.is_light_channel:
+                # get which of the participants is the lc to differentiate the address
```
mortelli: could you please explain what this comment is getting at?

raiden/transfer/channel.py
```diff
@@ -1825,12 +1825,22 @@ def handle_block(
...
+                    channel_state=channel_state
+                )
+                events.append(event)
+            else:
+                event = ContractSendChannelSettle(
+                    canonical_identifier=channel_state.canonical_identifier,
+                    triggered_by_block_hash=state_change.block_hash,
+                    channel_state=channel_state
+                )
+                events.append(event)
```
mortelli:
```suggestion
            if channel_state.is_light_channel:
                # get which of the participants is the lc to differentiate the address
                event = ContractSendChannelSettleLight(
                    canonical_identifier=channel_state.canonical_identifier,
                    triggered_by_block_hash=state_change.block_hash,
                    channel_state=channel_state
                )
            else:
                event = ContractSendChannelSettle(
                    canonical_identifier=channel_state.canonical_identifier,
                    triggered_by_block_hash=state_change.block_hash,
                    channel_state=channel_state
                )
            events.append(event)
```

raiden/transfer/views.py
```diff
@@ -638,6 +638,41 @@ def filter_channels_by_partneraddress(
...
+                                channelsResult.append(channel)
+
+    states = filter_channels_by_status(channelsResult, [CHANNEL_STATE_SETTLING])
+    # If multiple channel states are found, return the last one.
+    if states:
+        result.append(states[-1])
+
+    return result
+
+
```
mortelli: leaving the impressive amount of nested statements here aside, we do not need to repeat all of this code from the `filter_channels_by_partneraddress` method (the difference between these two is just 2 lines!)

we can instead extract the functionality of filtering channels by partner addresses, and then applying whatever status filter we need.

if this is too big a task for this PR, please create a separate issue for it and let's tackle it later.

mortelli: while we're comparing these 2: the status filter difference is clear, but what about the boolean difference?

as i understand it:
- `filter_channels_by_partneraddress` verifies that either:
  - the channel was not closed at all, or 
  - the channel was not closed successfully.
- `filter_channels_to_settle_by_partner_address` verifies that the channel was closed successfully.

these are opposites, right? is this the intended behavior?

raiden/raiden_event_handler.py
```diff
@@ -29,7 +32,7 @@
     EventPaymentSentFailed,
     EventPaymentSentSuccess,
     SendProcessed,
-    ContractSendChannelUpdateTransferLight)
+    ContractSendChannelUpdateTransferLight, ContractSendChannelSettleLight)
```
mortelli: there should be a newline here

raiden/raiden_event_handler.py
```diff
@@ -603,104 +611,215 @@ def handle_contract_send_channelunlock(
...
+            channel_settle_event.triggered_by_block_hash)
+
+        canonical_identifier = CanonicalIdentifier(
+            chain_identifier=raiden.chain.network_id,
+            token_network_address=channel_settle_event.token_network_identifier,
+            channel_identifier=channel_settle_event.channel_identifier,
+        )
+        payment_channel: PaymentChannel = raiden.chain.payment_channel(
+            canonical_identifier=canonical_identifier
+        )
```
mortelli: either is fine, but we should be consistent and use only 1 style for closing parentheses (either in-line or in a newline with the outer indentation), at least within the same function

raiden/raiden_event_handler.py
```diff
@@ -603,104 +611,215 @@ def handle_contract_send_channelunlock(
...
+
+    @staticmethod
+    def process_data_and_get_settlement_parameters(raiden: "RaidenService",
+                                                   token_network_identifier: TokenNetworkID,
+                                                   channel_identifier: ChannelID,
+                                                   triggered_by_block_hash: BlockHash) -> SettlementParameters:
+
+        log.debug("Processing settlement data")
+
+        assert raiden.wal, "The Raiden Service must be initialize to handle events"
```
mortelli: typo in `initialized`

raiden/api/python.py
marcosmartinez7: I think just for consistency with the other light client functions

raiden/api/validations/channel_validator.py
marcosmartinez7: Agree, we can just pass the ChainState here

raiden/network/proxies/token_network.py
marcosmartinez7: I have the same question here, if light_client is true then the transaction executed? the tx for light clients can fail too

raiden/api/python.py
JONAF2103: We are throwing an error on the api response, the light client will handle this case

raiden/api/python.py
JONAF2103: I think the code changed and this comment is not valid anymore

raiden/api/python.py
JONAF2103: wait_for_settle has retry_timeout as a required parameter, but it doesn't make sense to put a param here to let this be changed from outside since is not going to change for this case

raiden/api/python.py
JONAF2103: Yes just to keep consistency on the code that's all

raiden/api/validations/channel_validator.py
JONAF2103: The other cases has raiden too, i just based this method on the other methods on this file, are we going to start changing this now? i mean i can do it but i don't see the advantage since we have other methods on same file with the same structure using raiden as a parameter

raiden/lightclient/handlers/light_client_message_handler.py
JONAF2103: probably not i just used what was there, but i'm not sure if is_message_already_stored returns None or an empty list when the message doesn't exists

raiden/network/proxies/token_network.py
JONAF2103: agree

raiden/network/proxies/token_network.py
JONAF2103: This has been made as a requested change for generalize this, the original code was worse than this, basically if light_client is true i don't have a gas limit to check so i can asume that the transaction was executed but, if you see the next condition you will understand that you can have a transaction_error so basically it will control errors for both cases, i know the conditions are not the best but it's the only way to generalize this error handling, otherwise we need to have the old code here

raiden/network/proxies/token_network.py
JONAF2103: it does

raiden/transfer/channel.py
JONAF2103: it doesn't apply anymore, i can remove it

raiden/lightclient/handlers/light_client_message_handler.py
JONAF2103: confirmed, the result is None so we can remove that, here is a reference https://pynative.com/python-cursor-fetchall-fetchmany-fetchone-to-read-rows-from-table/

raiden/lightclient/handlers/light_client_message_handler.py
```diff
@@ -99,6 +100,18 @@ def is_light_client_protocol_message_already_stored(cls, payment_id: int,
                                               existing_message[7])
         return existing_message
 
+    @classmethod
+    def is_message_already_stored(cls,
```
marcosmartinez7, line 104: Why was this method necessary to implement the settlement?

raiden/transfer/views.py
marcosmartinez7: Why it was needed to create a new function instead of adding a parameter to `filter_channels_by_partneraddress` to check by channel status and txs status?

raiden/api/python.py
mortelli: `python.py:593`:
```python
chain_state = views.state_from_raiden(self.raiden)
```
`python.py:621`:
```python
block_identifier=views.state_from_raiden(self.raiden).block_hash,
```

raiden/api/rest.py
mortelli: this seems to still be true, both for `_close` as well as `_close_light`

raiden/api/rest.py
mortelli: this seems to be only called for channel closing procedures, so the function name change (`validate_channel_status_for_settle` → `validate_channel_status`) is not really reflective of the behavior for this method; specially considering it is static and could be called arbitrarily.

in addition to type hints, i would keep the previous name for clarity's sake.

raiden/api/python.py
mortelli: this error message is not reflective of the filter previously applied

raiden/api/python.py
JONAF2103: Ok got it it was different from now, thanks for the especification

raiden/api/validations/channel_validator.py
mortelli: thanks for adding these; could you also add a return type please?

raiden/api/validations/channel_validator.py
mortelli: careful, you're returning `channels` now, instead of `channels_to_settle`

raiden/lightclient/handlers/light_client_message_handler.py
JONAF2103: It's not just for settlement, before this you had duplicated messages since we don't use the payment id anymore, this method checks the type, message content and lc address to ensure that we don't have other message already created, this basically is for when you have the hub down and the you start it again and then it runs all over again and re-creates the message for the LC

raiden/network/proxies/token_network.py
mortelli: i understand we are trying to make an improvement, but i'm still having trouble understanding this code.

i see this function is called both in `settle` as in `settle_light`. this seems logical.

however, the name is confusing... what does it mean to _validate an error_? could you please explain what this function aims to do, business logic-wise?

also, some of these parameters don't make sense. the body of the function depends on the conditions of `is_light`, `gas_limit is not None`, and `transaction_error`. if these conditions aren't met, the function will just do nothing.

i suggest we use these conditions to call the method in the first place, and re-word what we are trying to do here.

raiden/lightclient/handlers/light_client_message_handler.py
marcosmartinez7: Ok i have created an issue RLP791 to refactor the method names later.

raiden/api/rest.py
JONAF2103: Don't follow you here, this is another function, if you see the code you will realize that we have several functions called validate_channel_status_for... i just used the function that was already there, so i don't know what you mean here.

JONAF2103: This function is not there anymore

raiden/transfer/views.py
JONAF2103: This doesn't apply anymore guys

raiden/api/python.py
JONAF2103: Done

raiden/transfer/channel.py
JONAF2103: Removed comment since it doesn't apply anymore

raiden/raiden_event_handler.py
JONAF2103: This was a fragment from somewhere else, good catch we should fix it all over the place

JONAF2103: Done

raiden/raiden_event_handler.py
JONAF2103: Done

raiden/raiden_event_handler.py
JONAF2103: Done changed to be consistent on the same function

raiden/transfer/views.py
JONAF2103: Finally we removed this so

mauricioirace: I like it.

Anyway if the code is in other places, I think it's not urgent to modifiy it, but it would be nice.

raiden/api/validations/channel_validator.py
JONAF2103: Done

raiden/api/python.py
```diff
@@ -576,6 +576,63 @@ def _set_total_deposit_preconditions(
...
+        # get the channel to settle
+        channel_iterator = filter(lambda channel:
+                                  channel.our_state.address == creator_address
+                                  and channel.partner_state.address == partner_address
+                                  and channel.token_address == token_address
+                                  and channel.identifier == channel_identifier, channels_to_settle)
+
+        channel_list = list(channel_iterator)
+
+        if len(channel_list) <= 0:
```
marcosmartinez7: You can change this to `if channel_list`

raiden/raiden_event_handler.py
```diff
@@ -603,104 +612,217 @@ def handle_contract_send_channelunlock(
...
+        )
+        payment_channel: PaymentChannel = raiden.chain.payment_channel(
+            canonical_identifier=canonical_identifier
+        )
+
+        pseudo_random_generator = random.Random()
+
+        message_identifier = message_identifier_from_prng(pseudo_random_generator)
+
+        # and now find out
```
marcosmartinez7: incomplete comment?

raiden/api/validations/channel_validator.py
```diff
@@ -118,3 +119,37 @@ def can_close_channel(token_address,
...
+            creator_address=creator_address
+        )
+
+        channels_to_settle: List[NettingChannelState] = []
+
+        for channel in channels:
+            if channel.partner_state.address == partner_address:
+                channels_to_settle.append(channel)
+
+        return channels
```
marcosmartinez7: Be careful here, youre appending to `channels_to_settle `and returning `channels`

marcosmartinez7 (approved):
Lgtm, just fix the incomplete comment

raiden/api/python.py
JONAF2103: Good to know

raiden/raiden_event_handler.py
JONAF2103: Exctracted code comment, doesn't apply anymore, but i will check just in case

raiden/api/python.py
JONAF2103: Should be ```if not channel_list``` right?

marcosmartinez7: Yeah sorry my bad

mortelli (approved):
LGTM, untested

raiden/raiden_event_handler.py
JONAF2103: Done

raiden/api/python.py
JONAF2103: Done

mauricioirace (approved):

PR #119, opened 2020-09-14
RLP-715: Same hub LC to LC payments refactor

marcosmartinez7:
### Description

This PR aims to resolve the issue with [making payments between two light clients that are registered at the same hub](https://jirainfuy.atlassian.net/browse/RLP-715). 

### Problem:

The `PaymentMappingState.secrethashes_to_task` dict has the `SecretHash` type as key. See https://github.com/rsksmart/lumino/blob/86327115cc5121d052f6077a477a043c0b03dfbf/raiden/transfer/state.py#L727

Supose the following scenario: 

- Node A sends a `LockedTransfer `to node B
- A payment task (`InitiatorPaymentTask`) is created for the node A, that task has a `secrethash=s1`
- Node B receives the `LockedTransfer`, the payment is dispatched to a payment task, in this case, to a target payment task. 
- The correct behaviour should be that the Node B does not find any payment task related with `s1`, thats because the payment task is created after the `LockedTransfer `handling 
- For light payments, since the `PaymentMappingState `is shared and used by any light payment task, the target state machine will find the initiator payment tasks related with `s1 `when receiving the `LockedTransfer`. 

That behaviour inst correct, because we need to differentiate the `InitiatorPaymentTasks `from `TargetPaymentTasks`. Each task has a different handling depending on the role (`initiator, mediator, target`). 

There are another problems that were fund later:

- We must detect if a channel is between two light clients. Thats because the messages for payments on that channel must be received by the two light clients.
- The light_client_protocol_message table has only one reference to client and must be modified in order to have two possible references to light clients. 
- 

### Alternative 1

Modify the key of the  `PaymentMappingState.secrethashes_to_task`  dict to use a `Tuple[Address, SecretHash]`, therefore each payment task is related with an address too. 

For example, for a normal p2p payment with `secrethash=s1`, we will have: 

```
PaymentMappingState{
    (0x1,s1) : list of PaymentTasks // initiator
    (0x2,s1) : list of PaymentTasks // target 
}
```

### Refactor affected code

- utils (migrations)
- raiden_service
- api/python
- node
- views
- transfer/state
- resolver/client
- storage/migrations
- test codebase

#### resolver/client

resolver/client/reveal_secret_with_resolver: used by Message_handler#handle_send_secretrequest. We must use the default raiden address.

#### views

`def get_transfer_role(chain_state: ChainState, secrethash: SecretHash) -> Optional[str]:
 `

Used at: message_handler#handle_message_refund_transfer. We must extract the address from the RefundTransfer message. 

`def get_transfer_secret(chain_state: ChainState, secrethash: SecretHash) -> Optional[Secret]:
`
Used at: message_handler#handle_message_refund_transfer. We must extract the address from the RefundTransfer message. 

`def get_transfer_task(chain_state: ChainState, secrethash: SecretHash) -> Optional[TransferTask]:
`
Unused

`def get_all_transfer_tasks(chain_state: ChainState) -> Dict[SecretHash, TransferTask]:
`
Used at: get pending transfers endpoint

#### node 

`def subdispatch_to_all_lockedtransfers(
 `

- the for loop key must be modified
- Used at handle_block

`def subdispatch_to_paymenttask
`
This method gets the payment task only by hash, we need to know for wich address. 

It also deletes a payment task only by secret hash.

Used at:
image

`def subdispatch_initiatortask(
`
This method gets the payment task only by hash, we need to know for wich address. 
It also appends a payment task by the secret hash

Used at: 
image

`subdispatch_mediatortask`

Same as subdispatch_initiatortask.

Used at:

image

`subdispatch_mediatortask`

Same as subdispatch_initiatortask.

Used at:
image

`handle_node_change_network_state
`
This function invokes the subdispatch_mediatortask by secrethash.

Used on ActionChangeNodeNetworkState state change for when transport#_address_reachability_changed executes.

`handle_receive_transfer_refund_cancel_route`

This method finds the tasks to create the new task cloning the initial payment task.

#### api/python

`create_light_client_payment
`
This function updates the payment task for refund transfers. It must provides the lc address for accessing to the mapping.

### Alternative 2

Modify the ChainState.payment_mapping to properly be a mapping. 

Now:         `self.payment_mapping = PaymentMappingState()`
After refactor:         `self.payment_mapping = Dict[Address,PaymentMappingState]`

Affected code looks pretty much the same as alternative # 1

## Selected alternative 

Alternative 2. Reasons: 
- Only affects layer 1 of the mapping (less code to refactor)
- cleaner: each node has a dict of tasks, tasks for different nodes are not in the same level of the map

### Light client protocol messages modifications

Now, each message is associated with a sender and a receiver.

For payments between a light client and a full node, sender or receiver will be null. 

For payments between two light clients, both lc addresses will act as sender and receiver.

The 14 messages are shared between the two light clients. 

### Channels modification

- added both_participants_are_light_clients to know if the channel is between two light clients registered on that hub. 

### channelidentifiers_to_channels modification

The structure of this mapping was: 

```
hub_address (0x123) : list of channels
handled_lc_1 (0x1) : list of channels
handled_lc_2 (0x2): list of channels
.
.
.
handled_lc_n: list_of_channels
```

The problem with that structure comes when we must store channels that are between two handled light clients into the same hub. Basically if we have a channel between 0x1 and 0x2 we only store the 0x1 key. 

Suppose that we send a payment from 0x1 to 0x2, the target, 0x2, must modifiy the balance proof received on our_side of the channel, but, the channel isnt present on the mapping using the 0x2 key. 

Therefore, we decided to put the channel associated with both keys. 

## Acceptance criteria

Full node:
- [ ] p2p payments should work
- [ ] mediated payments should work
- [ ] refund transfers should work
- [ ] expiration should work (both for p2p payments and mediated) 

Hub and LC: 
- [ ] p2p payments should work
- [ ] mediated payments should work
- [ ] refund transfers should work
- [ ] expiration should work (both for p2p payments and mediated) 
- [ ] each of the above scenarios must work into the same hub and into different hubs

Tests:
- [ ] maintain the same level of test coverage and results

#### All changes can be reviewed in the commit [here](https://github.com/rsksmart/lumino/pull/119/commits/9d57e3511ed6bdc9e08c217ed0c7c186e6699015).

- **raiden/api/python.py**:

  - Adding named parameters for LightClientMessageHandler.store_light_client_protocol_message call and LightClientPayment init call

- **raiden/api/rest.py**:

  - Updating calls to the new refactored method for LightClientMessageHandler.get_light_client_protocol_message_by_internal_identifier

- **raiden/api/v1/resources.py**:

  - Fixing a spacing warning

- **raiden/lightclient/handlers/light_client_message_handler.py**:

  - refactor build_light_client_protocol_message to use light_client_address instead of the old receiver and sender addresses
  - refactor LightClientMessageHandler class methods to remove receiver and sender addresses and use light_client_address instead
  - fixing a bug for update_light_client_payment_status inside LightClientMessageHandler class
  - refactor is_light_client_protocol_message_already_stored method name to get_message_for_payment, also added message_id and light_client_address as parameters
  - refactor method name is_message_already_stored to get_message_by_content
  - refactor method name get_light_client_protocol_message_by_identifier to get_message_by_identifier_for_lc
  - refactor method name get_light_client_protocol_message_by_internal_identifier to get_message_by_internal_identifier
  - adding internal helper method called map_message_from_result to map the result of the message queries to the class object
  - refactor method get_light_client_payment_locked_transfer to use the map method instead of creating the class manually
  - refactor method store_lc_processed to receive the light_client_address as a parameter to use in the new checks
  - refactor store_lc_delivered to receive the light_client_address as a parameter to use in the new checks
  - refactor call LightClientMessageHandler.is_light_client_protocol_message_already_stored_message_id to use the new method LightClientMessageHandler.get_message_for_order_and_address to check if the message exists

- **raiden/lightclient/handlers/light_client_service.py**:

  - refactor get_light_client_payment to be readable since the old version was really hard to understand

- **raiden/lightclient/models/light_client_payment.py**:

  - refactor LightClientPayment class with creator address only (specifying only for queries)

- **raiden/lightclient/models/light_client_protocol_message.py**:

  - refactor LightClientProtocolMessage class to use light_client_address instead of receiver and sender address, also adding from_dict missing method

- **raiden/message_handler.py**:

  - refactor handle_message_delivered and handle_message_processed to add the node_address (lc address)

- **raiden/network/transport/matrix/transport.py**:

  - removing the call `peer_ids = self._address_mgr.get_userids_for_address(address)` to fix RLP-832 bug
  - adding better logging for messages sent out on matrix using log instead of print

- **raiden/raiden_event_handler.py**:

  - refactor handle_store_message to use the new logic for message existence checks
  - adding new method get_existing_lc_message_from_store_event to encapsulate the existence check
  - refactor handle_send_lockexpired to use the new methods for message existence
  - refactor handle_contract_send_channelclose to use the new methods for message existence

- **raiden/raiden_service.py**:

  - refactor initiate_send_delivered_light to use the new methods for message existence
  - refactor initiate_send_processed_light to add named parameters on LightClientMessageHandler.store_light_client_protocol_message call

- **raiden/storage/sqlite.py**:

  - refactor syntax on update_light_client_payment_status to be readable
  - refactor write_light_client_protocol_message method to accept light_client_address instead of sender and receiver addresses
  - refactor method is_message_already_stored to get_message_by_content
  - refactor method is_light_client_protocol_message_already_stored to get_message_for_payment and also adding message_id and light_client_address as parameters for query
  - refactor is_light_client_protocol_message_already_stored_with_message_id to get_message_for_order_and_address adding light_client_address as a new parameter for query
  - adding new internal helper static method called get_client_values to reuse it in several occurrences
  - refactor light_clients_to_list_of_dicts to use the new method
  - refactor get_light_client to use the new method
  - refactor get_light_client_by_api_key to use the new method
  - refactor get_light_client_payment to be more readable and also to remove the sender and receiver addresses and change them to only the light_client_address
  - refactor get_light_client_protocol_message_by_identifier to get_message_by_identifier_for_lc and also adding the light_client_address as new parameter for query
  - refactor get_light_client_protocol_message_by_internal_identifier to get_message_by_internal_identifier
  - refactor get_light_client_protocol_message_by_internal_identifier to remove sender and receiver and adding light_client_address
  - refactor get_light_client_payment_locked_transfer to be more readable and to add the light_client_address as new parameter
  - refactor update_offchain_light_client_protocol_message_set_signed_message to add the light_client_address as new parameter
  - refactor is_message_already_stored to get_message_by_content

- **raiden/storage/utils.py**:

  - refactor light_client_protocol_message table to remove receiver and sender columns and adding the old light_client_address column back

- **raiden/transfer/channel.py**:

  - refactor events_for_expired_lock to add named parameters to StoreMessageEvent init call and also changing the receiver and sender addresses for the light_client_address
  - refactor handle_receive_lock_expired_light to add named parameters to StoreMessageEvent init call and also changing the receiver and sender addresses for the light_client_address

- **raiden/transfer/mediated_transfer/events.py**:

  - refactor StoreMessageEvent class to get only light_client_address and removing an unused dependency

- **raiden/transfer/mediated_transfer/initiator.py**:

  - refactor try_new_route_light to change the StoreMessageEvent init method to have named parameters and removing sender and receiver to add light_client_address
  - refactor handle_secretrequest_light to change the StoreMessageEvent init method to have named parameters and removing sender and receiver to add light_client_address
  - refactor handle_send_secret_reveal_light to change the StoreMessageEvent init method to have named parameters and removing sender and receiver to add light_client_address
  - refactor handle_offchain_secretreveal_light to change the StoreMessageEvent init method to have named parameters and removing sender and receiver to add light_client_address
  - refactor handle_send_secret_reveal_light to remove unused channel_state param

- **raiden/transfer/mediated_transfer/initiator_manager.py**:

  - refactor handle_store_refund_transfer_light to change the StoreMessageEvent init method to have named parameters and removing sender and receiver to add light_client_address

- **raiden/transfer/mediated_transfer/target.py**:

  - refactor methods handle_inittarget_light and handle_send_secret_reveal_light to change the StoreMessageEvent init method to have named parameters and removing sender and receiver to add light_client_address
  - refactor handle_offchain_secretreveal_light to change the StoreMessageEvent init method but also to add a missing call for store_received_reveal event

- **raiden/transfer/node.py**:

  - refactor methods handle_init_unlock_light, handle_init_send_lock_expired_light to change the StoreMessageEvent init method to have named parameters and removing sender and receiver to add light_client_address

- **raiden/transfer/views.py**:

  - refactor get_channelstate_for to use lambdas to fix a bug that we had when we created 2 channels in the same hub for the same LC

#### All changes can be review in the commits   [here](https://github.com/rsksmart/lumino/pull/119/commits/3c2b9e4c14a605f17327a253c2cc5bf54e683905) and [here](https://github.com/rsksmart/lumino/pull/119/commits/03a186d174a676d7f31fb739ddbafc0310091f81).

- **raiden/transfer/token_network.py**
  - Check if both_participants_are_light_clients before duplicating the channel state
  - Refactored token_network's in order to subdispatch state changes to both ends of the channel.
  - Renamed *subdispatch_to_channel_by_id_and_address* to  *subdispatch_to_channels_by_participant_address*


raiden/api/python.py
```diff
@@ -69,7 +69,7 @@
     NettingChannelState,
     TargetTask,
     TransferTask,
-    ChainState)
+    ChainState, PaymentMappingState)
```
mortelli: let's add a newline before this new import

raiden/api/python.py
```diff
@@ -193,25 +193,25 @@ def get_transfer_from_task(
 
 
 def transfer_tasks_view(
-    transfer_tasks: Dict[SecretHash, TransferTask],
+    payment_mapping: Dict[Address, PaymentMappingState],
```
mortelli: i think it's ok to keep this `Dict` type alias around since it was already present, but for a new structure, i'd always opt for native types rather than custom classes unless we need something specific from them.

just something to keep in mind

mortelli: `address_to_payment_mapping`, or `address_payment_mappings` seem (to me) better choices for this field name

raiden/api/python.py
```diff
@@ -193,25 +193,25 @@ def get_transfer_from_task(
...
 
-    for secrethash, transfer_task in transfer_tasks.items():
-        transfer, role = get_transfer_from_task(secrethash, transfer_task)
-
-        if transfer is None:
-            continue
-        if token_address is not None:
-            if transfer.token != token_address:
+    for node_address in payment_mapping.keys():
+        for secrethash, transfer_task in payment_mapping[node_address].items():
```
mortelli: `for k in d`—when `d` is a dictionary—already assigns key values to `k`, so you do not need to use `.keys()` in python 3.

moreover, you can call `.values()` instead, if you're not going to use the keys for anything (as it seems to be the case here).

raiden/api/python.py
```diff
@@ -1733,8 +1732,8 @@ def create_light_client_payment(
                 privkey=self.raiden.privkey,
             )
             if prev_secrethash:
-                current_payment_task = chain_state.payment_mapping.secrethashes_to_task[prev_secrethash]
-                chain_state.payment_mapping.secrethashes_to_task.update(
+                current_payment_task = chain_state.payment_mapping[creator_address].secrethashes_to_task[prev_secrethash]
+                chain_state.payment_mapping[creator_address].secrethashes_to_task.update(
```
mortelli: these seem long. maybe it would be useful to define getter-type functions on the chain state or the payment mapping types to make this less verbose.

raiden/message_handler.py
```diff
@@ -31,27 +31,27 @@
...
 from raiden.utils import pex, random_secret
-from raiden.utils.typing import MYPY_ANNOTATION
+from raiden.utils.typing import MYPY_ANNOTATION, Address
 
 log = structlog.get_logger(__name__)  # pylint: disable=invalid-name
 
 
 class MessageHandler:
-    def on_message(self, raiden: RaidenService, message: Message, is_light_client: bool = False) -> None:
+    def on_message(self, raiden: RaidenService, message: Message, node_address: Address, is_light_client: bool = False) -> None:
```
mortelli: `node_address` seems ambiguous. is this the sender or the recipient of the message?

raiden/message_handler.py
```diff
@@ -76,7 +76,7 @@ def on_message(self, raiden: RaidenService, message: Message, is_light_client: b
             log.error("Unknown message cmdid {}".format(message.cmdid))
 
     @staticmethod
-    def handle_message_secretrequest(raiden: RaidenService, message: SecretRequest,
+    def handle_message_secretrequest(raiden: RaidenService, message: SecretRequest, node_address: Address,
```
mortelli: same here

raiden/message_handler.py
```diff
@@ -100,23 +101,24 @@ def handle_message_secretrequest(raiden: RaidenService, message: SecretRequest,
             raiden.handle_and_track_state_change(secret_request)
 
     @staticmethod
-    def handle_message_revealsecret(raiden: RaidenService, message: RevealSecret, is_light_client=False) -> None:
+    def handle_message_revealsecret(raiden: RaidenService, message: RevealSecret, node_address: Address, is_light_client=False) -> None:
```
mortelli: and here.

also, i understand this is probably too much of a change to the codebase right now, but if we already have `message.sender`, it seems organic that the receiver is also part of the message variable.

raiden/network/resolver/client.py
```diff
@@ -18,7 +18,7 @@ def reveal_secret_with_resolver(
 
     assert isinstance(raiden.wal, WriteAheadLog), "RaidenService has not been started"
     current_state = raiden.wal.state_manager.current_state
-    task = current_state.payment_mapping.secrethashes_to_task[secret_request_event.secrethash]
+    task = current_state.payment_mapping[raiden.address].secrethashes_to_task[raiden.address, secret_request_event.secrethash]
```
mortelli: i would extract `raiden.address` here, since it is used three times in the method, and also name it something more descriptive.

additionally, down below `raiden.config` is used. could we not have a `config.address` there?

would there be any difference between these 2 addresses?

raiden/network/transport/matrix/transport.py
```diff
@@ -264,8 +264,9 @@ class MatrixTransport(Runnable):
     _room_sep = "_"
     log = log
 
-    def __init__(self, config: dict, current_server_name: str = None):
+    def __init__(self, address: Address, config: dict, current_server_name: str = None):
         super().__init__()
+        self.address=address
```
mortelli: add spaces around the `=`

raiden/network/transport/matrix/transport.py
```diff
@@ -1717,7 +1718,7 @@ def _receive_delivered_to_lc(self, delivered: Delivered):
         )
 
         assert self._raiden_service is not None
-        self._raiden_service.on_message(delivered, True)
+        self._raiden_service.on_message(delivered, self.get_address(), True)
```
mortelli, line 1730: there is no point in keeping `get_address` around now that you have the `address` field

raiden/transfer/mediated_transfer/state_change.py
```diff
@@ -479,11 +480,12 @@ def __init__(
         self.expiration = expiration
         self.secrethash = secrethash
         self.secret_request_message = secret_request_message
+        self.recipient = recipient
         self.revealsecret = None
 
     def __repr__(self) -> str:
-        return "<ReceiveSecretRequestLight paymentid:{} amount:{} secrethash:{} sender:{}>".format(
-            self.payment_identifier, self.amount, pex(self.secrethash), pex(self.sender)
+        return "<ReceiveSecretRequestLight paymentid:{} amount:{} secrethash:{} sender:{} recipient:[]>".format(
```
mortelli: you should use `{}` instead of `[]` here

raiden/ui/app.py
```diff
@@ -121,7 +121,7 @@ def _setup_matrix(config):
 
             light_client_transports.append(light_client_transport)
 
-        hub_transport = MatrixTransport(config["transport"]["matrix"])
+        hub_transport = MatrixTransport(config["address"], config["transport"]["matrix"])
```
mortelli, line 124: there are other places in the code in which the `MatrixTransport` class is instantiated, but it does not receive 2 parameters.

search for `MatrixTransport(` and fix these too

raiden/transfer/views.py
```diff
@@ -526,32 +526,24 @@ def secret_from_transfer_task(
...
     """
-    task = chain_state.payment_mapping.secrethashes_to_task.get(secrethash)
+    task = chain_state.payment_mapping[node_address].secrethashes_to_task.get(secrethash)
     if not task:
         return None
     return role_from_transfer_task(task)
 
 
-def get_transfer_secret(chain_state: ChainState, secrethash: SecretHash) -> Optional[Secret]:
+def get_transfer_secret(chain_state: ChainState, node_address: AddressHex, secrethash: SecretHash) -> Optional[Secret]:
```
mortelli: why are these `AddressHex` types instead of `Address` types?

raiden/transfer/views.py
```diff
@@ -526,32 +526,24 @@ def secret_from_transfer_task(
...
 
 
-def get_transfer_task(chain_state: ChainState, secrethash: SecretHash) -> Optional[TransferTask]:
-    return chain_state.payment_mapping.secrethashes_to_task.get(secrethash)
-
-
-def get_all_transfer_tasks(chain_state: ChainState) -> Dict[SecretHash, TransferTask]:
-    return chain_state.payment_mapping.secrethashes_to_task
-
-
```
mortelli: i haven't found usages of these 2, but they were important. please make absolutely sure we do not need them anymore.

in any case, i would build equivalent methods as sort of getter functions (for brevity's and readability's sake) as i previously mentioned.

raiden/transfer/state.py
```diff
@@ -292,7 +292,7 @@ def __init__(
         self.identifiers_to_paymentnetworks: Dict[PaymentNetworkID, PaymentNetworkState] = dict()
         self.nodeaddresses_to_networkstates: Dict[Address, str] = dict()
         self.our_address = our_address
-        self.payment_mapping = PaymentMappingState()
+        self.payment_mapping = Dict[AddressHex, PaymentMappingState] = dict()
```
mortelli: this also seems like it should be `Address` instead of `AddressHex`

raiden/transfer/node.py
```diff
@@ -836,35 +840,35 @@ def handle_receive_transfer_refund_cancel_route(
...
-    return subdispatch_to_paymenttask(chain_state, state_change, secrethash)
+    return subdispatch_to_paymenttask(chain_state, state_change, chain_state.our_address, secrethash)
 
 
 def handle_receive_secret_request_light(
     chain_state: ChainState, state_change: ReceiveSecretRequestLight
 ) -> TransitionResult[ChainState]:
     secrethash = state_change.secrethash
-    return subdispatch_to_paymenttask(chain_state, state_change, secrethash)
+    return subdispatch_to_paymenttask(chain_state, state_change, state_change.recipient, secrethash)
```
mortelli, line 901: could you please explain why the upper function needs `our_address` but the lower one needs `recipient`?

raiden/raiden_service.py
```diff
@@ -656,8 +656,8 @@ def get_block_number(self) -> BlockNumber:
         assert self.wal, f"WAL object not yet initialized. node:{self!r}"
         return views.block_number(self.wal.state_manager.current_state)
 
-    def on_message(self, message: Message, is_light_client: bool = False):
-        self.message_handler.on_message(self, message, is_light_client)
+    def on_message(self, message: Message, address: Address, is_light_client: bool = False):
```
mortelli: `address` should make reference to either the sender or the receiver through its param name

raiden/raiden_service.py
```diff
@@ -873,29 +873,30 @@ def _initialize_payment_statuses(self, chain_state: ChainState):
...
-                self.targets_to_identifiers_to_statuses[target][identifier] = PaymentStatus(
-                    payment_identifier=identifier,
-                    payment_hash_invoice=payment_hash_invoice,
-                    amount=transfer_description.amount,
-                    token_network_identifier=TokenNetworkID(
-                        balance_proof.token_network_identifier
-                    ),
-                    payment_done=AsyncResult(),
-                )
+            for node_address in chain_state.payment_mapping.keys():
```
mortelli: again you do not need `keys()` here, and you can get away with directly accessing the dict values instead

raiden/transfer/node.py
```diff
@@ -199,20 +199,19 @@ def subdispatch_to_all_lockedtransfers(
     chain_state: ChainState, state_change: StateChange, storage=None
 ) -> TransitionResult[ChainState]:
     events = list()
-
-    for secrethash in list(chain_state.payment_mapping.secrethashes_to_task.keys()):
-        result = subdispatch_to_paymenttask(chain_state, state_change, secrethash, storage)
-        events.extend(result.events)
-
+    for node_address in chain_state.payment_mapping.keys():
```
mortelli: unnecessary use of `keys()` here

raiden/transfer/node.py
```diff
@@ -199,20 +199,19 @@ def subdispatch_to_all_lockedtransfers(
...
+        for secrethash in list(chain_state.payment_mapping[node_address].secrethashes_to_task.keys()):
+            result = subdispatch_to_paymenttask(chain_state, state_change, secrethash, storage)
+            events.extend(result.events)
     return TransitionResult(chain_state, events)
 
 
 def subdispatch_to_paymenttask(
-    chain_state: ChainState, state_change: StateChange, secrethash: SecretHash, storage=None
+    chain_state: ChainState, state_change: StateChange, node_address: AddressHex, secrethash: SecretHash, storage=None
```
mortelli, line 213: are you sure this shouldn't be of the `Address` type instead?

raiden/transfer/node.py
```diff
@@ -440,12 +441,13 @@ def subdispatch_targettask(
     state_change: StateChange,
     token_network_identifier: TokenNetworkID,
     channel_identifier: ChannelID,
+    node_address: AddressHex,
```
mortelli, line 458: same `type` question here

raiden/transfer/node.py
```diff
@@ -673,19 +675,20 @@ def handle_node_change_network_state(
...
-        result = subdispatch_mediatortask(
-            chain_state=chain_state,
-            state_change=state_change,
-            secrethash=secrethash,
-            token_network_identifier=subtask.token_network_identifier,
-            storage=storage
-        )
-        events.extend(result.events)
-
+    for node_address in chain_state.payment_mapping.keys():
```
mortelli: use `values()` instead

mortelli (commented):
this looks well on its way :+1: 

some general remarks:
- getter functions on either the chain state type or the payment mappings dictionary would greatly help with code readability and reduced sentence length. 
- when node addresses are used in the context of messages, it should be evident whether we are talking about sender or receivers (i think in most cases—if not all—they mean receivers, but we can make this more explicit).
- i suspect this has to do with protocol logic, but i've observed that in some cases, the `node.py` handling functions sometimes receive "our address" as parameters, while in other cases they use "recipients". it's very important to make sure that each of these cases is correct, as the code didn't care about this before but it does now, and these can turn out to be breaking changes.
- the rest of the comments are about programming style, but you already expected that :zany_face:

raiden/api/webui/static/endpointConfig.js
```diff
@@ -1,6 +1,6 @@
-const backendUrl='http://localhost:5002';
-const nodeAddress = '0x62Aa1173561C4E92876B3D073B9767A2d07496f8';
-const rnsDomain = null;
+const backendUrl='http://localhost:5012'; 
+const nodeAddress = '0x55825d1DA0576780926568e208f7D3F55B89Cf1a'; 
+const rnsDomain = null 
 const chainEndpoint = 'http://localhost:4444'; 
 
 window.luminoUrl = backendUrl;
```
mortelli: whoops

raiden/api/python.py
marcosmartinez7: Yeah, i will do that refactor after the acceptance criteria of this pr is met.

raiden/message_handler.py
marcosmartinez7: Yes, but the receiver isnt on each Message of the protocol and we wanted to stick to the raiden format of the msgs for now

raiden/network/transport/matrix/transport.py
marcosmartinez7: sure, i will change that when the transport refactor is done

raiden/transfer/views.py
marcosmartinez7: They are basically the same. This is inherited from the raiden codebase, i found useful to use AddressHex because is a hexadecimal string, so the class name is more representative

raiden/ui/app.py
marcosmartinez7: Yes i know, thanks

raiden/transfer/views.py
mortelli: shouldn't we pick one and use the same one everywhere?

raiden/blockchain/state.py
```diff
@@ -77,7 +78,8 @@ def get_channel_state(
         open_transaction=open_transaction,
         close_transaction=close_transaction,
         settle_transaction=settle_transaction,
-        is_light_channel=is_light_channel
+        is_light_channel=is_light_channel,
+        both_participants_are_light_clients=both_participants_are_light_clients
```
marcosmartinez7, line 82: This must be changed in a future to an enum that represents: 

1: full nodes channel
2: full node with lc channel
3: lc - lc channel

now it will take a lot of time and refactor

raiden/network/transport/matrix/transport.py
```diff
@@ -264,8 +264,9 @@ class MatrixTransport(Runnable):
     _room_sep = "_"
     log = log
 
-    def __init__(self, config: dict, current_server_name: str = None):
+    def __init__(self, address: Address, config: dict, current_server_name: str = None):
```
marcosmartinez7, line 267: this will be modified after https://github.com/rsksmart/lumino/pull/110 merge

raiden/network/transport/matrix/transport.py
```diff
@@ -625,8 +626,8 @@ def _handle_invite(self, room_id: _RoomID, state: dict):
         """ Join rooms invited by whitelisted partners """
         if self._stop_event.ready():
             return
-
-        self.log.debug("Got invite", room_id=room_id)
+        print("///// RECIBI INVITE ////")
+        self.log.info("Got invite", room_id=room_id)
```
marcosmartinez7, line 627: must be removed

raiden/network/transport/matrix/transport.py
```diff
@@ -996,26 +992,29 @@ def _get_public_room(self, room_name, invitees: List[User]):
...
-                    self.log.debug(
+                    self.log.info(
                         f"Error joining room",
                         room_name=room_name,
                         error=error.content,
                         error_code=error.code,
                     )
             else:
+                print("////VOY A INVITAR /////")
+                print("invitados")
```
marcosmartinez7: must be removed

raiden/network/transport/matrix/transport.py
```diff
@@ -1551,26 +1550,29 @@ def _get_public_room(self, room_name, invitees: List[User]):
...
```
marcosmartinez7: must be removed

raiden/raiden_service.py
```diff
@@ -1327,30 +1328,43 @@ def initiate_send_delivered_light(self, sender_address: Address, receiver_addres
                                       delivered: Delivered, msg_order: int, payment_id: int,
                                       message_type: LightClientProtocolMessageType):
         lc_transport = self.get_light_client_transport(to_checksum_address(sender_address))
+        # check if receiver is a handled light client too
+        print("ME LLEGO UN DELIVERED CON ORDEN")
+        print(msg_order)
```
marcosmartinez7: must be removed

raiden/raiden_service.py
```diff
@@ -1327,30 +1328,43 @@ def initiate_send_delivered_light(self, sender_address: Address, receiver_addres
...
-                delivered,
-                True,
-                payment_id,
-                sender_address,
-                msg_order,
-                message_type,
-                self.wal
-            )
-            lc_transport.send_for_light_client_with_retry(receiver_address, delivered)
+            exists = LightClientMessageHandler.is_light_client_protocol_message_already_stored_message_id(
```
marcosmartinez7: must be added to processed handling too

raiden/raiden_service.py
```diff
@@ -1327,30 +1328,43 @@ def initiate_send_delivered_light(self, sender_address: Address, receiver_addres
...
+                    self.wal
+                )
+                lc_transport.send_for_light_client_with_retry(receiver_address, delivered)
 
     def initiate_send_processed_light(self, sender_address: Address, receiver_address: Address,
                                       processed: Processed, msg_order: int, payment_id: int,
                                       message_type: LightClientProtocolMessageType):
         lc_transport = self.get_light_client_transport(to_checksum_address(sender_address))
+        # check if receiver is a handled light client too
+        receiver_is_light_client = self.get_light_client_transport(to_checksum_address(receiver_address)) is not None
```
marcosmartinez7: Proc. msg handling must implement this?

raiden/api/python.py
```diff
@@ -1768,16 +1767,19 @@ def create_light_client_payment(
                                          amount,
                                          str(date.today()),
                                          LightClientPaymentStatus.Pending,
-                                         locked_transfer.payment_identifier)
+                                         locked_transfer.payment_identifier,
+                                         creator_address, partner_address)
```
marcosmartinez7: Not sure why this change is needed

raiden/lightclient/handlers/light_client_utils.py
```diff
@@ -58,7 +58,7 @@ def create_locked_transfer(cls, chain_state: ChainState, channel_state: NettingC
             channel_identifier=channel_state.canonical_identifier.channel_identifier,
             transferred_amount=transferred_amount,
             locked_amount=locked_amount,
-            recipient=channel_state.partner_state.address,
+            recipient=channel_state.partner_state.address, # TODO is this ok? what happenswith mediated payments?
```
marcosmartinez7: This seems to be ok, so we can delete the TODO

raiden/raiden_service.py
marcosmartinez7: We must check against the db not with the transport layer

raiden/transfer/token_network.py
```diff
@@ -105,15 +107,25 @@ def handle_channelnew(
...
 
     # Ignore duplicated channelnew events. For this to work properly on channel
     # reopens the blockchain events ChannelSettled and ChannelOpened must be
     # processed in correct order, this should be guaranteed by the filters in
     # the ethereum node
     if channel_identifier not in token_network_state.channelidentifiers_to_channels[our_address]:
+
         token_network_state.channelidentifiers_to_channels[our_address][channel_identifier] = channel_state
 
+        #if channel_state.test: ## mismo hub, 2 light clients
```
marcosmartinez7: This must be changed in order to do it only when both participants are handled lc

raiden/blockchain/state.py
marcosmartinez7: We should rename this to both_participants_are_light_clients_same_hub or something

raiden/raiden_service.py
```diff
@@ -1327,30 +1328,43 @@ def initiate_send_delivered_light(self, sender_address: Address, receiver_addres
                                       delivered: Delivered, msg_order: int, payment_id: int,
                                       message_type: LightClientProtocolMessageType):
         lc_transport = self.get_light_client_transport(to_checksum_address(sender_address))
+        # check if receiver is a handled light client too
+        print("ME LLEGO UN DELIVERED CON ORDEN")
+        print(msg_order)
+        print(sender_address.hex())
+        print(receiver_address.hex())
+        receiver_is_light_client = self.get_light_client_transport(to_checksum_address(receiver_address)) is not None
```
marcosmartinez7: We must check against the db not with the transport layer

marcosmartinez7: Or just check the channel_state.both_participants_are_lc

raiden/api/python.py
JONAF2103: We changed this to payment_states_by_address since it's a dict of PaymentMappingState by Address key

raiden/message_handler.py
JONAF2103: Done here

PR #116, opened 2020-09-09
RLP-767: remove send_for_light_client_with_retry

mortelli:
### problem
- under #110, transport variables are now of the `TransportLayer` abstract class type.
- the raiden service currently calls the `send_for_light_client_with_retry` method, which is not only _just_ implemented by `MatrixLightClientTransport`—and not `MatrixTransport`—but it is also **not** part of the `TransportLayer` abstraction.

### analysis
what is the difference between `send_for_light_client_with_retry` and `send_async`?

```python
# MatrixLightClientTransport
def send_for_light_client_with_retry(self, receiver: Address, message: Message):
    retrier = self._get_retrier(receiver)
    retrier.enqueue_global(message)

# MatrixTransport
def _get_retrier(self, receiver: Address) -> _RetryQueue:
    if receiver not in self._address_to_retrier:
        retrier = _RetryQueue(transport=self, receiver=receiver)
        self._address_to_retrier[receiver] = retrier
        retrier.start()
    return self._address_to_retrier[receiver]

# _RetryQueue
def enqueue_global(self, message: Message):
    self.enqueue(
        queue_identifier=QueueIdentifier(
            recipient=self.receiver, channel_identifier=CHANNEL_IDENTIFIER_GLOBAL_QUEUE
        ),
        message=message,
    )
```

---

```python
# MatrixTransport
def send_async(self, queue_identifier: QueueIdentifier, message: Message):
    receiver_address = queue_identifier.recipient

    if not is_binary_address(receiver_address):
        raise ValueError("Invalid address {}".format(pex(receiver_address)))

    if isinstance(message, (Delivered, Ping, Pong)):
        raise ValueError(
            "Do not use send_async for {} messages".format(message.__class__.__name__)
        )

    self.log.info(
        "Send async",
        receiver_address=pex(receiver_address),
        message=message,
        queue_identifier=queue_identifier,
    )

    self._send_with_retry(queue_identifier, message)

# MatrixTransport
def _send_with_retry(self, queue_identifier: QueueIdentifier, message: Message):
    retrier = self._get_retrier(queue_identifier.recipient)
    retrier.enqueue(queue_identifier=queue_identifier, message=message)
```

both:
- fetch a retrier for a message recipient
- use the retrier to put the given message on a queue

but:
- `send_async` receives the queue identifier directly, while `send_for_light_client_with_retry` ends-up building its own queue identifier with `CHANNEL_IDENTIFIER_GLOBAL_QUEUE` as its channel identifier.
[...]

PR #120, opened 2020-09-16
RLP-775: fixes to transport generalization, wave 1

mortelli:
This PR is the first wave of fixes for [the ongoing transport generalization PR](https://github.com/rsksmart/lumino/pull/110).

The objective here is to fix bugs introduced in said PR, but which are already evident based on the execution of the test suite. This was done based on [the `develop` test baseline](https://jirainfuy.atlassian.net/browse/RLP-760).

After this is merged, manual verifications should take place in order to then continue with PR #114 (and the rest of the transport tasks).

### Changes
- some unit tests had to be fixed, mainly due to `MatrixTransport` requiring an `address` field when initialized. these weren't business logic failures, but rather test code exceptions.
- the `MatrixTransport` implementation of the `link_exception` and `join` methods just flat out [didn't work](https://github.com/rsksmart/lumino/pull/120/files#r489537045). this **was a breaking failure.**
- the `structlog` dependency is bumped up to a more recent version due to a [nasty bug](https://github.com/rsksmart/lumino/pull/120#discussion_r489418906). this was also going to crash the node.
- some minor formatting changes were applied.

more in-depth details can be found in the diff section, as comments.

---

Wave 1 for issue [RLP-755](https://jirainfuy.atlassian.net/browse/RLP-775).

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -64,7 +64,7 @@ semantic-version==2.6.0
 semver==2.8.1
 simplegeneric==0.8.1
 six==1.11.0
-structlog==18.2.0
+structlog==20.1.0
 toolz==0.9.0
 traitlets==4.3.2
 urllib3==1.23
--- a/raiden/network/transport/matrix/transport.py
+++ b/raiden/network/transport/matrix/transport.py
@@ -254,13 +254,13 @@ class _RetryQueue(Runnable):
 class MatrixTransport(TransportLayer, Runnable):
     _room_prefix = "raiden"
     _room_sep = "_"
+    log = log
 
     def __init__(self, address: Address, config: dict, current_server_name: str = None):
         TransportLayer.__init__(self, address)
         Runnable.__init__(self)
         self._config = config
         self._raiden_service: Optional[RaidenService] = None
-        self.log = log
 
         available_servers = get_available_servers_from_config(self._config)
 
@@ -1735,6 +1735,7 @@ class MatrixLightClientTransport(MatrixTransport):
         except (InvalidAddress, UnknownAddress, UnknownTokenAddress):
             self.log.warning("Exception while processing message", exc_info=True)
             return
+
 
 class NodeTransport:
 
--- a/raiden/network/transport/matrix/transport.py
+++ b/raiden/network/transport/matrix/transport.py
@@ -254,13 +254,13 @@ class _RetryQueue(Runnable):
 class MatrixTransport(TransportLayer, Runnable):
     _room_prefix = "raiden"
     _room_sep = "_"
-    log = log
 
     def __init__(self, address: Address, config: dict, current_server_name: str = None):
         TransportLayer.__init__(self, address)
         Runnable.__init__(self)
         self._config = config
         self._raiden_service: Optional[RaidenService] = None
+        self.log = log
 
         available_servers = get_available_servers_from_config(self._config)
 
@@ -1337,10 +1337,10 @@ class MatrixTransport(TransportLayer, Runnable):
         return True
 
     def link_exception(self, callback: Any):
-        Runnable.link_exception(self, callback)
+        self.greenlet.link_exception(callback)
 
     def join(self, timeout=None):
-        Runnable.join(self, timeout)
+        self.greenlet.join(timeout)
 
 
 class MatrixLightClientTransport(MatrixTransport):
--- a/raiden/tests/integration/fixtures/transport.py
+++ b/raiden/tests/integration/fixtures/transport.py
@@ -63,6 +63,7 @@ def matrix_transports(
         server = local_matrix_servers[transport_index % len(local_matrix_servers)]
         transports.append(
             MatrixTransport(
+                bytearray(),  # must be set before starting the transport
                 {
                     "global_rooms": global_rooms,
                     "retries_before_backoff": retries_before_backoff,
--- a/raiden/tests/integration/network/transport/test_matrix_transport.py
+++ b/raiden/tests/integration/network/transport/test_matrix_transport.py
@@ -446,7 +446,12 @@ def test_matrix_message_retry(
     message = Processed(message_identifier=0)
     transport._raiden_service.sign(message)
     chain_state.queueids_to_queues[queueid] = [message]
-    retry_queue.enqueue_global(message)
+    retry_queue.enqueue(
+        queue_identifier=QueueIdentifier(
+            recipient=retry_queue.receiver, channel_identifier=CHANNEL_IDENTIFIER_GLOBAL_QUEUE
+        ),
+        message=message
+    )
 
     gevent.sleep(1)
 
--- a/raiden/tests/integration/network/transport/test_matrix_transport.py
+++ b/raiden/tests/integration/network/transport/test_matrix_transport.py
@@ -406,8 +406,10 @@ def test_matrix_message_retry(
     the receiver comes back again, the message should be sent again.
     """
     partner_address = factories.make_address()
+    raiden_service = MockRaidenService(None)
 
     transport = MatrixTransport(
+        address=raiden_service.address,
         config={
             "global_rooms": global_rooms,
             "retries_before_backoff": retries_before_backoff,
@@ -419,7 +421,6 @@ def test_matrix_message_retry(
         },
     )
     transport._send_raw = MagicMock()
-    raiden_service = MockRaidenService(None)
 
     transport.start(raiden_service, raiden_service.message_handler, None)
     transport.log = MagicMock()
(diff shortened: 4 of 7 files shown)


raiden/network/transport/matrix/transport.py
```diff
@@ -254,13 +254,13 @@ def __repr__(self):
...
     _room_prefix = "raiden"
     _room_sep = "_"
-    log = log
 
     def __init__(self, address: Address, config: dict, current_server_name: str = None):
         TransportLayer.__init__(self, address)
         Runnable.__init__(self)
         self._config = config
         self._raiden_service: Optional[RaidenService] = None
+        self.log = log
```
mortelli: not entirely sure what's going on here.

for starters, there is no apparent reason why `log` should be a _class_ attribute rather than an instance attribute. even moreso when it's just pointing to the variable on line `79`.

for some reason, if both:
1. `log` remains a class attribute
2. `MatrixTransport` implements multiple inheritance (`TransportLayer` in addition to the previously present `Runnable`)

then `MatrixTransport` **turns into an abstract class.** 

this seems baffling, but i've made sure it is the case by calling `MatrixTransport.__abstractmethods__` and verifying that `log` appears to be an abstract method. since the class has an abstract method, it cannot be instantiated. 

as a result, transport objects won't come to be, and the whole node will come crashing down in flames.

what's even more mystifying: if `log` is commented out—or, in this case, turned into an instance variable—then the problem apparently disappears. 

inspecting the `log` variable in runtime points to some very obscure types which i'm not familiar with, and don't really understand how they can lead into an abstract method being part of this class once we have multiple inheritance.

this issue came to light after running `pytest integration/test_balance_proof_check.py`. it still needs more looking into.

raiden/network/transport/matrix/transport.py
```diff
@@ -1337,10 +1337,10 @@ def _handle_to_device_message(self, event):
         return True
 
     def link_exception(self, callback: Any):
-        Runnable.link_exception(self, callback)
+        self.greenlet.link_exception(callback)
 
     def join(self, timeout=None):
-        Runnable.join(self, timeout)
+        self.greenlet.join(timeout)
```
mortelli, line 1343: these were changed because the `link_exception` and `join` methods in the `Runnable` class work through the `__getattr__` hack that that class implements:

https://github.com/rsksmart/lumino/blob/af37b84c33e5e2a5c661551c19864cb96ec06ef8/raiden/utils/runnable.py#L75-L78

since these 2 methods are not formally defined on the `Runnable` class, the previous version of the code failed. 

this is, in principle, fixed by accessing the embedded greenlet directly.

raiden/network/transport/matrix/transport.py
mortelli: this is probably related to a new failure in `integration/api/test_restapi.py`:
```shell
___________________________________________________ test_api_payments_target_error[matrix-False-2] ___________________________________________________

api_server_test_instance = <raiden.api.rest.APIServer object at 0x7efd5caf6fd0>, raiden_network = [<App 007eb599>, <App f0feeae8>]
token_addresses = [b'\x19\xac\xdf\xba\x08\x1b\xc7\x15\xe39@\x0b\x9b\xdf(\xb0\xe6\xd5\xf4\x06']

    @pytest.mark.parametrize("number_of_nodes", [2])
    def test_api_payments_target_error(api_server_test_instance, raiden_network, token_addresses):
        _, app1 = raiden_network
        amount = 200
        identifier = 42
        token_address = token_addresses[0]
        target_address = app1.raiden.address
    
        # stop app1 to force an error
        app1.stop()
    
        request = grequests.post(
            api_url_for(
                api_server_test_instance,
                "token_target_paymentresource",
                token_address=to_checksum_address(token_address),
                target_address=to_checksum_address(target_address),
            ),
            json={"amount": amount, "identifier": identifier},
        )
        response = request.send().response
        assert_proper_response(response, status_code=HTTPStatus.CONFLICT)
>       app1.start()

integration/api/test_restapi.py:812: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../app.py:150: in start
    self.raiden.start()
../raiden_service.py:477: in start
    self._initialize_whitelists(chain_state)
../raiden_service.py:1008: in _initialize_whitelists
    self._set_hub_transport_whitelist(chain_state)
../raiden_service.py:992: in _set_hub_transport_whitelist
    self.transport.hub_transport.whitelist(neighbour)
../network/transport/matrix/transport.py:461: in whitelist
    self.log.debug("Whitelist", address=to_normalized_address(address))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <MatrixTransport node:f0feeae8 id:139626653468544>, item = 'log'

    def __getattr__(self, item: str) -> Any:
>       return getattr(self.greenlet, item)
E       AttributeError: 'gevent._greenlet.Greenlet' object has no attribute 'log'

../utils/runnable.py:78: AttributeError

```

raiden/tests/integration/fixtures/transport.py
```diff
@@ -63,6 +63,7 @@ def matrix_transports(
         server = local_matrix_servers[transport_index % len(local_matrix_servers)]
         transports.append(
             MatrixTransport(
+                bytearray(),  # must be set before starting the transport
```
mortelli, line 66: we could have raiden service variables weaved into this fixture, and then extract the addresses from them so that they can be set in the matrix transport constructors.

this is not done is because:
- for some tests, the cardinality between transport layers and raiden services is not 1:1.
- for some tests, transport layers are re-started on different raiden services, and so their addresses must be re-set.

this makes it—in my opinion—not worth the hassle to parametrize.

as a result, `_address` fields are set to their respective raiden service addresses right before starting the transport layers.

raiden/network/transport/matrix/transport.py
mortelli: i found the [root cause](https://github.com/hynek/structlog/issues/229) of the problem.

this is the fix that i've pushed: 49c357338990e226900e9fb22631c5dc42077277

constraints.txt
```diff
@@ -64,7 +64,7 @@ semantic-version==2.6.0
 semver==2.8.1
 simplegeneric==0.8.1
 six==1.11.0
-structlog==18.2.0
+structlog==20.1.0
```
mortelli, line 67: see [this thread](https://github.com/rsksmart/lumino/pull/120#discussion_r489418906)

raiden/tests/integration/network/transport/test_matrix_transport.py
```diff
@@ -441,7 +446,12 @@ def test_matrix_message_retry(
     message = Processed(message_identifier=0)
     transport._raiden_service.sign(message)
     chain_state.queueids_to_queues[queueid] = [message]
-    retry_queue.enqueue_global(message)
+    retry_queue.enqueue(
```
mortelli, line 449: we no longer have `enqueue_global` (see #116)

mortelli:
I've successfully verified that executing the test suite in this branch throws a result which matches [the current test baseline](https://jirainfuy.atlassian.net/browse/RLP-760).

raiden/network/transport/matrix/transport.py
marcosmartinez7: Not sure why using static calls was the first apporach but, if the class has a `greenlet ` member, it looks natural that it delegates the join call to that member.

raiden/tests/integration/fixtures/transport.py
marcosmartinez7: Came to the same tought about passing raiden service addresses to the fixture. Agree with this.

marcosmartinez7 (approved):
lgtm

raiden/tests/integration/fixtures/transport.py
mortelli: it's certainly not ideal, and i would love to iterate on these tests so that they can be re-structured. 

but i believe this is out of scope for this PR.

raiden/network/transport/matrix/transport.py
mortelli: these weren't static calls; the code `ParentClass.func(self, ...)` is the equivalent of doing `super().func(...)`, but directly specifying which parent class you want to use.

`super` will actually go through all the parent classes, which is both unnecessary and potentially problematic (remember this class actually has 2 parents outside of `Object`).

PR #121, opened 2020-09-18
RLP-771: remove matrix external usage

mortelli:
This PR aims to remove external usage of Matrix code, so that it remains internal to the Matrix Transport Layer implementation only.

### Scope
There still remains some external usage of Matrix code not covered by this PR, which should be tackled through the following issues:
- [RLP-774](https://jirainfuy.atlassian.net/browse/RLP-774), which is about abstracting the on-boarding of light clients so that they can be registered through RIF Comms too.
- [RLP-797](https://jirainfuy.atlassian.net/browse/RLP-797), which is about having structured code for the configuration of the Lumino node in terms of transport.

### Overview
The main issue to fix here is the use of the `_setup_matrix` stand-alone function, which returns a transport layer object. This class contains the transport objects for both the node as a hub as well as for each registered light client.

Since having a RIF Comms setup function (or something along those lines) can be expected at some point, it makes sense to have an abstract constructor for the transport layer class, and have this contain whatever setup logic is needed for each transport layer implementation.

As a result, the transport `Layer` class was made abstract, and a `MatrixTransportLayer` class was added. The constructor for the latter contains what used to be the `_setup_matrix` code. 

Any transport layer implementation must be able to:
- instantiate itself based on a `config` dictionary
- provide access to its `hub_transport` field
- provide access to its `light_client_transports` field
- add a light client transport
- remove a light client transport
- create a new light client transport instance—this is used in the Rest API LC registration process, which cannot depend on a concrete transport implementation, but rather use the one indicated in the configuration. While this logic belongs in the transport layer, the method is static as it does not need anything from a layer instance or even its class.

Regarding this last point: a **temporary** transport configuration class was added, so that light client transports could be instantiated agnostically from the Rest API. Right now, this class is hard-coded to use `MatrixTransportLayer`.

While this class is not final, it could be the basis for the work to be done on the upcoming issue [RLP-797](https://jirainfuy.atlassian.net/browse/RLP-797).

### Additional changes
- the `get_matrix_light_client_instance` method was removed since it did nothing but instantiate `MatrixLightClientTransport` classes.
- the `address` property method was removed from the `transport.Node` class, since it added nothing to the standard use of the `address` attribute.
- `MatrixTransport` and `MatrixLightClientTransport` classes were renamed to `MatrixNode` and `MatrixLightClientNode` respectively, to be consistent with the `TransportLayer` nomenclature, plus the `TransportNode` parent class.
- the `MatrixNode` and `MatrixLightClientNode` constructors were simplified so that they receive an address, plus an arbitrary configuration dictionary (instead of *n* additional independent params) and—only in the case of light clients—an authorization params dictionary too.
- docstrings were added.
- formatting changes were applied (whitespace, indentation).

--- 

Closes [RLP-771](https://jirainfuy.atlassian.net/browse/RLP-771).

### DIFF
--- a/raiden/api/rest.py
+++ b/raiden/api/rest.py
@@ -2063,7 +2063,7 @@ class RestAPI:
             config["light_client_password"] = light_client["encrypt_signed_password"]
             config["light_client_display_name"] = light_client["encrypt_signed_display_name"]
             config["light_client_seed_retry"] = light_client["encrypt_signed_seed_retry"]
-            light_client_transport = transport_config.light_client_transport(
+            light_client_transport = transport_config.light_client_transport_class(
                 address=light_client["address"],
                 config=config
             )
--- a/raiden/api/rest.py
+++ b/raiden/api/rest.py
@@ -16,7 +16,6 @@ from flask_restful import Api, abort
 from gevent.pywsgi import WSGIServer
 from hexbytes import HexBytes
 from raiden.lightclient.handlers.light_client_message_handler import LightClientMessageHandler
-from raiden.network.transport.matrix import MatrixLightClientTransport
 from raiden_webui import RAIDEN_WEBUI_PATH
 
 from raiden.api.validations.api_error_builder import ApiErrorBuilder
--- a/raiden/api/rest.py
+++ b/raiden/api/rest.py
@@ -16,6 +16,7 @@ from flask_restful import Api, abort
 from gevent.pywsgi import WSGIServer
 from hexbytes import HexBytes
 from raiden.lightclient.handlers.light_client_message_handler import LightClientMessageHandler
+from raiden.network.transport.matrix import MatrixLightClientTransport
 from raiden_webui import RAIDEN_WEBUI_PATH
 
 from raiden.api.validations.api_error_builder import ApiErrorBuilder
@@ -142,7 +143,6 @@ from raiden.billing.invoices.util.time_util import is_invoice_expired, UTC_FORMA
 from raiden.billing.invoices.constants.errors import AUTO_PAY_INVOICE, INVOICE_EXPIRED, INVOICE_PAID
 
 from raiden.utils.signer import recover
-from raiden.ui.app import get_matrix_light_client_instance
 
 log = structlog.get_logger(__name__)
 
@@ -1552,7 +1552,7 @@ class RestAPI:
         self.raiden_api.initiate_send_secret_request_light(sender_address, receiver_address, secret_request)
 
     def initiate_send_lock_expired_light(self, sender_address: typing.Address, receiver_address: typing.Address,
-                                         lock_expired: LockExpired, payment_id:int):
+                                         lock_expired: LockExpired, payment_id: int):
         self.raiden_api.initiate_send_lock_expired_light(sender_address, receiver_address, lock_expired, payment_id)
 
     def initiate_payment_light(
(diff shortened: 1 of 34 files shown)


mortelli:
putting this on hold until we make sure #110 is working ok with what we have so far

raiden/api/rest.py
```diff
@@ -2014,26 +2014,33 @@ def get_data_for_registration_request(self, address):
...
+        )
 
-        # Recover lightclient addres from seed retry and signed_seed_retry
-        address_recovered_from_signed_seed_retry = recover(data=seed_retry.encode(),
-                                                           signature=decode_hex(signed_seed_retry))
+        # Recover light client address from seed retry and signed_seed_retry
+        address_recovered_from_signed_seed_retry = recover(
+            data=seed_retry.encode(),
+            signature=decode_hex(signed_seed_retry)
+        )
```
mortelli: only indentation changes here

raiden/api/rest.py
```diff
@@ -2048,22 +2055,25 @@ def register_light_client(self,
...
+            config["light_client_display_name"] = light_client["encrypt_signed_display_name"]
+            config["light_client_seed_retry"] = light_client["encrypt_signed_seed_retry"]
+            light_client_transport = transport_config.transport_layer.new_light_client_transport(
                 address=light_client["address"],
-                config=self.raiden_api.raiden.config["transport"]["matrix"],
-                password=light_client["encrypt_signed_password"],
-                display_name=light_client["encrypt_signed_display_name"],
-                seed_retry=light_client["encrypt_signed_seed_retry"])
+                config=config
+            )
```
mortelli: this now puts all the needed params for new LC instances inside of the `config` dict, and uses the transport layer static method to create these new transport instances

raiden/network/transport/matrix/layer.py
```diff
@@ -0,0 +1,108 @@
...
+from raiden.utils import Address
+from raiden.utils.cli import get_matrix_servers
+from transport.layer import Layer as TransportLayer
+from transport.node import Node as TransportNode
+
+
+class MatrixLayer(TransportLayer):
+
+    def __init__(self, config: Dict[str, Any]):
+        from raiden.ui.app import log
```
mortelli, line 23: this was put here to avoid an import cycle

raiden/network/transport/matrix/layer.py
```diff
@@ -0,0 +1,108 @@
...
+from raiden.storage import sqlite, serialize
+from raiden.utils import Address
+from raiden.utils.cli import get_matrix_servers
+from transport.layer import Layer as TransportLayer
+from transport.node import Node as TransportNode
+
+
+class MatrixLayer(TransportLayer):
+
+    def __init__(self, config: Dict[str, Any]):
```
mortelli, line 22: this should do everything `_setup_matrix` did, but with a constructor structure now

raiden/ui/app.py
```diff
@@ -12,13 +12,10 @@
 import json
 from eth_utils import encode_hex
 
-from raiden.network.transport.matrix.utils import get_available_servers_from_config, server_is_available
-from raiden.storage import serialize, sqlite
+from raiden.network.transport.matrix.layer import MatrixLayer
```
mortelli: lots of unused imports removed in this file due to taking the `_setup_matrix` method out of here

raiden/api/rest.py
marcosmartinez7: i dont understand this, on the config for matrix transport, you put in the "light_client_password" key, the password of a specific light client

every time we register a new one, the config.transport.matrix.light_client_passwordm config.transport.matrix.display_name etc will be changed

why we need this?

if it is only for the new_light_client_transport i suggest to pass another object, not to modify the config

mortelli:
[test baseline](https://jirainfuy.atlassian.net/browse/RLP-760) successfully verified :heavy_check_mark:

raiden/api/rest.py
mortelli: thanks, good point.

i've changed this so that the light client object receives an `auth_params` dictionary as an additional constructor parameter, and moved these variables out of the `config` param.

marcosmartinez7 (approved):
Lgtm!

mortelli:
> putting this on hold until we make sure #110 is working ok with what we have so far

this has now [been verified](https://github.com/rsksmart/lumino/pull/110#issuecomment-702364526).

as a result this PR is no longer on hold; as soon as i re-run the test suite plus the basic manual testing scenarios, it will be ready to be merged.

mortelli:
[as done in](https://github.com/rsksmart/lumino/pull/110#issuecomment-702364526) PR #110, i've manually verified basic operations in the following scenarios:
- `FN` ←→ `FN`
- `FN` ←→ `FN` ←→ `FN`
- `LC` ←→ `FN`
- `LC` ←→ `FN` ←→ `FN`

i've also made sure the test suite results match [the latest ones for the destination branch](https://github.com/anarancio/lumino-docs/blob/master/unit-testing/results/transport-generalization/01-10-2020/summary.txt).

with these 2 things now done, this PR is ready to be merged.

PR #122, opened 2020-09-21
RLP-796: rename transport entities

mortelli:
This PR renames and moves transport-related classes for increased clarity and consistency.

### Current naming
- class which contains transports objects for the running hub and connected light clients → `NodeTransport`
- class which contains transport logic for each address in `NodeTransport` → `transport.Layer` (imported as `TransportLayer`)

### Proposed naming
- `transport.Layer` → `transport.Node` (imported as `TransportNode`)
- `NodeTransport` → `transport.Layer` (imported as `TransportLayer`)

### Rationale
It doesn't make too much sense to have _1 transport layer per managed address_; and this leaves us (semantically) with _a node transport object which contains n transport layers_.  

In my opinion, it is preferable to call the containing object for all address "the transport layer" and then calling each of these objects "transport nodes", since they are actually representing a node (in the sense of an address).

Additionally, this naming is more congruent with the idea that a single transport layer should be chosen by configuration, and n nodes managed by it in terms of transport.

### Other considerations
The `NodeTransport`/`transport.Layer` class should **not** be under the `transport/matrix` namespace. It is therefore moved.

--- 

Closes [RLP-796](https://jirainfuy.atlassian.net/browse/RLP-796).


JONAF2103 (approved):
LGTM

marcosmartinez7 (approved):
lgtm, we should run the tests and fix them if necessary in order to merge this pr

PR #123, opened 2020-09-30
Fixing support for Lumino node to run without prompt

JONAF2103:
In this PR we are:

* Fixing getting the private key and public key from password file option

This allows to use the --address and --password-file options to avoid the prompt when lumino starts.


marcosmartinez7 (approved):

mauricioirace (approved):
Lgtm

PR #125, opened 2020-10-05
Bug/rlp 762/receive light payments not working after refund

JONAF2103:
This PR attempts to fix the bug described on https://jirainfuy.atlassian.net/browse/RLP-762

## Scenario

imagen

* Satoshi sent out a mediated payment to Andreas through the available channels of 0.3 tokens
* The first channel used is Satoshi -> Sergio that fails since Sergio doesn't have enough balance to transfer to Andreas
* The second channel used is Satoshi -> Vitalik that fails for the same reason
* The last channel used is Satoshi -> Isabel that succeed since it has enough balance
* Then when the mediated payment was successfully transferred we have 2 expirations (LockExpired) to wait for the failed payments through Sergio and Vitalik
* We try to send a payment from Sergio (or Vitalik) to Satoshi and the payment fails, we can't send any payments from those participants to Satoshi anymore from this point

## Error found

We discover through testing and debugging that the error is only when the node is in Hub Mode, this means that Satoshi must be a LC. Only in that case the error is happening, for full nodes this works just fine.
We see that the root cause of the error is a missing part of the code to handle a RefundTransfer when we are LightClients. So basically if you receive a RefundTransfer and the node is in Hub Mode the handle of that transfer is not working as it should and it doesn't updates the partner state correctly. Later because of that the partner state is unsync and we have a nonce error when we receive any payment on that channel.

## Solution

The solution was to add the missing handling logic for RefundTransfer when we are in Hub Mode and add other controls to keep the partner state up to date. Doing that the nonce never get's unsync and the payments work as expected.

## Notes

* We solve this escenario but we keep having the issue described on this https://jirainfuy.atlassian.net/browse/RLP-761, probably that scenario has another bug that's not related to this fix.

Here a demo:

demo

This affects the Payment flow for mediated payments

### DIFF
--- a/raiden/lightclient/handlers/light_client_message_handler.py
+++ b/raiden/lightclient/handlers/light_client_message_handler.py
@@ -77,34 +77,38 @@ class LightClientMessageHandler:
             storage.update_light_client_payment_status(payment_id, status)
 
     @classmethod
-    def is_light_client_protocol_message_already_stored(cls, payment_id: int,
-                                                        order: int,
-                                                        message_type: LightClientProtocolMessageType,
-                                                        message_protocol_type: str,
-                                                        wal: WriteAheadLog
-                                                        ):
-        existing_message = wal.storage.is_light_client_protocol_message_already_stored(payment_id, order,
-                                                                                       str(message_type.value),
-                                                                                       message_protocol_type)
+    def is_message_already_stored_checking_payment(cls,
+                                                   payment_id: int,
+                                                   order: int,
+                                                   message_type: LightClientProtocolMessageType,
+                                                   message_protocol_type: str,
+                                                   light_client_address: AddressHex,
+                                                   wal: WriteAheadLog):
+        existing_message = wal.storage\
+            .is_light_client_protocol_message_already_stored(payment_id=payment_id,
+                                                             order=order,
+                                                             message_type=str(message_type.value),
+                                                             message_protocol_type=message_protocol_type,
+                                                             light_client_address=str(light_client_address))
 
         if existing_message:
-            return LightClientProtocolMessage(existing_message[5] is not None,
-                                              existing_message[3],
-                                              existing_message[2],
-                                              existing_message[1],
-                                              existing_message[6],
-                                              existing_message[4],
-                                              existing_message[5],
-                                              existing_message[0],
-                                              existing_message[7])
+            return LightClientProtocolMessage(is_signed=existing_message[5] is not None,
+                                              message_order=existing_message[3],
+                                              light_client_payment_id=existing_message[2],
+                                              identifier=existing_message[1],
+                                              message_type=existing_message[6],
+                                              unsigned_message=existing_message[4],
+                                              signed_message=existing_message[5],
+                                              internal_msg_identifier=existing_message[0],
+                                              light_client_address=existing_message[7])
         return existing_message
 
     @classmethod
-    def is_message_already_stored(cls,
-                                  light_client_address: AddressHex,
-                                  message_type: LightClientProtocolMessageType,
-                                  unsigned_message: Message,
-                                  wal: WriteAheadLog):
+    def is_message_already_stored_checking_data(cls,
+                                                light_client_address: AddressHex,
+                                                message_type: LightClientProtocolMessageType,
+                                                unsigned_message: Message,
+                                                wal: WriteAheadLog):
 
         return message_type and light_client_address and unsigned_message and wal.storage\
             .is_message_already_stored(light_client_address,
@@ -120,15 +124,16 @@ class LightClientMessageHandler:
     @classmethod
     def get_light_client_protocol_message_by_identifier(cls, message_identifier: int, wal: WriteAheadLog):
         message = wal.storage.get_light_client_protocol_message_by_identifier(message_identifier)
-        return LightClientProtocolMessage(message[3] is not None,
-                                          message[1],
-                                          message[4],
-                                          message[0],
-                                          message[5],
-                                          message[2],
-                                          message[3],
-                                          None,
-                                          message[6])
+
+        return LightClientProtocolMessage(is_signed=message[3] is not None,
+                                          message_order=message[1],
+                                          light_client_payment_id=message[4],
+                                          identifier=message[0],
+                                          message_type=message[5],
+                                          unsigned_message=message[2],
+                                          signed_message=message[3],
+                                          internal_msg_identifier=message[7],
+                                          light_client_address=message[6])
 
     @classmethod
     def get_light_client_payment_locked_transfer(cls, payment_identifier: int, wal: WriteAheadLog):
@@ -182,6 +187,7 @@ class LightClientMessageHandler:
         # get first principal message by message identifier
         protocol_message = LightClientMessageHandler.get_light_client_protocol_message_by_identifier(
             message_identifier, wal)
+
         json_message = None
         if protocol_message.signed_message is None:
             json_message = protocol_message.unsigned_message
--- a/raiden/lightclient/handlers/light_client_message_handler.py
+++ b/raiden/lightclient/handlers/light_client_message_handler.py
@@ -73,8 +73,8 @@ class LightClientMessageHandler:
     def update_light_client_payment_status(cls, payment_id: int, status: LightClientPaymentStatus,
                                            storage: SerializedSQLiteStorage):
         exists_payment = storage.get_light_client_payment(payment_id)
-        if not exists_payment:
-            storage.update_light_client_payment_status(payment_id, status, storage)
+        if exists_payment:
+            storage.update_light_client_payment_status(payment_id, status)
 
     @classmethod
     def is_light_client_protocol_message_already_stored(cls, payment_id: int,
--- a/raiden/lightclient/handlers/light_client_message_handler.py
+++ b/raiden/lightclient/handlers/light_client_message_handler.py
@@ -1,5 +1,6 @@
 import json
 import string
+from typing import Any
 
 import structlog
 from eth_utils import to_checksum_address
@@ -54,20 +55,6 @@ class LightClientMessageHandler:
         )
 
     @classmethod
-    def update_offchain_light_client_protocol_message_set_signed_message(
-        cls, message: Message,
-        payment_id: int,
-        order: int,
-        message_type: LightClientProtocolMessageType,
-        wal: WriteAheadLog
-    ):
-        return wal.storage\
-            .update_offchain_light_client_protocol_message_set_signed_message(payment_id,
-                                                                              order,
-                                                                              message,
-                                                                              str(message_type.value))
-
-    @classmethod
     def update_onchain_light_client_protocol_message_set_signed_transaction(
         cls,
         internal_msg_identifier: int,
@@ -75,6 +62,21 @@ class LightClientMessageHandler:
         wal: WriteAheadLog
     ):
         return wal.storage.update_onchain_light_client_protocol_message_set_signed_transaction(
+            internal_msg_identifier=internal_msg_identifier,
+            signed_message=signed_message
+        )
+
+    @classmethod
+    def set_signed_message(cls,
+                           wal: WriteAheadLog,
+                           address: AddressHex,
+                           signed_message: Any,
+                           internal_msg_identifier: int = None,
+                           payment_id: int = None,
+                           order: int = None,
+                           message_type: LightClientProtocolMessageType = None):
+        return wal.storage.set_signed_message(
+            address=address,
             internal_msg_identifier=internal_msg_identifier,
             signed_message=signed_message
         )
@@ -92,78 +94,36 @@ class LightClientMessageHandler:
             storage.update_light_client_payment_status(payment_id, status)
 
     @classmethod
-    def is_message_already_stored_checking_payment(cls,
-                                                   payment_id: int,
-                                                   order: int,
-                                                   message_type: LightClientProtocolMessageType,
-                                                   message_protocol_type: str,
-                                                   light_client_address: AddressHex,
-                                                   wal: WriteAheadLog):
-        existing_message = wal.storage\
-            .is_light_client_protocol_message_already_stored(payment_id=payment_id,
-                                                             order=order,
-                                                             message_type=str(message_type.value),
-                                                             message_protocol_type=message_protocol_type,
-                                                             light_client_address=str(light_client_address))
+    def get_message(cls,
+                    wal: WriteAheadLog,
+                    light_client_address: AddressHex = None,
+                    is_signed: bool = None,
+                    message: Message = None,
+                    message_type: LightClientProtocolMessageType = None,
+                    message_protocol_type: str = None,
+                    identifier: int = None,
+                    internal_identifier: int = None,
+                    payment_id: int = None,
+                    order: int = None):
+        message = wal.storage.query_message(identifier=identifier,
+                                            internal_identifier=internal_identifier,
+                                            payment_id=payment_id,
+                                            order=order,
+                                            message_type=message_type,
+                                            message_protocol_type=message_protocol_type,
+                                            light_client_address=light_client_address)
 
-        if existing_message:
-            return LightClientProtocolMessage(is_signed=existing_message[5] is not None,
-                                              message_order=existing_message[3],
-                                              light_client_payment_id=existing_message[2],
-                                              identifier=existing_message[1],
-                                              message_type=existing_message[6],
-                                              unsigned_message=existing_message[4],
-                                              signed_message=existing_message[5],
-                                              internal_msg_identifier=existing_message[0],
-                                              light_client_address=existing_message[7])
-        return existing_message
-
-    @classmethod
-    def is_message_already_stored_checking_data(cls,
-                                                light_client_address: AddressHex,
-                                                message_type: LightClientProtocolMessageType,
-                                                unsigned_message: Message,
-                                                wal: WriteAheadLog):
-
-        return message_type and light_client_address and unsigned_message and wal.storage\
-            .is_message_already_stored(light_client_address,
-                                       message_type.value,
-                                       unsigned_message)
-
-    @classmethod
-    def is_light_client_protocol_message_already_stored_message_id(cls, message_id: int, payment_id: int, order: int,
-                                                                   wal: WriteAheadLog):
-        return wal.storage.is_light_client_protocol_message_already_stored_with_message_id(message_id, payment_id,
-                                                                                           order)
-
-    @classmethod
-    def get_light_client_protocol_message_by_identifier(cls, message_identifier: int, wal: WriteAheadLog):
-        message = wal.storage.get_light_client_protocol_message_by_identifier(message_identifier)
-
-        return LightClientProtocolMessage(is_signed=message[3] is not None,
-                                          message_order=message[1],
-                                          light_client_payment_id=message[4],
-                                          identifier=message[0],
-                                          message_type=message[5],
-                                          unsigned_message=message[2],
-                                          signed_message=message[3],
-                                          internal_msg_identifier=message[7],
-                                          light_client_address=message[6])
-
-    @classmethod
-    def get_light_client_protocol_message_by_internal_identifier(cls, internal_msg_identifier: int, wal: WriteAheadLog):
-        message = wal.storage.get_light_client_protocol_message_by_internal_identifier(internal_msg_identifier)
         if message:
-            return LightClientProtocolMessage(message[3] is not None,
-                                              message[1],
-                                              message[4],
-                                              message[0],
-                                              message[5],
-                                              message[2],
-                                              message[3],
-                                              message[7],
-                                              message[6])
-        return None
+            return LightClientProtocolMessage(is_signed=message[5] is not None,
+                                              message_order=message[3],
+                                              light_client_payment_id=message[2],
+                                              identifier=message[1],
+                                              message_type=message[6],
+                                              unsigned_message=message[4],
+                                              signed_message=message[5],
+                                              internal_msg_identifier=message[0],
+                                              light_client_address=message[7])
+        return message
 
     @classmethod
     def get_light_client_payment_locked_transfer(cls, payment_identifier: int, wal: WriteAheadLog):
@@ -215,8 +175,7 @@ class LightClientMessageHandler:
         # If exists for that payment, the same message by the order, then discard it.
         message_identifier = message.message_identifier
         # get first principal message by message identifier
-        protocol_message = LightClientMessageHandler.get_light_client_protocol_message_by_identifier(
-            message_identifier, wal)
+        protocol_message = LightClientMessageHandler.get_message(identifier=message_identifier, wal=wal)
 
         json_message = None
         if protocol_message.signed_message is None:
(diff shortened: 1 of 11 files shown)


raiden/message_handler.py
```diff
@@ -163,10 +167,16 @@ def handle_message_refundtransfer(raiden: RaidenService, message: RefundTransfer
                 sender=from_transfer.balance_proof.sender,  # pylint: disable=no-member
             )
             raiden.handle_and_track_state_change(state_change)
+
             if is_light_client:
-                store_refund = StoreRefundTransferLight(message)
-                raiden.handle_and_track_state_change(store_refund)
+                state_change = ActionTransferRerouteLight(
+                    transfer=from_transfer,
+                    secret=random_secret(),
```
marcosmartinez7: There is something important here. 

On full nodes, the `ActionTransferReroute` receives a `secret ` parameter. That parameter is used to create a new locked transfer message (to try another route).

On full nodes the secret is created randomly by the initiator of the payment. The same occurs on light payments where the initiator is the light client, the secret is created by the initiator, **but, on the client side**. 

That means, that when the new locked transfer associated with the reroute is sent to the hub, the secret is already there. 

Therefore, the `secret `isnt needed as a param of the `ActionTransferRerouteLight `constructor

raiden/transfer/channel.py
```diff
@@ -1658,9 +1660,10 @@ def handle_receive_lock_expired_light(
         block_number=block_number,
     )
 
-    events: List[Event] = list()
+    events: List[Event]
```
marcosmartinez7, line 1663: Personally i dont like the idea of defining the variable and initialize it on a if-else block

```
events: List[Event]
if(something)
   events = [obj1, obj2]
else
   events = [obj2]
```

I think that initialize and collect is better

```
 events: List[Event] = []
if(something)
   events.append(obj1, obj2)
else
   events.append(obj2)
```

raiden/transfer/channel.py
```diff
@@ -1745,11 +1755,16 @@ def handle_receive_lockedtransfer_light(
         channel_state.partner_state.balance_proof = mediated_transfer.balance_proof
         channel_state.partner_state.merkletree = merkletree
 
+        print("raiden/transfer/channel.py:1806 PARTNER STATE UPDATED WITH BP {}".format(channel_state.partner_state.balance_proof))
```
marcosmartinez7: Remove this

raiden/transfer/channel.py
```diff
@@ -1671,11 +1674,18 @@ def handle_receive_lock_expired_light(
             payment_id,
             1,
             state_change.lock_expired,
-            True,
+            False,
```
marcosmartinez7: Why are you storing the message as unsinged? 

The reception of lock expired for light clients always receive a signed message by the LC

raiden/transfer/channel.py
```diff
@@ -1671,11 +1674,18 @@ def handle_receive_lock_expired_light(
...
             LightClientProtocolMessageType.PaymentExpired,
             state_change.lock_expired.recipient
         )
-        events = [store_lock_expired]
+
+        send_processed = SendProcessed(
```
marcosmartinez7: This isnt correct. 

Here we are on the situation that a HUB receives a lock expired msg that its recipient is a handled light client right? 

The ACK messages must be signed by the LC, not by the hub. Think about the SendProcessed handling 

```
    @staticmethod
    def handle_send_processed(raiden: "RaidenService", processed_event: SendProcessed):
        processed_message = message_from_sendevent(processed_event)
        raiden.sign(processed_message) # HERE THE HUB SIGNS THE MESSAGE
        raiden.transport.hub_transport.send_async(processed_event.queue_identifier, processed_message)
```

Basically the hub sends the ACK signing the message itself.

If this works is basically because the processed has the same id of the lock expired and it removes the L.E from the retry queue, but the protocol must be done between 2 participants that signs the messages, a 3rd one (the HUB on this case) must not sign any offchain msg.

This makes me think, we have already implemented the reception of L.E for light clients (basically the LC receives a L.E from the long polling and sends a delivered and processed message), so, i dont know why this was needed

raiden/transfer/channel.py
```diff
@@ -1745,11 +1755,16 @@ def handle_receive_lockedtransfer_light(
...
         channel_state.partner_state.merkletree = merkletree
 
+        print("raiden/transfer/channel.py:1806 PARTNER STATE UPDATED WITH BP {}".format(channel_state.partner_state.balance_proof))
+
         lock = mediated_transfer.lock
         channel_state.partner_state.secrethashes_to_lockedlocks[lock.secrethash] = lock
-        events = []
-
-
+        send_processed = SendProcessed(
```
marcosmartinez7: Same here.

raiden/transfer/mediated_transfer/initiator_manager.py
```diff
@@ -383,6 +390,100 @@ def handle_transferreroute(
...
 
 
+def handle_transfer_reroute_light(
+    payment_state: InitiatorPaymentState,
+    state_change: ActionTransferRerouteLight,
+    channel_identifiers_to_channels: ChannelMap,
+    storage: SerializedSQLiteStorage
+) -> TransitionResult[InitiatorPaymentState]:
+
+    print("raiden/transfer/mediated_transfer/initiator_manager.py:396 running handle_transfer_reroute_light")
```
marcosmartinez7: Remove this.

raiden/transfer/channel.py
marcosmartinez7: I mean, if the ACK were missing, i dont know how p2p payments works after a lock expired...

AFAIK here we dont create/dispatch SendProcessed or SendDelivered cause the LC sdk has the logic or responding with those messages when it receives a L.E

raiden/transfer/mediated_transfer/initiator_manager.py
```diff
@@ -383,6 +390,100 @@ def handle_transferreroute(
...
+        return TransitionResult(payment_state, list())
+
+    events: List[Event] = [store_refund_transfer]
+    events.extend(channel_events)
+
+    return TransitionResult(payment_state, events)
+
+
+def handle_lock_expired_light(
+    payment_state: InitiatorPaymentState,
```
marcosmartinez7, line 444: If this method wasnt implemented we need to think about p2p expirations.

What happened BEFORE this PR if we have this scenario?

1. LC wants to send a payment to FN
2. LC sends the payment but the hub goes offline or the lc stop polling
3. the payment expires
4. the partner sends a LE

marcosmartinez7: Lets talk with @debrayperalta and make a code walktrough of that scenario.

raiden/transfer/node.py
```diff
@@ -845,6 +848,20 @@ def handle_receive_transfer_refund_cancel_route(
     return subdispatch_to_paymenttask(chain_state, state_change, new_secrethash, storage)
 
 
+def handle_receive_transfer_refund_cancel_route_light(
+    chain_state: ChainState, state_change: ActionTransferRerouteLight, storage
```
marcosmartinez7, line 852: ActionTransferRerouteLight is handled by this function at node.py lvl but by handle_transfer_reroute_light at initiator lvl, i dont understand that.

raiden/transfer/node.py
```diff
@@ -845,6 +848,20 @@ def handle_receive_transfer_refund_cancel_route(
...
+) -> TransitionResult[ChainState]:
+
+    new_secrethash = state_change.secrethash
+    current_payment_task = chain_state.payment_mapping.secrethashes_to_task[
```
marcosmartinez7: This is already made at https://github.com/rsksmart/lumino/blob/develop/raiden/api/python.py#L1831

Basically, this is the way of create a new payment task associated with the same payment 

This isnt correct because we are doing it twice .

Also, im still not seeing the race condition that we talked about and where the payment task is removed, therefore a quick call will be good in order to continue the review of this pr.

marcosmartinez7 (changes requested):
We need to discuss and modify the Processed and Delivered handling for reception of lock expired.

We also need to check the copy of payment tasks on ActionTransferRerouteLight handling

raiden/lightclient/handlers/light_client_message_handler.py
```diff
@@ -74,8 +74,8 @@ def store_light_client_payment(cls, payment: LightClientPayment, storage: Serial
     def update_light_client_payment_status(cls, payment_id: int, status: LightClientPaymentStatus,
                                            storage: SerializedSQLiteStorage):
         exists_payment = storage.get_light_client_payment(payment_id)
-        if not exists_payment:
-            storage.update_light_client_payment_status(payment_id, status, storage)
+        if exists_payment:
+            storage.update_light_client_payment_status(payment_id, status)
```
mortelli: i know you didn't write it, but `exists_payment` is not a good variable name, because it is actually a payment (or `None`).

we can do:
```python
if storage.get_light_client_payment(payment_id):
    storage.update_light_client_payment_status(payment_id, status)
```
or
```python
payment = storage.get_light_client_payment(payment_id)
if payment:
    storage.update_light_client_payment_status(payment_id, status)
```

raiden/transfer/channel.py
```diff
@@ -1671,11 +1674,18 @@ def handle_receive_lock_expired_light(
             payment_id,
             1,
             state_change.lock_expired,
-            True,
+            False,
             LightClientProtocolMessageType.PaymentExpired,
             state_change.lock_expired.recipient
         )
```
mortelli: can you please explain why this changed from `True` to `False`?

also, do you think we could add named parameters here for better readability?

raiden/transfer/channel.py
mortelli: do we absolutely need this `print`?

raiden/transfer/channel.py
mortelli: can you explain the reason for this change? (empty list to non-empty list)

raiden/transfer/mediated_transfer/initiator_manager.py
mortelli: you probably didn't mean to leave this one here, right?

raiden/transfer/mediated_transfer/initiator_manager.py
```diff
@@ -383,6 +390,100 @@ def handle_transferreroute(
...
+        raise RaidenUnrecoverableError("Refund Transfer not found, state has changed with action "
+                                       "ActionTransferRerouteLight but refund_transfer param was not set")
+
+    store_refund_transfer = StoreMessageEvent(refund_transfer.message_identifier,
+                                              refund_transfer.payment_identifier,
+                                              1,
+                                              refund_transfer,
+                                              True,
+                                              LightClientProtocolMessageType.PaymentRefund,
+                                              refund_transfer.recipient)
```
mortelli: could you please add named parameters here for readability's sake?

raiden/transfer/mediated_transfer/initiator_manager.py
```diff
@@ -383,6 +390,100 @@ def handle_transferreroute(
...
+
+    try:
+        initiator_state = payment_state.initiator_transfers.get(state_change.transfer.lock.secrethash)
+        channel_identifier = initiator_state.channel_identifier
+        channel_state = channel_identifiers_to_channels[initiator_state.transfer.initiator].get(channel_identifier)
+    except KeyError:
+        return TransitionResult(payment_state, list())
```
mortelli, line 418: what is the logic behind this?

raiden/transfer/mediated_transfer/initiator_manager.py
```diff
@@ -710,6 +796,16 @@ def state_transition(
             block_number=block_number,
             storage=storage,
         )
+    elif type(state_change) == ActionTransferRerouteLight:
+        assert isinstance(state_change, ActionTransferRerouteLight), MYPY_ANNOTATION
+        msg = "ActionTransferRerouteLight should be accompanied by a valid payment state"
+        assert payment_state, msg
```
mortelli:
```suggestion
        assert payment_state, "ActionTransferRerouteLight should be accompanied by a valid payment state"
```

mortelli (commented):
as a general comment: i think the gif in the description is really cool. i would have found it helpful to also have a short design overview for the solution.

i couldn't really figure out why some things were deleted and others were added, or why some values were changed. i think a short, high-level explanation of how the problem is solved can help a lot with this type of stuff (it doesn't have to be perfectly written, just a quick overview).

also, please double check the branch names (the original issue number was wrong as well as the prefix).

(you also have 1 new alert for unused import on LGTM)

raiden/lightclient/handlers/light_client_message_handler.py
JONAF2103: This was a bug fix change, i din't change anything thinking on a refactor since i was focus on fix the bug rather than refactor something

JONAF2103: But i agree we can improve that

raiden/message_handler.py
JONAF2103: Ok i understand, so it should be something like this?

```            
if is_light_client:
                state_change = ActionTransferRerouteLight(
                    transfer=from_transfer,
                    secret=from_transfer.lock.secrethash,
                    refund_transfer=message
                )
                raiden.handle_and_track_state_change(state_change)
```

JONAF2103: Because if not, i don't know from where i should get the secret hash

raiden/transfer/channel.py
JONAF2103: Well the pycharm pep8 checker is indicating that the value of events: List[Event] = [] is useless since it will be always overwritten later, unless you put a condition on the else of course, but this is not the case, so basically it shows a warning saying that the empty list value is never used

raiden/transfer/channel.py
JONAF2103: This is not reception from the LC, is storing a message to the LC to be signed, we receive a lock expired from the mediator here, and the we store a message for the light client to say that the payment expired and we should send a lock expired, it was wrong since i was saying it was signed and it was failing for that since the LC hasn't signed that yet, it signs it when it send the message out not before that.

raiden/transfer/channel.py
JONAF2103: I can remove that if you want but not processed message will be out, the LC dont send that out, i checked that, i didn't knew that the processed message should be signed by the LC too, anyway is not being sent out at all so i don't know what to tell you here. I can remove it anyway

raiden/transfer/channel.py
JONAF2103: Sorry to many prints, i will do it

JONAF2103: No, i will remove it, it was used just for debugging

raiden/transfer/channel.py
JONAF2103: I did that above in the other case

JONAF2103: You refer to the empty list or the processed message, anyway i explain the cause of both above

raiden/transfer/mediated_transfer/initiator_manager.py
JONAF2103: yup removing it

raiden/transfer/mediated_transfer/initiator_manager.py
JONAF2103: Sorry this code was replicated from somewhere else, i will do

raiden/transfer/mediated_transfer/initiator_manager.py
JONAF2103: I didn't made this logic, but what i could understand is that since this is a state transition we don't want to have any exceptions here, in case something is wrong we just return an empty transition list, this is replicated all over the place and i don't know why it was made like that.

raiden/transfer/mediated_transfer/initiator_manager.py
JONAF2103: I'm not really sure about what happen in that case, in this case i know for sure that it wasn't handled right and that's why it wasn't working. We can review it with him sure.

raiden/lightclient/handlers/light_client_message_handler.py
mortelli: no worries, it's not a blocker :+1:

raiden/transfer/node.py
JONAF2103: i lost you here, can you explain more what you don't understand, baically ActionTransferRerouteLight was missing and it wasn't handled, you told me to add this to handle that to make the balance proof to be updated.

raiden/transfer/node.py
JONAF2103: Im lost here again, this is a replicate of the code above:

```
def handle_receive_transfer_refund_cancel_route(
    chain_state: ChainState, state_change: ActionTransferReroute, storage

    new_secrethash = state_change.secrethash
    current_payment_task = chain_state.payment_mapping.secrethashes_to_task[
        state_change.transfer.lock.secrethash
    ]
    chain_state.payment_mapping.secrethashes_to_task.update(
        {new_secrethash: copy.deepcopy(current_payment_task)}
    )
    return subdispatch_to_paymenttask(chain_state, state_change, new_secrethash, storage)
```
I just added that to support the new ActionTransferRerouteLight event that's all. Maybe this is related to the change of the secret hash that you mentioned to me above.

raiden/transfer/channel.py
JONAF2103: Because in that place the message is not signed by the LC yet, the message is being created for the LC.

raiden/message_handler.py
marcosmartinez7: Not really, you just dont need a secret here. When the lc generates the new payment, it will generate the new secret.

JONAF2103: Ok and how the node will manage the state change here:

```
def handle_receive_transfer_refund_cancel_route_light(
    chain_state: ChainState, state_change: ActionTransferRerouteLight, storage

    new_secrethash = state_change.secrethash
    current_payment_task = chain_state.payment_mapping.secrethashes_to_task[
        state_change.transfer.lock.secrethash
    ]
    chain_state.payment_mapping.secrethashes_to_task.update(
        {new_secrethash: copy.deepcopy(current_payment_task)}
    )
    return subdispatch_to_paymenttask(chain_state, state_change, new_secrethash, storage)
```
I mean i can remove the secret hash but it will never find the payment task unless you tell me that the ```raiden.handle_and_track_state_change(state_change)``` call from above is not triggering this code then i don't get how this is working at all.

raiden/transfer/channel.py
marcosmartinez7: This method is used on target.handle_lock_expired_light, to handle the ReceiveLockExpiredLight state_change.

The ReceiveLockExpiredLight state change is generated on message_handler.handle_message_lockexpired. 

That call is generated on message_handler.on_message, basically when you receive a LockExpired.

According to that flow, i dont understand why you think this isnt related with reception from LC.

raiden/transfer/channel.py
marcosmartinez7: Lets talk about this on the daily

raiden/transfer/channel.py
JONAF2103: Because in this case is also used on initiator_manager.handle_lock_expired_light, we added that because it was missing, and because of that we have the partner state being updated with the balance proof, before this that state was always none in the case of mediated payments.

raiden/transfer/node.py
marcosmartinez7: Accordly to what we have talked, we should remove the clone and just return the subdistpatch

raiden/message_handler.py
marcosmartinez7: We should use the refund transfer secret hash to receive the refund transfer into the channel

raiden/transfer/channel.py
marcosmartinez7: This msg is stored on true because is a signed message from the partner, and the lc should respond with a delivered and processed.

raiden/transfer/channel.py
mortelli: > Lets talk about this on the daily

can you guys please briefly document the resolution on this, if there is one?

raiden/transfer/mediated_transfer/initiator_manager.py
mortelli: thanks for the explanation

raiden/transfer/mediated_transfer/initiator_manager.py
mortelli: was this reviewed?

raiden/transfer/channel.py
marcosmartinez7: Basically we discussed why isnt correct and we must check why the LC was not responding to that messages.

raiden/transfer/mediated_transfer/initiator_manager.py
marcosmartinez7: Not yet

raiden/transfer/node.py
JONAF2103: We solve this in the google meet, i removed the clone code and the secret hash as we spoke

raiden/transfer/mediated_transfer/initiator_manager.py
JONAF2103: We changed this already as we spoke with @marcosmartinez7 in a meeting, i don't know if we want to review this with @debrayperalta anyway, please confirm @marcosmartinez7 when you can, thanks!

raiden/lightclient/handlers/light_client_message_handler.py
JONAF2103: Done

raiden/transfer/mediated_transfer/initiator_manager.py
JONAF2103: Done

JONAF2103:
@mortelli thanks for the review, as you told me here https://github.com/rsksmart/lumino/pull/125#pullrequestreview-502324108 i updated the description of this PR with a better content. 

"as a general comment: i think the gif in the description is really cool. i would have found it helpful to also have a short design overview for the solution." **at least we have a basic explanation of the problem now :)**

"i couldn't really figure out why some things were deleted and others were added, or why some values were changed. i think a short, high-level explanation of how the problem is solved can help a lot with this type of stuff (it doesn't have to be perfectly written, just a quick overview)" **this is related to the missing part of the handling code and the fixes that we refer on the description**

"also, please double check the branch names (the original issue number was wrong as well as the prefix)." **sorry about that it was my mistake, i didn't change it since we will lost all the comments here if i close this PR to create another one on the correct branch**

"(you also have 1 new alert for unused import on LGTM)" **this is fixed now**

mortelli:
> @mortelli thanks for the review, as you told me here [#125 (review)](https://github.com/rsksmart/lumino/pull/125#pullrequestreview-502324108) i updated the description of this PR with a better content.
> 
> "as a general comment: i think the gif in the description is really cool. i would have found it helpful to also have a short design overview for the solution." **at least we have a basic explanation of the problem now :)**
> 
> "i couldn't really figure out why some things were deleted and others were added, or why some values were changed. i think a short, high-level explanation of how the problem is solved can help a lot with this type of stuff (it doesn't have to be perfectly written, just a quick overview)" **this is related to the missing part of the handling code and the fixes that we refer on the description**
> 
> "also, please double check the branch names (the original issue number was wrong as well as the prefix)." **sorry about that it was my mistake, i didn't change it since we will lost all the comments here if i close this PR to create another one on the correct branch**
> 
> "(you also have 1 new alert for unused import on LGTM)" **this is fixed now**

TYVM :heart:

JONAF2103:
Closing this PR to continue the work on this other PR since here we have other changes that are not applying anymore and also the branch name was wrong, here is the new PR https://github.com/rsksmart/lumino/pull/132

PR #127, opened 2020-10-06
RLP-825: rename transport layer fields

mortelli:
This PR solves [RLP-825](https://jirainfuy.atlassian.net/browse/RLP-825), which is about renaming transport layer fields that include the term "hub".

### fields
- `hub_transport` → `full_node`
- `light_client_transports` → `light_clients`
- `_hub_transport` → `_full_node`
- `_light_client_transports` → `_light_clients`
### methods
- `new_light_client_transport` → `new_light_client` 
- `add_light_client_transport` → `add_light_client`
- `remove_light_client_transport` → `remove_light_client`

---

for most of these, they already a field/method for a _transport_ layer class, so it is redundant to include `transport` in the name.

### DIFF
--- a/raiden/network/transport/matrix/layer.py
+++ b/raiden/network/transport/matrix/layer.py
@@ -52,9 +52,9 @@ class MatrixLayer(TransportLayer):
 
             light_clients = storage.get_all_light_clients()
 
-            light_client_transports = []
-            for light_client in light_clients:
+            self._light_clients: List[TransportNode] = []
 
+            for light_client in light_clients:
                 current_server_name = None
 
                 if light_client["current_server_name"]:
@@ -83,10 +83,9 @@ class MatrixLayer(TransportLayer):
                     auth_params,
                 )
 
-                light_client_transports.append(light_client_transport)
+                self._light_clients.append(light_client_transport)
 
             self._full_node: TransportNode = MatrixTransportNode(config["address"], config["transport"]["matrix"])
-            self._light_clients: List[TransportNode] = light_client_transports
 
         except RaidenError as ex:
             click.secho(f"FATAL: {ex}", fg="red")
--- a/raiden/network/transport/matrix/layer.py
+++ b/raiden/network/transport/matrix/layer.py
@@ -85,27 +85,27 @@ class MatrixLayer(TransportLayer):
 
                 light_client_transports.append(light_client_transport)
 
-            self._hub_transport: TransportNode = MatrixTransportNode(config["address"], config["transport"]["matrix"])
-            self._light_client_transports: List[TransportNode] = light_client_transports
+            self._full_node: TransportNode = MatrixTransportNode(config["address"], config["transport"]["matrix"])
+            self._light_clients: List[TransportNode] = light_client_transports
 
         except RaidenError as ex:
             click.secho(f"FATAL: {ex}", fg="red")
             sys.exit(1)
 
     @property
-    def hub_transport(self) -> TransportNode:
-        return self._hub_transport
+    def full_node(self) -> TransportNode:
+        return self._full_node
 
     @property
-    def light_client_transports(self) -> List[TransportNode]:
-        return self._light_client_transports
+    def light_clients(self) -> List[TransportNode]:
+        return self._light_clients
 
     @staticmethod
-    def new_light_client_transport(address: Address, config: dict, auth_params: dict) -> TransportNode:
+    def new_light_client(address: Address, config: dict, auth_params: dict) -> TransportNode:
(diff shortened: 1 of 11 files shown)


JONAF2103:
I saw your changes on hub_transport to full_node member name, i think is not necessary to make any difference there, maybe we can just call it node, since the code will work for both (full and hub nodes) right?

marcosmartinez7:
> I saw your changes on hub_transport to full_node member name, i think is not necessary to make any difference there, maybe we can just call it node, since the code will work for both (full and hub nodes) right?

I think node is too wide and can be confused with a light client too. So i think light client transport and full node transport is clearer.

If we use node transport and light client transport i think that node is to generic/wide, because a light client is a node too.

JONAF2103:
Ok so for what you are saying then all the calls to raiden.transport.full_node is only for full nodes and not for hub nodes then? i just get confused about it since calls like ``def handle_send_balanceproof(raiden: "RaidenService", balance_proof_event: SendBalanceProof):`` are for both cases right? but maybe i just made a confusion on that i don't know

marcosmartinez7:
> Ok so for what you are saying then all the calls to raiden.transport.full_node is only for full nodes and not for hub nodes then? i just get confused about it since calls like `def handle_send_balanceproof(raiden: "RaidenService", balance_proof_event: SendBalanceProof):` are for both cases right? but maybe i just made a confusion on that i don't know

A Hub node is a Full node with extended capabilities.

But... im ok with using node too.

JONAF2103:
"A Hub node is a Full node with extended capabilities." is this totally true? i mean for example some of the rest apis are not available on hub mode, so i thought that hub mode had different capabilities than the full node that's why i thought they were different, but ok i got it.

mortelli:
> Ok so for what you are saying then all the calls to raiden.transport.full_node is only for full nodes and not for hub nodes then? i just get confused about it since calls like `def handle_send_balanceproof(raiden: "RaidenService", balance_proof_event: SendBalanceProof):` are for both cases right? but maybe i just made a confusion on that i don't know

even though FNs and hubs are different things, i think `full node` fits hubs better than `hub` fits full nodes.

_node_ could also work, but i think that when we mention _light clients_ we usually say **full** node to make contrast in this context (rather than just _node_) and it's these 2 fields that we have for the transport layer, basically. i also fear that `node` might be too generic.

i'm not entirely sure about this raiden service call you mention, though.

for the record though, i _also_ thought about just calling this **node** :laughing:

marcosmartinez7:
> "A Hub node is a Full node with extended capabilities." is this totally true? i mean for example some of the rest apis are not available on hub mode, so i thought that hub mode had different capabilities than the full node that's why i thought they were different, but ok i got it.

Yes it is in a theorical way, not for now.

I

raiden/network/transport/matrix/layer.py
```diff
@@ -85,27 +85,27 @@ def __init__(self, config: Dict[str, Any]):
 
                 light_client_transports.append(light_client_transport)
```
marcosmartinez7: Can we just use self._light_clients.append() insteadof create a new list and assign it to the member of the class?

raiden/raiden_service.py
```diff
@@ -998,14 +998,14 @@ def _initialize_whitelists(self, chain_state: ChainState):
...
         for event_queue in events_queues.values():
             for event in event_queue:
                 if isinstance(event, SendLockedTransfer):
                     transfer = event.transfer
                     if transfer.initiator == self.address:
-                        self.transport.hub_transport.whitelist(address=transfer.target)
+                        self.transport.full_node.whitelist(address=transfer.target)
                 if isinstance(event, SendLockedTransferLight):
                     transfer = event.signed_locked_transfer
                     for light_client_transport in light_client_transports:
```
marcosmartinez7: i think we can just iterate over self.transport.light_clients here

marcosmartinez7 (approved):
Lgtm, just some changes related with old code that im not considering as blockers

raiden/raiden_service.py
mortelli: done

raiden/network/transport/matrix/layer.py
mortelli: done

mortelli:
- verified [manual scenarios](https://github.com/rsksmart/lumino/pull/110#issuecomment-702364526) :heavy_check_mark: 
- verified [unit tests](https://github.com/anarancio/lumino-docs/blob/master/unit-testing/results/transport-generalization/01-10-2020/summary.txt) :heavy_check_mark:

PR #128, opened 2020-10-06
RLP-818 Ignoring processed messages for LC

JONAF2103:
In this PR we are:

* Adding query condition to the long polling to get only the not processed messages by the LC for the specific types ('SettlementRequired', 'RequestRegisterSecret', 'UnlockLightRequest').
* Adding a new aux method for sql updates, we are repeating the code all over the place, i just extracted that code to an aux method to call that instead
* Adding a new method to update a protocol message by internal id

This affects the SDK Long Polling

### DIFF
--- a/raiden/lightclient/handlers/light_client_message_handler.py
+++ b/raiden/lightclient/handlers/light_client_message_handler.py
@@ -54,15 +54,18 @@ class LightClientMessageHandler:
         )
 
     @classmethod
-    def update_stored_msg_set_signed_data(
+    def update_offchain_light_client_protocol_message_set_signed_message(
         cls, message: Message,
         payment_id: int,
         order: int,
         message_type: LightClientProtocolMessageType,
         wal: WriteAheadLog
     ):
-        return wal.storage.update_light_client_protocol_message_set_signed_data(payment_id, order, message,
-                                                                                str(message_type.value))
+        return wal.storage\
+            .update_offchain_light_client_protocol_message_set_signed_message(payment_id,
+                                                                              order,
+                                                                              message,
+                                                                              str(message_type.value))
 
     @classmethod
     def store_light_client_payment(cls, payment: LightClientPayment, storage: SerializedSQLiteStorage):
--- a/raiden/storage/sqlite.py
+++ b/raiden/storage/sqlite.py
@@ -1376,24 +1376,21 @@ class SQLiteStorage:
         cursor = self.conn.cursor()
         cursor.execute(
             """
-            SELECT lcpm1.identifier,
-                   lcpm1.message_order,
-                   lcpm1.unsigned_message,
-                   lcpm1.signed_message,
-                   lcpm1.light_client_payment_id,
-                   lcpm1.internal_msg_identifier,
-                   lcpm1.message_type
-            FROM light_client_protocol_message lcpm1
-            WHERE lcpm1.internal_msg_identifier >= ?
-            AND lcpm1.light_client_address = ?
-            AND lcpm1.internal_msg_identifier NOT IN (
-                SELECT lcpm2.internal_msg_identifier
-                FROM light_client_protocol_message lcpm2
-                WHERE lcpm2.internal_msg_identifier >= ?
-                AND lcpm2.message_type IN ('SettlementRequired', 'RequestRegisterSecret', 'UnlockLightRequest')
-                AND lcpm2.signed_message IS NOT NULL
+            SELECT identifier,
+                   message_order,
+                   unsigned_message,
+                   signed_message,
+                   light_client_payment_id,
+                   internal_msg_identifier,
+                   message_type
+            FROM light_client_protocol_message
+            WHERE internal_msg_identifier >= ?
+            AND light_client_address = ?
+            AND (
+                message_type NOT IN ('SettlementRequired', 'RequestRegisterSecret', 'UnlockLightRequest')
+                OR signed_message IS NULL
             )
-            ORDER BY lcpm1.light_client_payment_id, lcpm1.message_order ASC
+            ORDER BY light_client_payment_id, message_order ASC
             """,
             (from_message, light_client),
         )
@@ -1474,7 +1471,7 @@ class SerializedSQLiteStorage(SQLiteStorage):
                      serialized_signed_msg, message[0], message[5], message[6]))
         return result
 
-    def update_light_client_protocol_message_set_signed_data(self, payment_id, msg_order, signed_message, message_type):
+    def update_offchain_light_client_protocol_message_set_signed_message(self, payment_id, msg_order, signed_message, message_type):
         with self.write_lock, self.conn:
             cursor = self.conn.cursor()
             cursor.execute(
@@ -1487,9 +1484,9 @@ class SerializedSQLiteStorage(SQLiteStorage):
             last_id = cursor.lastrowid
         return last_id
 
-    def update_light_client_protocol_message_with_signed_message(self,
-                                                                 internal_msg_identifier: int,
-                                                                 signed_message: "Message"):
+    def update_onchain_light_client_protocol_message_set_signed_transaction(self,
+                                                                            internal_msg_identifier: int,
+                                                                            signed_message: "Message"):
         return self.update(
             """
                 UPDATE light_client_protocol_message
--- a/raiden/storage/sqlite.py
+++ b/raiden/storage/sqlite.py
@@ -8,7 +8,6 @@ from eth_utils import to_checksum_address
 from raiden.constants import RAIDEN_DB_VERSION, SQLITE_MIN_REQUIRED_VERSION
 from raiden.exceptions import InvalidDBData, InvalidNumberInput
 from raiden.lightclient.models.client_model import ClientType
-from raiden.messages import Message
 from raiden.storage.serialize import SerializationBase
 from raiden.storage.utils import DB_SCRIPT_CREATE_TABLES, TimestampedEvent
 from raiden.utils import get_system_spec
@@ -1490,7 +1489,7 @@ class SerializedSQLiteStorage(SQLiteStorage):
 
     def update_light_client_protocol_message_with_signed_message(self,
                                                                  internal_msg_identifier: int,
-                                                                 signed_message: Message):
+                                                                 signed_message: "Message"):
         return self.update(
             """
                 UPDATE light_client_protocol_message
--- a/raiden/storage/sqlite.py
+++ b/raiden/storage/sqlite.py
@@ -8,6 +8,7 @@ from eth_utils import to_checksum_address
 from raiden.constants import RAIDEN_DB_VERSION, SQLITE_MIN_REQUIRED_VERSION
 from raiden.exceptions import InvalidDBData, InvalidNumberInput
 from raiden.lightclient.models.client_model import ClientType
+from raiden.messages import Message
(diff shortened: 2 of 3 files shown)


raiden/storage/sqlite.py
```diff
@@ -1474,6 +1488,18 @@ def update_light_client_protocol_message_set_signed_data(self, payment_id, msg_o
             last_id = cursor.lastrowid
         return last_id
 
+    def update_light_client_protocol_message_with_signed_message(self,
```
marcosmartinez7: We already have a `update_light_client_protocol_message_set_signed_data ` method

The only difference is that this one receives the msg order and type. 

Maybe we can rename this new method to `update_light_client_protocol_message_set_signed_data_by_id`

Or just keep one of them.

JONAF2103: Probably we will need to remove the older method or rename this one but it will depend on the path we want to take from now to do those updates, actually renaming this to `update_light_client_protocol_message_set_signed_data_by_id` could be a problem since which id? the internal id like this one does or the message_identifier like we use in other places? that's why i didn't make any refactor on that, i agree that we should have only one method for this but we don't have the internal identifier on the cases that are being used right now.

raiden/storage/sqlite.py
```diff
@@ -1376,11 +1376,24 @@ def get_light_client_messages(self, from_message, light_client):
...
+            FROM light_client_protocol_message lcpm1
+            WHERE lcpm1.internal_msg_identifier >= ?
+            AND lcpm1.light_client_address = ?
+            AND lcpm1.internal_msg_identifier NOT IN (
+                SELECT lcpm2.internal_msg_identifier
+                FROM light_client_protocol_message lcpm2
+                WHERE lcpm2.internal_msg_identifier >= ?
+                AND lcpm2.message_type IN ('SettlementRequired', 'RequestRegisterSecret', 'UnlockLightRequest')
+                AND lcpm2.signed_message IS NOT NULL
+            )
```
mauricioirace:
```suggestion
             AND (NOT lcpm1.message_type IN ('SettlementRequired', 'RequestRegisterSecret', 'UnlockLightRequest')
             OR lcpm1.signed_message IS NULL)
```
(not p or q is equivalent to p -> q, preventing us from the inner select I think)

mauricioirace (changes requested):
LGTM but I don't like the inner nested select

raiden/storage/sqlite.py
marcosmartinez7: It should be

```
AND (NOT lcpm1.message_type IN ('SettlementRequired', 'RequestRegisterSecret', 'UnlockLightRequest')
             AND lcpm1.signed_message IS NOT NULL)
```

right? 

Ignore the settlement required, register secret and unlock light when signed message is already present

raiden/storage/sqlite.py
marcosmartinez7: The `update_light_client_protocol_message_set_signed_data  `is related with a payment because is for offchain msgs

The new method `update_light_client_protocol_message_with_signed_message ` isnt always related with a payment (register secret is but unlock and settlement not) 

Maybe we can rename both methods to:

- `update_offchain_light_client_protocol_message_set_signed_data(payment,order, type,msg)` and 
- `update_onchain_light_client_protocol_message_set_signed_data(id)`

what do u guys think? 

@mauricioirace @mortelli @JONAF2103

mauricioirace: I agree with your proposal @marcosmartinez7 
Maybe we could use something like:
`update_offchain_light_client_protocol_message_set_signed_message` and `update_onchain_light_client_protocol_message_set_signed_transaction` to tell the function caller what should it send as signed data. But any of the two approaches seems fine to me.

marcosmartinez7: Agree!

raiden/storage/sqlite.py
JONAF2103: Your proposal isn't quiet right @mauricioirace with your SQL you are saying that you want all messages that have not message type in ('SettlementRequired', 'RequestRegisterSecret', 'UnlockLightRequest') or signed_message null, so if you have a message with type SettlementRequired or you have a message with signed_message not null this query will not retrieve that, basically you need to query a subset and retrieve all the records that are not in that set for the particular case (not always), make sense? that's why i used the nested query, also this is SQLite we don't have all the capabilities from SQL to make the query the best way possible, a better approach to this could be using UNION for example but doesn't work well with SQLite.

marcosmartinez7: What about
```

AND (NOT lcpm1.message_type IN ('SettlementRequired', 'RequestRegisterSecret', 'UnlockLightRequest')
             ABD lcpm1.signed_message IS NOT NULL)

```
?

marcosmartinez7: No, it doesnt work, it doesnt count the payments msgs.

mauricioirace: If have a message that is SettlementRequired, it will retrieved if it has not been signed (signed_message is null). On the other hand, if I have a message that has been signed, it will be retrieved only if it's not one of that list. 

I think that's correct, isn't it?

JONAF2103: As we review it you are right @mauricioirace sound good, i changed the sql to that proposal, thanks!

marcosmartinez7 (approved):

mauricioirace (approved):
LGTM

raiden/storage/sqlite.py
```diff
@@ -1617,3 +1639,11 @@ def flag_light_client_as_pending_for_deletion(self, address):
...
 
         return last_id
+
+    def update(self, update_sql: str, params: Any) -> int:
+        """ Aux method to generalize the update calls """
+        with self.write_lock, self.conn:
+            cursor = self.conn.cursor()
+            cursor.execute(update_sql, params)
+            last_id = cursor.lastrowid
+        return last_id
```
mortelli, line 1649: why is this called `update`? couldn't it just execute _any_ SQL it receives?

raiden/lightclient/handlers/light_client_message_handler.py
```diff
@@ -54,15 +54,18 @@ def store_light_client_protocol_message(cls, identifier: int, message: Message,
         )
 
     @classmethod
-    def update_stored_msg_set_signed_data(
+    def update_offchain_light_client_protocol_message_set_signed_message(
```
mortelli, line 57: i don't think this name makes too much sense. it's 9 (!) words long and redundant.

to start with we have `update offchain light client protocol message`:
- the word `protocol` doesn't really add much (`light client message` would be just as clear).
- i've read @marcosmartinez7's comment, but i don't think it's necessary to say whether this is off-chain or on-chain.

and then `set signed message`: we already knew this was an update, but now we are also saying this writes a specific field. so why say _update_ in the first place?

i would've named this `set_light_client_message_signed_message` or `set_signed_message_for_light_client_message`.

raiden/storage/sqlite.py
```diff
@@ -1474,6 +1484,18 @@ def update_light_client_protocol_message_set_signed_data(self, payment_id, msg_o
             last_id = cursor.lastrowid
         return last_id
 
+    def update_onchain_light_client_protocol_message_set_signed_transaction(self,
```
mortelli, line 1487: this says _set transaction_ but it receives a signed _message_

this isn't structurally equivalent to the other call :man_shrugging:

raiden/storage/sqlite.py
```diff
@@ -1376,10 +1376,20 @@ def get_light_client_messages(self, from_message, light_client):
...
+                   unsigned_message,
+                   signed_message,
+                   light_client_payment_id,
+                   internal_msg_identifier,
+                   message_type
             FROM light_client_protocol_message
             WHERE internal_msg_identifier >= ?
             AND light_client_address = ?
+            AND (
+                message_type NOT IN ('SettlementRequired', 'RequestRegisterSecret', 'UnlockLightRequest')
```
mortelli, line 1390: question here, what about messages that:
1. have a type ∉ (`SettlementRequired`, `RequestRegisterSecret`, `UnlockLightRequest`), but **also**:
2. have `signed_message IS NOT NULL`

do you want to retrieve these?

mortelli (commented):
signatures aren't my favourite, but i don't see any blockers.

just 1 question regarding the SQL

raiden/lightclient/handlers/light_client_message_handler.py
JONAF2103: I mean if we are going to be strict on this well, `set_light_client_message_signed_message` is not good neither, since we are inside the LightClientMessageHandler then we should just call this method `set_signed_data` so the call could be `LightClientMessageHandler.set_signed_data`  but i don't know i can change it again, i agree that the new name is really large but we need to stablish a criteria on this or each time i will need to change it several times.

mortelli: it's fine, leave it like this :+1:  

just keep this in mind for next time, and i suggest we just be bold and change it to something like `set_signed_data` directly. 

your suggestion is actually ideal :clap:

raiden/storage/sqlite.py
JONAF2103: Well according to the crew, seems that those messages are being retrieved too since the first condition of the OR should be TRUE and that should work.

raiden/storage/sqlite.py
JONAF2103: Basically this 2 methods are totally different in terms of business, are for 2 different use cases (payment related and not) so i guess the guys think those names are ok, anyway again, i think we should stablish some sort of criteria here because my first suggestion wasn't good neither.

raiden/storage/sqlite.py
JONAF2103: actually it just executes an update, since it gets the lastrowid, we could improve this to be more generic style to work not just with updates, maybe return the cursor, i don't know but at least this is better than copy/pasting the segment of code to query the database, at least the method call is cleaner.

raiden/lightclient/handlers/light_client_message_handler.py
marcosmartinez7: Im ok with the two proposals, just wanted to differenciate on the signature offchain from onchain.

raiden/storage/sqlite.py
marcosmartinez7: Yes you want.

That scenarios are, for example, when you create a payment from a light client.

The sdk invokes the create payment endpoint and creates an unsigned lockedtransfer. 

Then the sdk gets that msg from the polling and sign it.

raiden/lightclient/handlers/light_client_message_handler.py
JONAF2103: let's be bold :)

mortelli (approved):

raiden/lightclient/handlers/light_client_message_handler.py
JONAF2103: Ok as we spoke we have an issue already to rename and refactor the class LightClientMessageHandler so we are going to keep things like this for now and make all the refactor on that issue.

raiden/storage/sqlite.py
mauricioirace: I had not noticed this before, but indeed, if we receive a signed transaction, the typing won't work, will it?

PR #129, opened 2020-10-06
Feature/rlp 824/set signed transaction for settlement light

JONAF2103:
In this PR we are:

* Adding new method to get a message from the internal identifier
* Adding a check to the settlement light to check if the message exists before we start processing the request
* Adding the update with the signed transaction call to update the message after light settlement transaction is sent out

This affects the LC messages and LC settlement process

### DIFF
--- a/raiden/api/rest.py
+++ b/raiden/api/rest.py
@@ -1667,7 +1667,7 @@ class RestAPI:
 
         if message.is_signed:
             return ApiErrorBuilder.build_and_log_error(
-                errors="Failed trying to settle a channel that's already settled",
+                errors="Message already signed",
                 status_code=HTTPStatus.CONFLICT,
                 log=log
             )
--- a/raiden/api/rest.py
+++ b/raiden/api/rest.py
@@ -1621,6 +1621,7 @@ class RestAPI:
 
     def settlement_light(self,
                          registry_address: typing.PaymentNetworkID,
+                         internal_msg_identifier: int,
                          token_address: typing.TokenAddress,
                          creator_address: typing.Address,
                          partner_address: typing.Address,
@@ -1631,6 +1632,7 @@ class RestAPI:
             "Settling channel light",
             node=pex(creator_address),
             registry_address=to_checksum_address(registry_address),
+            internal_msg_identifier=internal_msg_identifier,
             token_address=to_checksum_address(token_address),
             creator_address=to_checksum_address(creator_address),
             partner_address=to_checksum_address(partner_address),
@@ -1650,9 +1652,23 @@ class RestAPI:
             )
             return result
 
+        message = LightClientMessageHandler.get_light_client_protocol_message_by_internal_identifier(
+            internal_msg_identifier=internal_msg_identifier,
+            wal=self.raiden_api.raiden.wal
+        )
+
+        if not message:
+            return ApiErrorBuilder.build_and_log_error(
+                errors="Light Client Message with internal_msg_identifier = {} not found"
+                       .format(internal_msg_identifier),
+                status_code=HTTPStatus.BAD_REQUEST,
+                log=log
+            )
+
         try:
             channel_state = self.raiden_api.channel_settle_light(
                 registry_address=registry_address,
+                internal_msg_identifier=internal_msg_identifier,
                 token_address=token_address,
                 creator_address=creator_address,
                 partner_address=partner_address,
--- a/raiden/api/rest.py
+++ b/raiden/api/rest.py
@@ -1665,10 +1665,22 @@ class RestAPI:
                 log=log
             )
 
+        if message.is_signed:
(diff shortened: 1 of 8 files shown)


raiden/api/rest.py
```diff
@@ -1650,6 +1652,32 @@ def settlement_light(self,
...
+            return ApiErrorBuilder.build_and_log_error(
+                errors="Light Client Message with internal_msg_identifier = {} not found"
+                       .format(internal_msg_identifier),
+                status_code=HTTPStatus.BAD_REQUEST,
+                log=log
+            )
+
+        if message.is_signed:
+            return ApiErrorBuilder.build_and_log_error(
+                errors="Failed trying to settle a channel that's already settled",
```
marcosmartinez7: I think this err should be that the message is already signed,not that the channel is already settled. The settlement may not happen cause an error.

marcosmartinez7 (approved):
lgtm

raiden/api/rest.py
JONAF2103: yes that's true i did it like this to avoid having to add a new error handling on the SDK but we can speak with @debrayperalta to add that to the SDK and change this message

raiden/api/rest.py
```diff
@@ -1621,6 +1621,7 @@ def initiate_payment_light(
 
     def settlement_light(self,
                          registry_address: typing.PaymentNetworkID,
+                         internal_msg_identifier: int,
```
mortelli, line 1624: a few questions here—i am trying to wrap my head around this parameter:
- will this internal identifier _always_ be provided from the caller?
  - if so: how did we manage without this field before?
- doesn't this come from an API call? why is it called _internal_ if it comes from an API caller?

raiden/api/rest.py
mortelli: i agree with @marcosmartinez7

raiden/api/rest.py
```diff
@@ -1650,6 +1652,32 @@ def settlement_light(self,
...
+            )
+
+        if message.is_signed:
+            return ApiErrorBuilder.build_and_log_error(
+                errors="Failed trying to settle a channel that's already settled",
+                status_code=HTTPStatus.CONFLICT,
+                log=log
+            )
+
+        LightClientMessageHandler.update_onchain_light_client_protocol_message_set_signed_transaction(
```
mortelli, line 1675: we are ignoring the returned value here

raiden/storage/sqlite.py
```diff
@@ -1400,6 +1410,26 @@ def get_light_client_protocol_message_by_identifier(self, identifier):
...
+        cursor.execute(
+            """
+            SELECT identifier,
+                   message_order,
+                   unsigned_message,
+                   signed_message,
+                   light_client_payment_id,
+                   message_type,
+                   light_client_address,
+                   internal_msg_identifier
```
mortelli, line 1424: what is the difference between `identifier` and `internal_msg_identifier`?

it seems very suspicious to me that we are using an apparently secondary id in this `WHERE` clause

mortelli (commented):
code looks good, just have a few business logic questions regarding this new internal identifier

raiden/api/rest.py
JONAF2103: Is the name of the parameter for identify a message from the LC, basically we didn't change the name since we don't want to add that change to the SDK, the SDK receives that parameter and send it to us as it is again, so we can identify the message in the database, this is new, before this we have the problem where we don't know how to retrieve or identify a message that the LC has signed.

raiden/api/rest.py
JONAF2103: because we don't need it, in case of some update error this operation will repeat, but if this fails probably the database is not working and the lumino node shouldn't be working neither

raiden/storage/sqlite.py
JONAF2103: identifier is the message random identifier generated for the payment process, the internal_msg_identifier is the identifier generated by the database, 2 different things

raiden/api/rest.py
mortelli: got it, thanks.

raiden/api/rest.py
mortelli: i've created issue [RLP-843](https://jirainfuy.atlassian.net/browse/RLP-843) to deal with this.

raiden/storage/sqlite.py
mortelli: i've created issue [RLP-843](https://jirainfuy.atlassian.net/browse/RLP-843) to deal with this.

mortelli (approved):
please merge once you have made sure that tests are OK :+1:

JONAF2103:
Testing done, everything works as expected

PR #131, opened 2020-10-12
Fix/rlp 822/signed register secret

mauricioirace:
Changes:
- Check if register secret transaction is already signed before sending it to the blockchain.

### DIFF
--- a/raiden/api/rest.py
+++ b/raiden/api/rest.py
@@ -900,15 +900,14 @@ class RestAPI:
 
             if not message:
                 return ApiErrorBuilder.build_and_log_error(
-                    errors="Light Client Message with internal_msg_identifier = {} not found"
-                        .format(internal_msg_identifier),
+                    errors=f"Light Client Message with internal_msg_identifier = {internal_msg_identifier} not found",
                     status_code=HTTPStatus.BAD_REQUEST,
                     log=log
                 )
 
             if message.is_signed:
                 return ApiErrorBuilder.build_and_log_error(
-                    errors="Message already signed",
+                    errors="Light Client Message with internal_msg_identifier = {internal_msg_identifier} already signed",
                     status_code=HTTPStatus.CONFLICT,
                     log=log
                 )
--- a/raiden/api/rest.py
+++ b/raiden/api/rest.py
@@ -891,8 +891,34 @@ class RestAPI:
         return api_response(result=closed_channels)
 
     @requires_api_key
-    def register_secret_light(self, signed_tx: typing.SignedTransaction):
+    def register_secret_light(self, internal_msg_identifier: int, signed_tx: typing.SignedTransaction):
         try:
+            message = LightClientMessageHandler.get_light_client_protocol_message_by_internal_identifier(
+                internal_msg_identifier=internal_msg_identifier,
+                wal=self.raiden_api.raiden.wal
+            )
+
+            if not message:
+                return ApiErrorBuilder.build_and_log_error(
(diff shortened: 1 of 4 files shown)


raiden/api/rest.py
```diff
@@ -891,8 +891,34 @@ def leave(self, registry_address: typing.PaymentNetworkID, token_address: typing
...
+                )
+
+            if message.is_signed:
+                return ApiErrorBuilder.build_and_log_error(
+                    errors="Message already signed",
+                    status_code=HTTPStatus.CONFLICT,
+                    log=log
+                )
+
+            LightClientMessageHandler.update_onchain_light_client_protocol_message_set_signed_transaction(
```
mortelli, line 915: this might be out of scope for this PR, but why does `update_onchain_light_client_protocol_message_set_signed_transaction` even have a return?

it seems we are not using it on any call

raiden/api/rest.py
```diff
@@ -891,8 +891,34 @@ def leave(self, registry_address: typing.PaymentNetworkID, token_address: typing
...
+                    status_code=HTTPStatus.BAD_REQUEST,
+                    log=log
+                )
+
+            if message.is_signed:
+                return ApiErrorBuilder.build_and_log_error(
+                    errors="Message already signed",
+                    status_code=HTTPStatus.CONFLICT,
+                    log=log
+                )
```
mortelli: if the `BAD_REQUEST` error includes the internal message identifier in its error text, then the `CONFLICT` error should probably do so as well, no?

raiden/api/rest.py
```diff
@@ -1671,6 +1699,32 @@ def settlement_light(self,
...
+                status_code=HTTPStatus.BAD_REQUEST,
+                log=log
+            )
+
+        if message.is_signed:
+            return ApiErrorBuilder.build_and_log_error(
+                errors="Message already signed",
+                status_code=HTTPStatus.CONFLICT,
+                log=log
+            )
```
mortelli: same here about the error txt and internal id

raiden/lightclient/handlers/light_client_message_handler.py
```diff
@@ -68,10 +68,16 @@ def update_offchain_light_client_protocol_message_set_signed_message(
                                                                               str(message_type.value))
 
     @classmethod
-    def update_stored_msg_set_signed_tx_by_message_id(
-        cls, message_id: MessageID, signed_tx: SignedTransaction, wal: WriteAheadLog
+    def update_onchain_light_client_protocol_message_set_signed_transaction(
```
mortelli: there's an unused import in this file

mortelli (commented):
no major issues.

since this PR merges to something other than `develop`, i wouldn't worry too much about getting multiple approvals.

raiden/api/rest.py
mauricioirace: I agree, I think a return is not needed, but, I also agree it's out of the scope of this PR

JONAF2103 (approved):
LGTM

raiden/api/rest.py
mortelli: thanks for changing this :clap: 

would you mind also having it this way in the `settlement_light` function? 

i'm asking because it's in the same file and i don't think it's too big of a change not to have it. 

it would help consistency, i think

mortelli (approved):
LGTM, just [1 minor remark](https://github.com/rsksmart/lumino/pull/131#discussion_r503481223)

PR #133, opened 2020-10-13
RLP-715/Possible fix LC-FN-LC Pay

marcosmartinez7:
This PR aims to solve the same hub LC to LC mediated payments. 

Suposse the following topology: 

`LC1 --- FN --- LC2
`
A mediated transfer from LC1 to LC2 implies two payments, LT1 (LC1 to FN) and LT2 (FN to LC2)

The raiden protocol establsih that two LT that are part of the same mediated transfer have the same Payment_identifier, therefore the LT received by the hub acting as a target node will not be stored. 

Why isnt stored? The msg existence query checks for: 

- same payment id (true)
- same msg order (true, the msg order is 1 for LT)
- same type (true, locked transfer)

Solution: 

Check also for message_identifier to determine if it is the same msg or not.

Note: this happens only If `LC1 `and `LC2 `are registered on the same hub

There is also another modification needed: 

- message `5 (SecretRequest) `and` 7 (RevealSecret) `are msgs between the `LC1 `and `LC2`, therefore must be assigned to both light clients.

Check comments about `handle_send_secret_reveal_light ` and `handle_inittarget_light `on the code

### DIFF
--- a/raiden/transfer/mediated_transfer/initiator.py
+++ b/raiden/transfer/mediated_transfer/initiator.py
@@ -560,7 +560,7 @@ def handle_send_secret_reveal_light(
 
     initiator_state.revealsecret = revealsecret
     initiator_state.received_secret_request = True
-    receiver_light_client_address = state_change.receiver if channel_state.both_participants_are_light_clients else None
+    receiver_light_client_address = state_change.receiver
 
     store_message_event = StoreMessageEvent(message_identifier, transfer_description.payment_identifier, 7,
                                             state_change.reveal_secret, True,
--- a/raiden/transfer/mediated_transfer/target.py
+++ b/raiden/transfer/mediated_transfer/target.py
@@ -243,7 +243,7 @@ def handle_inittarget_light(
                                                             receiver_light_client_address=transfer.target)
 
             secret_request_message = SecretRequest.from_event(secret_request)
-            sr_receiver_light_client_address = transfer.initiator if channel_state.both_participants_are_light_clients else None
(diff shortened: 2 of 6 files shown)


raiden/transfer/mediated_transfer/target.py
```diff
@@ -243,7 +243,7 @@ def handle_inittarget_light(
                                                             receiver_light_client_address=transfer.target)
 
             secret_request_message = SecretRequest.from_event(secret_request)
-            sr_receiver_light_client_address = transfer.initiator if channel_state.both_participants_are_light_clients else None
+            sr_receiver_light_client_address = transfer.initiator
```
marcosmartinez7, line 246: This cannot be always the transfer.initiator because on payments between LC and FN the initiator can be a not registered client, therefore the client FK will fail

raiden/transfer/mediated_transfer/initiator.py
```diff
@@ -560,7 +560,7 @@ def handle_send_secret_reveal_light(
 
     initiator_state.revealsecret = revealsecret
     initiator_state.received_secret_request = True
-    receiver_light_client_address = state_change.receiver if channel_state.both_participants_are_light_clients else None
+    receiver_light_client_address = state_change.receiver
```
marcosmartinez7, line 563: This cannot be always the state_change.receiver because on payments between LC and FN the initiator can be a not registered client, therefore the client FK will fail

mortelli:
putting this on hold since #145 might fix it

marcosmartinez7:
Solved on #145

PR #135, opened 2020-10-16
RLP-840: RIF Comms base

mortelli:
This is the first PR for issue [RLP-840](https://jirainfuy.atlassian.net/browse/RLP-840), which is about implementing the RIF Comms transport layer.

### Changes
This PR:
- moves `raiden.network.transport` to `transport`.
- implements skeleton classes for the RIF Comms transport layer implementation.
- applies _Reformat Code_ + _Optimize Imports_ to all affected files.


JONAF2103 (commented):
It looks good for now, just some comments:

1. The methods that are not implemented; maybe we can add a raise not implemented error instead of leaving them empty. Node, MessageParams and Layer classes have empty methods.
2. The file  transport/udp/__init__.py is really necesary?

marcosmartinez7 (approved):

mortelli:
> It looks good for now, just some comments:
> 
>     1. The methods that are not implemented; maybe we can add a raise not implemented error instead of leaving them empty. Node, MessageParams and Layer classes have empty methods.

this is a great idea :+1:  

i've gone ahead and done it for the `RifCommsLayer` class. however, i was not able to find any empty methods in the `MessageParams` class.

>     2. The file  transport/udp/**init**.py is really necesary?

~~yes. we need to have `udp` as a package to import code from `udp/transport.py`, `udp/utils.py` and `udp/healthcheck.py` elsewhere in the codebase (which is currently happening in several places).~~

turns out we don't _strictly_ need this, but it seems like a good idea anyway. read up on this [here](https://stackoverflow.com/questions/32152373/python-why-can-i-import-modules-without-init-py-at-all/42879905)

mortelli:
unit test results [here](https://github.com/anarancio/lumino-docs/blob/master/unit-testing/results/feature/RLP-840_rif-comms-base/summary.txt)

PR #137, opened 2020-10-19
bugfix/RLP-836/cannot-receive-payments

mauricioirace:
Change log:
- Broadcast some channel state changes to multiple channels,to prevent issuees caused by channel duplication

### DIFF
--- a/raiden/transfer/token_network.py
+++ b/raiden/transfer/token_network.py
@@ -29,63 +29,73 @@ StateChangeWithChannelID = Union[
     ContractReceiveUpdateTransfer,
 ]
 
-# State changes that should be duplicated in case there are two light clients
-DuplicableStateChanges = [
-    ContractReceiveChannelNewBalance
-]
-def subdispatch_to_channel_by_id_and_address(
+
+def subdispatch_to_channels_by_participant_address(
     token_network_state: TokenNetworkState,
     state_change: StateChangeWithChannelID,
     block_number: BlockNumber,
     block_hash: BlockHash,
-    node_address: AddressHex = None
+    participant_address: AddressHex = None
 ) -> TransitionResult:
-    events = list()
+    if participant_address is None:
+        return TransitionResult(token_network_state, [])
+    channel_states = get_channels_to_dispatch_statechange(
+        participant_address, state_change, token_network_state
+    )
+    return subdispatch_to_channels(block_hash, block_number, channel_states, state_change, token_network_state)
 
+
+def subdispatch_to_channels(block_hash, block_number, channel_states, state_change, token_network_state):
+    events = []
+    ids_to_channels = token_network_state.channelidentifiers_to_channels
+    for channel_state in channel_states:
+        result = channel.state_transition(
+            channel_state=channel_state,
+            state_change=state_change,
+            block_number=block_number,
+            block_hash=block_hash,
+        )
+
+        partner_to_channelids = token_network_state.partneraddresses_to_channelidentifiers[
+            channel_state.partner_state.address
+        ]
+
+        channel_identifier = state_change.channel_identifier
+        if result.new_state is None:
+            del ids_to_channels[channel_state.our_state.address][channel_identifier]
+            partner_to_channelids.remove(channel_identifier)
+        else:
+            ids_to_channels[channel_state.our_state.address][channel_identifier] = result.new_state
+
+        events.extend(result.events)
+    return TransitionResult(token_network_state, events)
+
+
+#  TODO: We should find another approach for ths function. It's "a bit" confusing
+def get_channels_to_dispatch_statechange(
+    participant_address: AddressHex,
+    state_change: StateChangeWithChannelID,
+    token_network_state: TokenNetworkState
+) -> List[NettingChannelState]:
+    channel_states = []
     ids_to_channels = token_network_state.channelidentifiers_to_channels
 
-    channel_states = []
-    if node_address is not None:
-        if node_address in ids_to_channels:
-            channel_state = ids_to_channels[node_address].get(state_change.channel_identifier)
-            channel_states.append(channel_state)
-            if channel_state.both_participants_are_light_clients\
-                and type(state_change) in DuplicableStateChanges:
-                partner_channel_state = token_network_state.channelidentifiers_to_channels[channel_state.partner_state.address].get(
-                    channel_state.identifier
-                )
-                channel_states.append(partner_channel_state)
-        else:
-            lc_address = views.get_lc_address_by_channel_id_and_partner(token_network_state, node_address,
+    if participant_address in ids_to_channels:
+        channel_state = ids_to_channels[participant_address].get(state_change.channel_identifier)
+        channel_states.append(channel_state)
+        if channel_state.both_participants_are_light_clients:
+            partner_channel = ids_to_channels[channel_state.partner_state.address]
+            partner_channel_state = partner_channel.get(channel_state.identifier)
+            channel_states.append(partner_channel_state)
+    else:
+        lc_address = views.get_lc_address_by_channel_id_and_partner(token_network_state, participant_address,
                                                                     state_change.canonical_identifier)
-            node_address = lc_address
-            if lc_address in token_network_state.channelidentifiers_to_channels:
-                channel_state = token_network_state.channelidentifiers_to_channels[lc_address].get(
-                    state_change.canonical_identifier.channel_identifier)
-                channel_states.append(channel_state)
+        if lc_address in ids_to_channels:
+            lc_channel = ids_to_channels[lc_address]
+            lc_channel_state = lc_channel.get(state_change.canonical_identifier.channel_identifier)
+            channel_states.append(lc_channel_state)
 
-        for channel_state in channel_states:
-            result = channel.state_transition(
-                channel_state=channel_state,
-                state_change=state_change,
-                block_number=block_number,
-                block_hash=block_hash,
-            )
-
-            partner_to_channelids = token_network_state.partneraddresses_to_channelidentifiers[
-                channel_state.partner_state.address
-            ]
-
-            channel_identifier = state_change.channel_identifier
-            if result.new_state is None:
-                del ids_to_channels[channel_state.our_state.address][channel_identifier]
-                partner_to_channelids.remove(channel_identifier)
-            else:
-                ids_to_channels[channel_state.our_state.address][channel_identifier] = result.new_state
-
-            events.extend(result.events)
-
-    return TransitionResult(token_network_state, events)
+    return channel_states
 
 
 def handle_channel_close(
@@ -94,12 +104,12 @@ def handle_channel_close(
     block_number: BlockNumber,
     block_hash: BlockHash,
 ) -> TransitionResult:
-    return subdispatch_to_channel_by_id_and_address(
+    return subdispatch_to_channels_by_participant_address(
         token_network_state=token_network_state,
         state_change=state_change,
         block_number=block_number,
         block_hash=block_hash,
-        node_address=state_change.participant1
+        participant_address=state_change.participant1
     )
 
 
@@ -151,12 +161,12 @@ def handle_balance(
     block_hash: BlockHash,
     participant: AddressHex
 ) -> TransitionResult:
-    return subdispatch_to_channel_by_id_and_address(
+    return subdispatch_to_channels_by_participant_address(
         token_network_state=token_network_state,
         state_change=state_change,
         block_number=block_number,
         block_hash=block_hash,
-        node_address=participant
+        participant_address=participant
     )
 
 
@@ -182,12 +192,12 @@ def handle_closed(
             ]
             node_address = participant1
 
-    return subdispatch_to_channel_by_id_and_address(
+    return subdispatch_to_channels_by_participant_address(
         token_network_state=token_network_state,
         state_change=state_change,
         block_number=block_number,
         block_hash=block_hash,
-        node_address=node_address
+        participant_address=node_address
     )
 
 
@@ -197,12 +207,12 @@ def handle_settled(
     block_number: BlockNumber,
     block_hash: BlockHash,
 ) -> TransitionResult:
-    return subdispatch_to_channel_by_id_and_address(
+    return subdispatch_to_channels_by_participant_address(
         token_network_state=token_network_state,
         state_change=state_change,
         block_number=block_number,
         block_hash=block_hash,
-        node_address=state_change.participant1
+        participant_address=state_change.participant1
     )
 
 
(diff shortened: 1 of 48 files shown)


raiden/transfer/token_network.py
```diff
@@ -41,18 +44,27 @@ def subdispatch_to_channel_by_id_and_address(
...
     ids_to_channels = token_network_state.channelidentifiers_to_channels
 
-    channel_state = None
+    channel_states = []
     if node_address is not None:
         if node_address in ids_to_channels:
             channel_state = ids_to_channels[node_address].get(state_change.channel_identifier)
+            channel_states.append(channel_state)
+            if channel_state.both_participants_are_light_clients\
+                and type(state_change) in DuplicableStateChanges:
```
mauricioirace: This could be encapsulaed

raiden/transfer/token_network.py
```diff
@@ -41,18 +44,27 @@ def subdispatch_to_channel_by_id_and_address(
...
+                channel_states.append(partner_channel_state)
         else:
             lc_address = views.get_lc_address_by_channel_id_and_partner(token_network_state, node_address,
                                                                     state_change.canonical_identifier)
             node_address = lc_address
             if lc_address in token_network_state.channelidentifiers_to_channels:
                 channel_state = token_network_state.channelidentifiers_to_channels[lc_address].get(
                     state_change.canonical_identifier.channel_identifier)
-        if channel_state:
+                channel_states.append(channel_state)
```
mauricioirace: This could be also extracted into a method

raiden/transfer/token_network.py
```diff
@@ -41,18 +44,27 @@ def subdispatch_to_channel_by_id_and_address(
...
+            if channel_state.both_participants_are_light_clients\
+                and type(state_change) in DuplicableStateChanges:
+                partner_channel_state = token_network_state.channelidentifiers_to_channels[channel_state.partner_state.address].get(
+                    channel_state.identifier
+                )
+                channel_states.append(partner_channel_state)
         else:
             lc_address = views.get_lc_address_by_channel_id_and_partner(token_network_state, node_address,
                                                                     state_change.canonical_identifier)
             node_address = lc_address
```
mauricioirace: I think we can delete this line now

raiden/transfer/token_network.py
```diff
@@ -29,7 +29,10 @@
     ContractReceiveUpdateTransfer,
 ]
 
-
+# State changes that should be duplicated in case there are two light clients
+DuplicableStateChanges = [
```
mauricioirace: This is sort of a constant, not a class

mortelli: do we expect this list to grow?

raiden/transfer/token_network.py
mortelli: with the code as it is now, yeah, it's unused

raiden/transfer/token_network.py
```diff
@@ -41,18 +44,27 @@ def subdispatch_to_channel_by_id_and_address(
...
+    channel_states = []
     if node_address is not None:
         if node_address in ids_to_channels:
             channel_state = ids_to_channels[node_address].get(state_change.channel_identifier)
+            channel_states.append(channel_state)
+            if channel_state.both_participants_are_light_clients\
+                and type(state_change) in DuplicableStateChanges:
+                partner_channel_state = token_network_state.channelidentifiers_to_channels[channel_state.partner_state.address].get(
+                    channel_state.identifier
+                )
```
mortelli: this indentation is too confusing and the line too long, may i suggest:

```suggestion
                partner_channel = token_network_state.channelidentifiers_to_channels[channel_state.partner_state.address]
                partner_channel_state = partner_channel.get(channel_state.identifier)
```

mortelli (commented):
i have read issue RLP-836 and taken a look at this PR, but still don't understand why this is the solution.

in particular, this statement is unclear to me:
> State changes that should be duplicated in case there are two light clients

as well as the `if` statement which uses this value.

could you please explain why this is the fix?

an high-level overview of the solution would be a huge help :) 

thanks!

raiden/transfer/token_network.py
marcosmartinez7: I think we can rename this to represent that "we are dispatching to multiple channels because the channel is related with two light clients of the same hub"

raiden/transfer/token_network.py
```diff
@@ -29,7 +29,10 @@
     ContractReceiveUpdateTransfer,
 ]
 
-
+# State changes that should be duplicated in case there are two light clients
+DuplicableStateChanges = [
+    ContractReceiveChannelNewBalance
+]
 def subdispatch_to_channel_by_id_and_address(
```
marcosmartinez7: i think we can invoke another funcion `subdispatch_to_channel_list `or something, to subdispatch a state change to multiple channels

raiden/transfer/token_network.py
```diff
@@ -30,50 +30,78 @@
...
     token_network_state: TokenNetworkState,
     state_change: StateChangeWithChannelID,
     block_number: BlockNumber,
     block_hash: BlockHash,
-    node_address: AddressHex = None
+    participant_address: AddressHex = None
 ) -> TransitionResult:
-    events = list()
+    if participant_address is None:
+        return TransitionResult(token_network_state, [])
```
mortelli:
```suggestion
    if not participant_address:
        return TransitionResult(token_network_state, [])
```

raiden/transfer/token_network.py
```diff
@@ -30,50 +30,78 @@
...
+    block_hash: BlockHash,
+    block_number: BlockNumber,
+    channel_states: List[NettingChannelState],
+    state_change: StateChangeWithChannelID,
+    token_network_state: TokenNetworkState
+) -> TransitionResult:
+    events = []
     ids_to_channels = token_network_state.channelidentifiers_to_channels
+    for channel_state in channel_states:
+        result = channel.state_transition(
```
mortelli: i would name this something more explicit like `channel_result`

raiden/transfer/token_network.py
```diff
@@ -30,50 +30,78 @@
...
-                partner_to_channelids.remove(channel_identifier)
-            else:
-                ids_to_channels[node_address][channel_identifier] = result.new_state
+        events.extend(result.events)
+    return TransitionResult(token_network_state, events)
 
-            events.extend(result.events)
 
-    return TransitionResult(token_network_state, events)
+#  TODO: We should find another approach for ths function. It's "a bit" confusing
```
mortelli: either we deal with this or we remove it.

if necessary, please create a ticket. 

there is not much added value for this comment, i think

raiden/transfer/token_network.py
```diff
@@ -30,50 +30,78 @@
...
+    ids_to_channels = token_network_state.channelidentifiers_to_channels
+
+    if participant_address in ids_to_channels:
+        channel_state = ids_to_channels[participant_address].get(state_change.channel_identifier)
+        channel_states.append(channel_state)
+        if channel_state.both_participants_are_light_clients:
+            partner_channel = ids_to_channels[channel_state.partner_state.address]
+            partner_channel_state = partner_channel.get(channel_state.identifier)
+            channel_states.append(partner_channel_state)
+    else:
```
mortelli: the code structure looks good, but i don't understand what this `if` is evaluating.

what does it mean that `participant_address in ids_to_channels == True`?

can we maybe express this boolean condition in a clearer, more business logic-oriented way?

at first i thought this was about light clients, but it seems both branches talk about LCs, so i'm not so sure now

mortelli (commented):
it's much better now!

thanks for iterating this :smiley:

raiden/transfer/token_network.py
mauricioirace: This was part of the discussion we had yesterday.

It's seems like both branches can be lc or full nodes, even the else branch can be a full node. I found the naming wrong if that's the case

I understand that:
If the participant is our (we are the full node owner of the address, or the corresponding lc is handled by us) then it goes to the first branch.

Else, it check if we have a channel with that address as a partner.

I think I know a few ways to improve this, but we don't even have an agreement on what does the 'else' does.

mortelli: > This was part of the discussion we had yesterday.
> 
> It's seems like both branches can be lc or full nodes, even the else branch can be a full node. I found the naming wrong if that's the case
> 
> I understand that:
> If the participant is our (we are the full node owner of the address, or the corresponding lc is handled by us) then it goes to the first branch.
> 
> Else, it check if we have a channel with that address as a partner.
> 
> I think I know a few ways to improve this, but we don't even have an agreement on what does the 'else' does.

if that's the case, could we name the condition something which reflects this?

e.g. `if participant_is_ours` or something like this?

mauricioirace: Agree.

mauricioirace: the other option I thought was renaming `ids_to_channels` to something like `our_channels`, so `participant_address in our_channels` would be a more explicit way of saying 'Is this participant address a side of some channel we own?'. 

this  would apply both to this condition and to the one in line 99

raiden/transfer/token_network.py
```diff
@@ -30,64 +30,90 @@
 ]
 
 
-def subdispatch_to_channel_by_id_and_address(
+def subdispatch_to_channels_by_participant_address(
```
marcosmartinez7, line 33: I like this because: 

- we were not using channel id
- participant seems better than address

but.. im a little bit concern about the usage, we should test basically all the onchain operations manually and run the test suite

raiden/transfer/token_network.py
```diff
@@ -30,64 +30,90 @@
...
 
-    return TransitionResult(token_network_state, events)
+def get_channels_to_dispatch_statechange(
+    participant_address: AddressHex,
+    state_change: StateChangeWithChannelID,
+    token_network_state: TokenNetworkState
+) -> List[NettingChannelState]:
+    channel_states = []
+    ids_to_channels = token_network_state.channelidentifiers_to_channels
+    participant_is_ours = participant_address in ids_to_channels
```
marcosmartinez7, line 98: maybe a comment that participant_is_ours represents ¨is a handled lc or is the hub itself¨ would be nice here

raiden/transfer/token_network.py
```diff
@@ -30,64 +30,90 @@
...
-                partner_to_channelids.remove(channel_identifier)
-            else:
-                ids_to_channels[node_address][channel_identifier] = result.new_state
+        events.extend(channel_result.events)
+    return TransitionResult(token_network_state, events)
 
-            events.extend(result.events)
 
-    return TransitionResult(token_network_state, events)
+def get_channels_to_dispatch_statechange(
```
marcosmartinez7, line 80: I think this function does what we need, but both the name andthe StateChangeWithChannelID doesnt. 

`StateChangeWithChannelID ` represents: "a set of state changes that must be subdispatched to two channels because the channels are from two onboarded light clients"

`get_channels_to_dispatch_statechange` logic is strongly related with two light clients of the hub, it basically return 1 or 2 channel states depending on `both_participants_are_light_clients` value. Basically, given an address and a channel id, we check if that channel is also into the state of another LC

I suggest to change StateChangeWithChannelID for MultipleChannelDispatchStateChange or something. 

About the function, the name is ok but we need docstring

marcosmartinez7 (changes requested):

raiden/transfer/token_network.py
JONAF2103: I Agree with @marcosmartinez7 here, i still don't like all this logic just to add 1 or 2 participant channel states to the list but i don't see any other way to do it, anyway adding a docstring and changing the state change name should help us to understand if we come back here later.

raiden/transfer/token_network.py
mauricioirace: I'll comment 'is a handled lc or is the node itself'  to be more accurate

raiden/transfer/token_network.py
mauricioirace: Agree, I'll run the tests at the end of the day. 
The method is only used internally in this file and seems like it hadn't break anythink that wasn't already broken (channel close and update balance proof) but would be wrong not to run the test suite

raiden/transfer/token_network.py
mauricioirace: StateChangeWithChannelID was introduced a long time ago. It represents state changes that hold a channel identifier. I thik MultipleChannelDispatchStateChange would lost the meaning of the type

marcosmartinez7 (approved):
lgtm now, thanks!

marcosmartinez7 (approved):

raiden/transfer/token_network.py
```diff
@@ -30,50 +30,88 @@
...
+def subdispatch_to_channels_by_participant_address(
     token_network_state: TokenNetworkState,
     state_change: StateChangeWithChannelID,
     block_number: BlockNumber,
     block_hash: BlockHash,
-    node_address: AddressHex = None
+    participant_address: AddressHex = None
 ) -> TransitionResult:
-    events = list()
+    if not participant_address:
```
mortelli, line 40: it seems a bit strange to me that a caller would omit the `participant_address` parameter if this method is just going to return an empty `TransitionResult`. but i guess this is out of scope for this PR

raiden/transfer/token_network.py
```diff
@@ -30,50 +30,88 @@
...
+    """
+    Retrieve which channels the state_change should be dispatched to.
+    In most situations, the state_change should be dispatched to this node's side of the channel.
+    If this node runs in hub mode, and both sides of the channel are light client handled by this hub,
+    then it's needed to dispatch the state_change to both channel_states (one per light client)
+    :param participant_address: address of the participan that fired the state_change
+    :param state_change: state_change to dispatch
+    :param token_network_state: token network where to look for channels
+    :return: Channel where  the state_change should be dispatched
+    """
```
mortelli: great :clap:

raiden/transfer/token_network.py
```diff
@@ -30,50 +30,88 @@
...
+) -> List[NettingChannelState]:
+    """
+    Retrieve which channels the state_change should be dispatched to.
+    In most situations, the state_change should be dispatched to this node's side of the channel.
+    If this node runs in hub mode, and both sides of the channel are light client handled by this hub,
+    then it's needed to dispatch the state_change to both channel_states (one per light client)
+    :param participant_address: address of the participan that fired the state_change
+    :param state_change: state_change to dispatch
+    :param token_network_state: token network where to look for channels
+    :return: Channel where  the state_change should be dispatched
```
mortelli: - missed a `t` in `participan`
- extra space in `where  the`

raiden/transfer/token_network.py
```diff
@@ -30,50 +30,88 @@
...
+    """
+    channel_states = []
+    ids_to_channels = token_network_state.channelidentifiers_to_channels
+    # is a handled lc or is the node itself'
```
mortelli: is the `'` supposed to be a `?`?

mortelli (approved):
you got the green light from me as long as the UTs and MTs turn out OK :+1:

JONAF2103 (approved):
LGTM, you should update the docs with the typo errors.

PR #138, opened 2020-10-20
RLP-832 Handle invite 2 lc's

JONAF2103:
This PR attempts to fix a bug on the matrix client sync method.

## Problem

When we have a fresh account and we start the node using that account (or in the LC case we onboard with a brand new account). The sync method used to get all the status of the network just end up getting stuck because of the amount of data that matrix returns. The specific error is in the presence process, basically we have an array with all the presence changes on all the public rooms on matrix for each user, and for each of those events we are creating threads and processing the data, the amount of events causes the main thread of the sync process to be slow and that's why we have a sync problem.

## Solution

For now the solution is to remove the presence listener logic and we decided to not update that since we are not using it. In the future if we keep using matrix we will need to get back this code and filter the events in a smarter way so we don't have such amount of data to process.

### Changes:

* Removing the presence listener
* Removing the presence listener call on sync

This affects the sync of matrix, it fixes an error on the first execution of the client when we update all the network state.

Here are the [test results](https://github.com/rsksmart/lumino/files/5423169/RLP-832.tests.log)

### DIFF
--- a/raiden/network/transport/matrix/client.py
+++ b/raiden/network/transport/matrix/client.py
@@ -390,10 +390,6 @@ class GMatrixClient(MatrixClient):
             self._post_hook_func(self.sync_token)
 
     def _handle_response(self, response, first_sync=False):
-        # Handle presence after rooms
-        for presence_update in response["presence"]["events"]:
-            for callback in self.presence_listeners.values():
-                self.call(callback, presence_update)
 
         for to_device_message in response["to_device"]["events"]:
             for listener in self.listeners:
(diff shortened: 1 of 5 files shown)


marcosmartinez7:
@JONAF2103 can you put here the jira issue created for reviewing this in a future?

marcosmartinez7 (approved):

mortelli (approved):
please make sure the test suite and the basic manual test cases are working OK after this change

JONAF2103:
Manual tests are working, the baseline gave me some differences on testing matrix, i will fix that and update this PR

mauricioirace (approved):
LGTM

JONAF2103:
Test Results are ok, i will merge this now

PR #140, opened 2020-10-21
RLP-774: generalize LC onboarding

mortelli:
This PR provides a solution for issue [RLP-774](https://jirainfuy.atlassian.net/browse/RLP-774), which is about generalizing LC onboarding calls to be transport-agnostic, with the goal of eventually adding a RIF Comms implementation.

## Overview
The main objective with these changes is to abstract (or hide away) Matrix-specific code from the light client onboarding process. This process consists of two steps in the form of API calls:
1. requesting onboarding info
2. requesting light client registration

To allow a RIF Comms onboarding process to be implemented, Matrix-specific parameters in these calls were removed and replaced with a generic dictionary. Although it is not an elegant solution, it will allow the system to keep just these 2 calls for the cases of both Matrix as well as RIF Comms.

As a result, the transport layer abstraction was extended to include these functionalities in the form of 2 new functions: `light_client_onboarding_data` and `register_light_client`. The code for these functions (in the case of the Matrix layer in particular) is the code which used to reside in the APIs.

## Details
- `environment_type` had to be added as an (internal) field to the Matrix transport layer when it is instantiated, as it will be later needed when responding to onboarding registration data requests.
- although it is rather unsightly, the Raiden API is needed at the moment of registering a new light client in the transport layer. this field is not accessible from the transport layer due to bad design/coupling, so it is passed as a parameter.
- the `new_light_client` method was removed from the transport layer abstraction, since it's no longer necessary to call it externally (the 2 previously mentioned API calls contain this functionality).
  - although they are not currently used, the `add_light_client` and `remove_light_client` functions still remain in the generic and matrix-specific transport layers.
- the `light_clients/` and `light_clients/matrix/credentials/` endpoints were merged into a single endpoint (`light_clients/`) which uses `GET` and `POST`.
  - `light_clients_messages` was renamed to `light_clients/messages` for consistency's sake.
- a type hint for the transport layer field was added to the `RaidenService` class.
- several Rest and Raiden API calls and classes were renamed in order to better reflect what they do.
- some type hints in the transport node had to be changed to strings in order to prevent cyclical imports.
- _Reformat code_ and _Optimize Imports_ have been applied to the modified files.

## Next steps
This PR will break the SDK. The code there will have to be updated to use the new endpoint names and structures.

Additionally, although the Matrix onboarding logic is currently hidden behind this specific transport layer implementation, it's clear some functions will have to change for the [RIF Comms onboarding implementation](https://jirainfuy.atlassian.net/browse/RLP-846) to work. Namely:
- `RaidenAPI.get_all_light_clients`
- `RaidenAPI.save_light_client`
- `RaidenAPI.validate_light_client`
- `SQLiteStorage.get_light_client`
- `SQLiteStorage.save_light_client`
- `SQLiteStorage.get_all_light_clients`
- `SQLiteStorage.get_light_client_by_api_key`

and there might be others as well.

These functions use Matrix-specific parameters. We can either refactor them, or rename them and add RIF Comms counterparts.

---

This PR builds on #110. Once it is merged, we can also merge #110 and close related issues [RLP-750](https://jirainfuy.atlassian.net/browse/RLP-750) and [RLP-775](https://jirainfuy.atlassian.net/browse/RLP-775).

### DIFF
--- a/raiden/api/python.py
+++ b/raiden/api/python.py
@@ -1,25 +1,28 @@
 import copy
+import hashlib
+import os
+import random
+from binascii import hexlify
+from datetime import datetime, date
 from http import HTTPStatus
 
-import gevent
-import structlog
-from gevent import Greenlet
-import random
-import string
-import hashlib
-from binascii import hexlify
-import os
 import dateutil.parser
-from datetime import datetime, date
+import gevent
+import string
+import structlog
 from dateutil.relativedelta import relativedelta
-from eth_utils import is_binary_address, to_checksum_address, to_canonical_address, to_normalized_address, encode_hex
-
 from ecies import encrypt
+from eth_utils import is_binary_address, to_checksum_address, to_canonical_address, to_normalized_address, encode_hex
+from gevent import Greenlet
 
 import raiden.blockchain.events as blockchain_events
 from raiden import waiting, routing
 from raiden.api.validations.api_error_builder import ApiErrorBuilder
 from raiden.api.validations.channel_validator import ChannelValidator
+from raiden.billing.invoices.decoder.invoice_decoder import decode_invoice
+from raiden.billing.invoices.encoder.invoice_encoder import parse_options, encode_invoice
+from raiden.billing.invoices.options_args import OptionsArgs
+from raiden.billing.invoices.util.time_util import get_utc_unix_time, get_utc_expiration_time
 from raiden.constants import (
     GENESIS_BLOCK_NUMBER,
     RED_EYES_PER_TOKEN_NETWORK_LIMIT,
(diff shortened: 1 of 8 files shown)


raiden/api/python.py
```diff
@@ -1727,25 +1711,12 @@ def get_all_light_clients(self):
...
-        client = make_client(available_servers)
-        server_url = client.api.base_url
-        server_name = urlparse(server_url).netloc
-        data_to_sign = {
-            "display_name_to_sign": "@" + to_normalized_address(address) + ":" + server_name,
-            "password_to_sign": server_name,
-            "seed_retry": "seed"}
-        return data_to_sign
-
-    def register_light_client(self,
```
mortelli, line 1743: - the `get_data_for_registration_request` logic now resides in the matrix transport layer implementation
- `register_light_client` was renamed to `save_light_client`

raiden/network/transport/matrix/layer.py
```diff
@@ -99,9 +106,81 @@ def full_node(self) -> TransportNode:
...
+        server_url = client.api.base_url
+        server_name = urlparse(server_url).netloc
+        return {
+            "display_name_to_sign": "@" + to_normalized_address(address) + ":" + server_name,
+            "password_to_sign": server_name,
+            "seed_retry": "seed",
+        }
+
+    def register_light_client(self, raiden_api: RaidenAPI, registration_data: dict) -> TransportNode:
+        config = raiden_api.raiden.config["transport"]["matrix"]
```
marcosmartinez7, line 124: This lgtm for now but i think we can modify this method in a near future this way: 

- receive `Wal `and `config `instead of `raiden_api`
- move the `save_light_client `method to light_Client_service
- remove the `start_transport_on_runtime` method and just invoke transport.start

this way we can remove the `RaidenApi `param and dependency.

marcosmartinez7 (approved):
Lgtm, just one comment for the future of the implementation

raiden/network/transport/matrix/layer.py
marcosmartinez7: https://jirainfuy.atlassian.net/browse/RLP-858

mortelli: thanks, agreed.

PR #142, opened 2020-10-22
Bug fix/rlp 834/updating with develop

JONAF2103:
This PR merges develop into bugFix/RLP-715/lc-to-lc-payments-same-hub, it's based on this open PR https://github.com/rsksmart/lumino/pull/119

```
Merge branch 'develop' of github.com:rsksmart/lumino into bugFix/RLP-834/updating-with-develop

# Conflicts:
#	raiden/api/python.py
#	raiden/lightclient/handlers/light_client_message_handler.py
#	raiden/raiden_event_handler.py
#	raiden/raiden_service.py
#	raiden/storage/sqlite.py
#	raiden/storage/utils.py
#	raiden/transfer/mediated_transfer/target.py
```

PR #143, opened 2020-10-22
Bug fix/rlp 838/fixing tests

JONAF2103:
This PR aims to fix all the broken tests for the refactor made on https://github.com/rsksmart/lumino/pull/119, it's based on this PR https://github.com/rsksmart/lumino/pull/142

* Fixing tests
* Fixing some bugs found by running the tests
* Fixing the test setup (this fixes a lot of broken tests)

This affects the tests

Tests checked with develop [baseline](https://github.com/anarancio/lumino-docs/blob/master/unit-testing/results/develop/2020-10-06/test_results.txt) and everything seems to be ok, here the [results](https://github.com/rsksmart/lumino/files/5438975/RLP-838.tests.log)

### DIFF
--- a/raiden/tests/integration/long_running/test_integration_events.py
+++ b/raiden/tests/integration/long_running/test_integration_events.py
@@ -151,8 +151,8 @@ def wait_both_channel_deposit(
         app_partner.raiden,
         registry_address,
         token_address,
-        app_partner.raiden.address,
         app_deposit.raiden.address,
+        app_partner.raiden.address,
         total_deposit,
         retry_timeout,
     )
@@ -228,10 +228,6 @@ def run_test_channel_deposit(raiden_chain, deposit, retry_timeout, token_address
         registry_address, token_address, app0.raiden.address, app1.raiden.address, deposit
     )
 
-    wait_both_channel_deposit(app0, app1, registry_address, token_address, deposit, retry_timeout)
-
-    assert_synced_channel_state(token_network_identifier, app0, deposit, [], app1, 0, [])
-
     RaidenAPI(app1.raiden).set_total_channel_deposit(
         registry_address, token_address, app1.raiden.address, app0.raiden.address, deposit
     )
@@ -239,7 +235,6 @@ def run_test_channel_deposit(raiden_chain, deposit, retry_timeout, token_address
     wait_both_channel_deposit(app1, app0, registry_address, token_address, deposit, retry_timeout)
 
     assert_synced_channel_state(token_network_identifier, app0, deposit, [], app1, deposit, [])
-
 
 @pytest.mark.parametrize("number_of_nodes", [2])
 @pytest.mark.parametrize("channels_per_node", [0])
--- a/raiden/tests/integration/network/transport/test_matrix_transport.py
+++ b/raiden/tests/integration/network/transport/test_matrix_transport.py
@@ -4,6 +4,7 @@ from unittest.mock import MagicMock
 
 import gevent
 import pytest
+from eth_utils import to_checksum_address
 from gevent import Timeout
 
 import raiden
@@ -23,7 +24,7 @@ from raiden.transfer import views
 from raiden.transfer.identifiers import QueueIdentifier
 from raiden.transfer.mediated_transfer.events import CHANNEL_IDENTIFIER_GLOBAL_QUEUE
 from raiden.transfer.state_change import ActionChannelClose, ActionUpdateTransportAuthData
-from raiden.utils import pex
+from raiden.utils import privatekey_to_address
 from raiden.utils.signer import LocalSigner
 from raiden.utils.typing import Address, List, Optional, Union
 
@@ -40,7 +41,7 @@ class MessageHandler:
     def __init__(self, bag: set):
         self.bag = bag
 
-    def on_message(self, _, message):
+    def on_message(self, _, message, __, ___):
         self.bag.add(message)
 
 
@@ -52,6 +53,7 @@ def mock_matrix(
     local_matrix_servers,
     private_rooms,
     global_rooms,
+    private_keys
 ):
 
     from raiden.network.transport.matrix.client import User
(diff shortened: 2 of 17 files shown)


raiden/tests/integration/long_running/test_integration_events.py
```diff
@@ -228,10 +228,6 @@ def run_test_channel_deposit(raiden_chain, deposit, retry_timeout, token_address
         registry_address, token_address, app0.raiden.address, app1.raiden.address, deposit
     )
 
-    wait_both_channel_deposit(app0, app1, registry_address, token_address, deposit, retry_timeout)
```
marcosmartinez7, line 231: Could you explain why you removed this?

raiden/tests/integration/network/transport/test_matrix_transport.py
```diff
@@ -85,7 +87,9 @@ def mock_receive_message(klass, message):  # pylint: disable=unused-argument
         private_rooms=private_rooms,
     )
 
-    transport = MatrixTransport(config)
+    address = to_checksum_address(privatekey_to_address(private_keys[0]))
+
+    transport = MatrixTransport(address=address, config=config)
```
marcosmartinez7, line 92: nice catch here :)

raiden/tests/integration/network/transport/test_matrix_transport.py
```diff
@@ -151,7 +155,7 @@ def ping_pong_message_success(transport0, transport1):
     return all_messages_received
 
 
-def is_reachable(transport: MatrixTransport, address: Address) -> bool:
+def is_reaceqhable(transport: MatrixTransport, address: Address) -> bool:
```
marcosmartinez7: typo here, be careful cause this function is used on: 

- test_matrix_invite_private_room_unhappy_case2
- test_matrix_invite_private_room_unhappy_case3
- test_matrix_user-roaming

raiden/tests/integration/network/transport/test_matrix_transport.py
```diff
@@ -944,8 +957,6 @@ def test_matrix_user_roaming(matrix_transports):
 
     transport0.start(raiden_service0, message_handler0, "")
 
-    assert is_reachable(transport1, raiden_service0.address)
```
marcosmartinez7, line 947: why this was removed?

raiden/transfer/views.py
```diff
@@ -517,7 +517,10 @@ def get_transfer_role(chain_state: ChainState, node_address: AddressHex, secreth
     in a transfer. If a transfer task is not found for the secrethash then the
     function returns None
     """
-    task = chain_state.payment_mapping[node_address].secrethashes_to_task.get(secrethash)
+    task = None
```
marcosmartinez7, line 520: Some comments here:

1. why this was needed? 
2. its ok to protect ourselves of an exception but this scenario cant happen, if happens, we have a major issue, so i prefeer to keep the code as it was
3. i think that if a PR is for solving tests and we modify something of the core of lumino, that must be another pr (with, probably, another round of manual and unit tests)

marcosmartinez7 (changes requested):
Great job in general

Some minor changes with naming :)

raiden/transfer/views.py
JONAF2103: This was because you have chain_state.payment_mapping as a dict empty sometimes and accessing a dict without that key blows the code, the background of why this dict is empty some times i don't have it, but seems to me that when you delete the task you use del and del deletes the key also so you loose the access to the object with the secret hashes too. The tests were failing before this so that's why i added that.

raiden/tests/integration/network/transport/test_matrix_transport.py
JONAF2103: Because of the change that we made on the user presence, now is reachable it returns always the same because of that so is not a valid call.

raiden/tests/integration/network/transport/test_matrix_transport.py
JONAF2103: Yes good catch, i don't know how the complete set of tests pass with this :S

JONAF2103: Make sense, i removed all the usages so it's ok. I fixed this typo

raiden/tests/integration/long_running/test_integration_events.py
JONAF2103: Actually i just moved it to the end, since we were making the deposit for the first app and the waiting but the waiting method waits for both deposits so it was failing for timeout, so now we make both deposits and then we wait, so now it make sense, before this it was working on develop since the key was only the channel id and not the address, but it was wrong and now we can see that, so to fix it i just moved that to the end to wait for both apps to make the deposits

raiden/transfer/views.py
marcosmartinez7: hmm but that situation of having the empty dict and accesing it, is related with some test? 

and if the answer is yes, why wasnt failing previously?

JONAF2103: Because before that you didn't access the dict with the node address, is like a nesting problem, before this you had something like this:
```python
chain_state.secrethashes_to_task.get(secrethash)
```
where secrethash could be or not in the dict and that was ok, since it never will fail, because if the hash wasn't there the if will check it, but now you have:
```python
chain_state.payment_mapping[node_address].secrethashes_to_task.get(secrethash)
```
where chain_state.payment_mapping can be = {} so the first "get" chain_state.payment_mapping[node_address] will be None and None can't be accessed directly so the rest of the call .secrethashes_to_task.get(secrethash) will fail for those cases.

JONAF2103: Maybe a better solution to this problem could be to make sure that when you create the dict chain_state.payment_mapping you add the addresses to the dict there, but i think that you don't have the addresses there so you need to make this check here.

marcosmartinez7 (approved):

PR #144, opened 2020-10-22
Bugfix/rlp 854/invalid close signature

mauricioirace:
TODO: Add notes and changes

### DIFF
--- a/raiden/api/python.py
+++ b/raiden/api/python.py
@@ -624,10 +624,9 @@ class RaidenAPI:
         )
 
         channel_proxy = self.raiden.chain.payment_channel(
+            creator_address,
             canonical_identifier=channel_state.canonical_identifier
         )
-
-        channel_proxy.swap_participants(creator_address)
 
         channel_proxy.set_total_deposit_light(
             total_deposit=total_deposit,
@@ -691,6 +690,7 @@ class RaidenAPI:
         )
 
         channel_proxy = self.raiden.chain.payment_channel(
+            creator_address,
             canonical_identifier=channel_state.canonical_identifier
         )
 
--- a/raiden/api/python.py
+++ b/raiden/api/python.py
@@ -632,6 +632,7 @@ class RaidenAPI:
         channel_state = channel_list[0]
 
         channel_proxy = self.raiden.chain.payment_channel(
+            participant1=creator_address,
             canonical_identifier=channel_state.canonical_identifier
         )
 
@@ -1874,7 +1875,7 @@ class RaidenAPI:
                 identifier=locked_transfer.message_identifier,
                 message=locked_transfer,
                 signed=False,
-                light_client_address=payment.light_client_address,
+                light_client_address=creator_address,
                 order=order,
                 message_type=LightClientProtocolMessageType.PaymentSuccessful,
                 wal=self.raiden.wal,
--- a/raiden/api/python.py
+++ b/raiden/api/python.py
@@ -42,7 +42,8 @@ from raiden.exceptions import (
     RaidenRecoverableError,
     UnknownTokenAddress,
     InvoiceCoding,
-    UnhandledLightClient)
+    UnhandledLightClient
+)
 from raiden.lightclient.handlers.light_client_message_handler import LightClientMessageHandler
 from raiden.lightclient.handlers.light_client_service import LightClientService
 from raiden.lightclient.handlers.light_client_utils import LightClientUtils
@@ -50,7 +51,6 @@ from raiden.lightclient.lightclientmessages.hub_response_message import HubRespo
 from raiden.lightclient.lightclientmessages.payment_hub_message import PaymentHubMessage
 from raiden.lightclient.models.light_client_payment import LightClientPayment, LightClientPaymentStatus
 from raiden.lightclient.models.light_client_protocol_message import LightClientProtocolMessageType
-
 from raiden.messages import RequestMonitoring, LockedTransfer, RevealSecret, Unlock, Delivered, SecretRequest, \
     Processed, LockExpired
 from raiden.settings import DEFAULT_RETRY_TIMEOUT, DEVELOPMENT_CONTRACT_VERSION
@@ -200,7 +200,7 @@ def transfer_tasks_view(
     view = list()
 
     for node_address in payment_mapping.keys():
-        for secrethash, transfer_task in payment_mapping[node_address].items():
+        for secrethash, transfer_task in payment_mapping[node_address].secrethashes_to_task.items():
             transfer, role = get_transfer_from_task(secrethash, transfer_task)
             if transfer is None:
                 continue
@@ -275,6 +275,9 @@ class RaidenAPI:
         time_elapsed = diff_minutes - 30
         if time_elapsed > 30:
             raise TokenAppExpired("Token app expired")
+
+    def register_secret_light(self, signed_tx: typing.SignedTransaction):
+        self.raiden.default_secret_registry.proxy.broadcast_signed_transaction_and_wait(signed_tx)
 
     def token_network_register(
         self,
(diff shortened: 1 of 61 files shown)


raiden/api/webui/static/endpointConfig.js
```diff
@@ -1,5 +1,5 @@
-const backendUrl='http://localhost:5012'; 
-const nodeAddress = '0x55825d1DA0576780926568e208f7D3F55B89Cf1a'; 
+const backendUrl='http://localhost:5002';
```
mauricioirace: Again...

raiden/blockchain_events_handler.py
```diff
@@ -171,6 +133,77 @@ def handle_channel_new(raiden: "RaidenService", event: Event):
...
+                                   token_network_identifier,
+                                   participant1: ChecksumAddress,
+                                   participant2: ChecksumAddress,
+                                   raiden):
+    is_participant1_handled_lc = LightClientService.is_handled_lc(to_checksum_address(encode_hex(participant1)),
+                                                                  raiden.wal)
+    is_participant2_handled_lc = LightClientService.is_handled_lc(to_checksum_address(encode_hex(participant2)),
+                                                                  raiden.wal)
+    is_light_channel = is_participant1_handled_lc or is_participant2_handled_lc
+    both_participants_are_light_clients = is_participant1_handled_lc and is_participant2_handled_lc
```
mauricioirace: Here we are checking for second time if particpant is lc. This could be improved

raiden/api/webui/static/endpointConfig.js
marcosmartinez7: not again decia

marcosmartinez7:
It would be nice a quick call to present this PR :)

raiden/transfer/token_network.py
```diff
@@ -327,14 +319,6 @@ def state_transition(
         iteration = handle_channel_close(
             token_network_state, state_change, block_number, block_hash
         )
-    elif type(state_change) == ActionChannelSetFee:
```
JONAF2103, line 332: Why is this removed, we remove the Channel set fee action? maybe i miss understood the point of the change

raiden/api/python.py
```diff
@@ -624,11 +624,10 @@ def set_total_channel_deposit_light(
         )
 
         channel_proxy = self.raiden.chain.payment_channel(
+            creator_address,
             canonical_identifier=channel_state.canonical_identifier
         )
 
-        channel_proxy.swap_participants(creator_address)
```
marcosmartinez7, line 705: why this was removed?

raiden/blockchain_events_handler.py
```diff
@@ -171,6 +133,77 @@ def handle_channel_new(raiden: "RaidenService", event: Event):
...
+    participant2,
+    token_network_identifier,
+    transaction_hash,
+    raiden):
+    channel_state = create_channel(block_hash, block_number, channel_identifier, participant1, participant2, raiden,
+                                   token_network_identifier, transaction_hash)
+
+    partner_address = channel_state.partner_state.address
+
+    if ConnectionManager.BOOTSTRAP_ADDR != partner_address:
```
marcosmartinez7: this should be at create_channel

raiden/api/webui/static/endpointConfig.js
mortelli: please undo the changes to this file

raiden/network/blockchain_service.py
```diff
@@ -212,7 +213,7 @@ def service_registry(self, address: Address) -> ServiceRegistry:
 
         return self.address_to_service_registry[address]
 
-    def payment_channel(self, canonical_identifier: CanonicalIdentifier) -> PaymentChannel:
+    def payment_channel(self, participant1: Address, canonical_identifier: CanonicalIdentifier) -> PaymentChannel:
```
mortelli: i don't think it makes much sense to call this `participant1` if it's the only participant parameter in this signature.

do we have any alternatives to this name? `creator_address`?

raiden/transfer/channel.py
```diff
@@ -1561,11 +1562,12 @@ def handle_action_close(
...
-    events = events_for_close(
-        channel_state=channel_state, block_number=block_number, block_hash=block_hash,
-        signed_close_tx=close.signed_close_tx
-    )
+    events = []
+    if close.participant1 == channel_state.our_state.address:
+        events = events_for_close(
+            channel_state=channel_state, block_number=block_number, block_hash=block_hash,
+            signed_close_tx=close.signed_close_tx
+        )
```
mortelli, line 1571: can you please explain the reason behind this change?

why do we not care about `close.participant2`, for example?

raiden/network/blockchain_service.py
```diff
@@ -224,17 +225,20 @@ def payment_channel(self, canonical_identifier: CanonicalIdentifier) -> PaymentC
 
         with self._payment_channel_creation_lock:
             dict_key = (token_network_address, channel_id)
```
mortelli: let's please rename this to something meaningful, e.g. `channel_identifier`

raiden/blockchain_events_handler.py
```diff
@@ -87,67 +88,28 @@ def handle_channel_new(raiden: "RaidenService", event: Event):
...
-            light_client_address = participant2
-
-        if ConnectionManager.BOOTSTRAP_ADDR != partner_address:
-            raiden.start_health_check_for(partner_address, light_client_address)
-
+            create_light_channel(block_hash, block_number, channel_identifier,
+                                 participant1, participant2, token_network_identifier,
+                                 transaction_hash, raiden)
+        if is_participant2_handled_lc:
+            create_light_channel(block_hash, block_number, channel_identifier,
```
mortelli: be careful, this can create 2 channels. is this intended?

raiden/blockchain_events_handler.py
```diff
@@ -87,67 +88,28 @@ def handle_channel_new(raiden: "RaidenService", event: Event):
...
-            channel_state=channel_state,
-            block_number=block_number,
-            block_hash=block_hash,
-        )
-        raiden.handle_and_track_state_change(new_channel)
-
-        partner_address = channel_state.partner_state.address
-
-        light_client_address = None
+    if is_light_channel:
```
mortelli, line 107: this variable serves no purpose. just do:

```python
if is_p1_lc:
  create_lc(...)
if is_p2_lc:  # should this be elif instead?
  create_lc(...)
elif:
  ...
```

raiden/blockchain_events_handler.py
```diff
@@ -87,67 +88,28 @@ def handle_channel_new(raiden: "RaidenService", event: Event):
...
+        if is_participant2_handled_lc:
+            create_light_channel(block_hash, block_number, channel_identifier,
+                                 participant2, participant1, token_network_identifier,
+                                 transaction_hash, raiden)
+    elif raiden.address == participant1:
+        create_channel(block_hash, block_number, channel_identifier, participant1, participant2, raiden,
+                       token_network_identifier, transaction_hash)
+    elif raiden.address == participant2:
+        create_channel(block_hash, block_number, channel_identifier, participant2, participant1, raiden,
+                       token_network_identifier, transaction_hash)
```
mortelli: i would rather you defined the participant variables outside (one way or the other, depending on `raiden.address`) and then just have 1 call to `create_channel`. 

for me, it's much easier to read and maintain this way

mortelli: (same with the `create_light_channel` calls)

raiden/blockchain_events_handler.py
mortelli: (i can suggest another code structure for this method depending on the answer to this question)

raiden/blockchain_events_handler.py
mortelli: if this is true, then i don't know if there is much point in having this function instead of calling `create_channel` directly

raiden/blockchain_events_handler.py
mortelli: you can use this `and` as an in-line assignment on line `202` instead

mortelli: the same applies to the variable set on `182`

raiden/blockchain_events_handler.py
```diff
@@ -171,6 +133,77 @@ def handle_channel_new(raiden: "RaidenService", event: Event):
...
+        token_address=typing.TokenAddress(token_address),
+        payment_network_identifier=raiden.default_registry.address,
+        token_network_address=token_network_identifier,
+        reveal_timeout=raiden.config["reveal_timeout"],
+        payment_channel_proxy=channel_proxy,
+        opened_block_number=block_number,
+        is_light_channel=is_light_channel,
+        both_participants_are_light_clients=both_participants_are_light_clients
+    )
+    return channel_state
```
mortelli: you can just call `return` on line `194`

mortelli (commented):
heads-up: there's a conflict in `raiden/transfer/token_network.py`

raiden/blockchain_events_handler.py
mauricioirace: It's ok as an if, as both could be light clients.

I don't think the code you suggest is actually equivalent to the original one.

raiden/blockchain_events_handler.py
mauricioirace: yes, It's intended: it's a workaround taken from the base branch. We duplicate the channels in case both participants are lc, but with the participants exchanged

raiden/blockchain_events_handler.py
mauricioirace: What about something like:

```suggestion
    if is_light_channel or raiden.address in (participant1, participant2):
        if raiden.address == participant1 or is_participant1_handled_lc:
            create_channel(
                block_hash,
                block_number,
                channel_identifier,
                participant1,
                participant2,
                token_network_identifier,
                transaction_hash,
                raiden
            )
        if raiden.address == participant2 or is_participant2_handled_lc:
            create_channel(
                block_hash,
                block_number,
                channel_identifier,
                participant2,
                participant1,
                token_network_identifier,
                transaction_hash,
                raiden)
    # Raiden node is not participant of channel. Lc are not participants
    else:
```

raiden/transfer/token_network.py
mauricioirace: yes, I mean, it never existed

raiden/transfer/channel.py
mauricioirace: Now (since banche RLP-715) we have two channels per channel (one per side), but we only want to close it once, or else, it will fail. Also, the `close.signed_close_tx` will not verify for both sides of the channel.

PR #145, opened 2020-10-26
RLP-848 Refactoring light client message protocol

JONAF2103:
This PR aims to fix the problem with the LC messages when the payments are between LC's of the same Hub. It's based on this PR https://github.com/rsksmart/lumino/pull/119

* Removing receiver_light_client_address column
* Changing sender_light_client_address column to light_client_address
* Refactoring occurrences
* Refactoring exists methods to make better checks

This affects the payment flow for all cases

Tests compared to this [summary](https://github.com/anarancio/lumino-docs/blob/master/unit-testing/results/develop/2020-10-26/summary.txt) are ok, here the [results](https://github.com/rsksmart/lumino/files/5494634/RLP-848.test.log)

### DIFF
--- a/raiden/lightclient/handlers/light_client_message_handler.py
+++ b/raiden/lightclient/handlers/light_client_message_handler.py
@@ -127,7 +127,7 @@ class LightClientMessageHandler:
             message_protocol_type=message_protocol_type,
             light_client_address=light_client_address
         )
-        cls.map_message_from_result(result)
+        return cls.map_message_from_result(result)
 
     @classmethod
     def get_message_already_stored_by_content(cls,
@@ -149,6 +149,7 @@ class LightClientMessageHandler:
                                                                    payment_id: int,
                                                                    order: int,
                                                                    wal: WriteAheadLog):
+        # TODO: Agregar chequeo por lc address o en su defecto por la address que no sea la del sender
         return wal.storage.is_light_client_protocol_message_already_stored_with_message_id(
             message_id=message_id,
             payment_id=payment_id,
@@ -156,10 +157,14 @@ class LightClientMessageHandler:
         )
 
     @classmethod
-    def get_light_client_protocol_message_by_identifier(cls,
-                                                        message_identifier: int,
-                                                        wal: WriteAheadLog):
-        result = wal.storage.get_light_client_protocol_message_by_identifier(message_identifier)
+    def get_message_by_identifier_from_sender(cls,
+                                              message_identifier: int,
+                                              sender_light_client_address: AddressHex,
+                                              wal: WriteAheadLog):
+        result = wal.storage.get_message_by_identifier_from_sender(
+            message_identifier,
+            sender_light_client_address
+        )
         return cls.map_message_from_result(result)
 
     @classmethod
@@ -232,8 +237,11 @@ class LightClientMessageHandler:
         # If exists for that payment, the same message by the order, then discard it.
         message_identifier = message.message_identifier
         # get first principal message by message identifier
-        protocol_message = LightClientMessageHandler.get_light_client_protocol_message_by_identifier(
-            message_identifier, wal)
+        protocol_message = LightClientMessageHandler.get_message_by_identifier_from_sender(
+            message_identifier=message_identifier,
+            sender_light_client_address=message.sender,
+            wal=wal
+        )
         json_message = None
         if protocol_message.signed_message is None:
             json_message = protocol_message.unsigned_message
@@ -294,12 +302,17 @@ class LightClientMessageHandler:
                 return to_checksum_address(delivered_sender) != payment_initiator
 
     @classmethod
-    def store_lc_delivered(cls, message: Delivered, wal: WriteAheadLog):
+    def store_lc_delivered(cls,
+                           message: Delivered,
+                           wal: WriteAheadLog):
         # If exists for that payment, the same message by the order, then discard it.
         message_identifier = message.delivered_message_identifier
         # get first by message identifier
-        protocol_message = LightClientMessageHandler.get_light_client_protocol_message_by_identifier(
-            message_identifier, wal)
+        protocol_message = LightClientMessageHandler.get_message_by_identifier_from_sender(
+            message_identifier=message_identifier,
+            sender_light_client_address=message.sender,
+            wal=wal
+        )
         if protocol_message.signed_message is None:
             json_message = protocol_message.unsigned_message
         else:
--- a/raiden/lightclient/handlers/light_client_message_handler.py
+++ b/raiden/lightclient/handlers/light_client_message_handler.py
@@ -31,15 +31,15 @@ def build_light_client_protocol_message(identifier: int,
         signed_msg = None
         unsigned_msg = message
     return LightClientProtocolMessage(
-        signed,
-        order,
-        payment_id,
-        identifier,
-        message_type,
-        unsigned_msg,
-        signed_msg,
-        None,
-        light_client_address
+        is_signed=signed,
+        message_order=order,
+        light_client_payment_id=payment_id,
+        identifier=identifier,
+        message_type=message_type,
+        unsigned_message=unsigned_msg,
+        signed_message=signed_msg,
+        internal_msg_identifier=None,
+        light_client_address=light_client_address
     )
 
 
@@ -113,6 +113,7 @@ class LightClientMessageHandler:
     @classmethod
     def get_message_already_stored_for_payment(cls,
                                                message_id: int,
+                                               light_client_address: AddressHex,
                                                payment_id: int,
                                                order: int,
                                                message_type: LightClientProtocolMessageType,
@@ -123,7 +124,8 @@ class LightClientMessageHandler:
             payment_id=payment_id,
             order=order,
             message_type=str(message_type.value),
-            message_protocol_type=message_protocol_type
+            message_protocol_type=message_protocol_type,
+            light_client_address=light_client_address
         )
         cls.map_message_from_result(result)
 
--- a/raiden/lightclient/handlers/light_client_message_handler.py
+++ b/raiden/lightclient/handlers/light_client_message_handler.py
@@ -147,13 +147,13 @@ class LightClientMessageHandler:
         return None
 
     @classmethod
-    def get_message_for_payment_and_order(cls,
+    def get_message_for_order_and_address(cls,
                                           message_id: int,
                                           payment_id: int,
                                           order: int,
                                           light_client_address: AddressHex,
                                           wal: WriteAheadLog):
-        return cls.map_message_from_result(wal.storage.get_message_for_payment_and_order(
+        return cls.map_message_from_result(wal.storage.get_message_for_order_and_address(
             message_id=message_id,
             payment_id=payment_id,
             order=order,
@@ -268,7 +268,7 @@ class LightClientMessageHandler:
             cls.log.error("Unable to find principal message for {} {}: ".format(message.__class__.__name__,
                                                                                 message_identifier))
         else:
-            exists = LightClientMessageHandler.get_message_for_payment_and_order(
+            exists = LightClientMessageHandler.get_message_for_order_and_address(
                 message_id=message_identifier,
                 payment_id=protocol_message.light_client_payment_id,
                 order=order,
@@ -360,7 +360,7 @@ class LightClientMessageHandler:
             cls.log.error("Unable to find principal message for {} {}: ".format(message.__class__.__name__,
                                                                                 message_identifier))
         else:
-            exists = LightClientMessageHandler.get_message_for_payment_and_order(
+            exists = LightClientMessageHandler.get_message_for_order_and_address(
                 message_id=message_identifier,
                 payment_id=protocol_message.light_client_payment_id,
                 order=order,
--- a/raiden/lightclient/handlers/light_client_message_handler.py
+++ b/raiden/lightclient/handlers/light_client_message_handler.py
@@ -139,7 +139,7 @@ class LightClientMessageHandler:
                                wal: WriteAheadLog):
 
         if message_type and light_client_address and message:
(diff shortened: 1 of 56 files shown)


raiden/lightclient/handlers/light_client_message_handler.py
```diff
@@ -319,13 +351,23 @@ def store_lc_delivered(cls, message: Delivered, wal: WriteAheadLog):
             cls.log.error("Unable to find principal message for {} {}: ".format(message.__class__.__name__,
                                                                                 message_identifier))
         else:
-            exists = LightClientMessageHandler.is_light_client_protocol_message_already_stored_message_id(
-                message_identifier, protocol_message.light_client_payment_id, order, wal)
+            exists = LightClientMessageHandler.get_message_for_payment_and_order(
```
marcosmartinez7: maybe get_message_for_order_and_address is better? i mean, this method take into account the light client address to, because given a payment and an order, you can get 2 locked transfers for example

raiden/lightclient/handlers/light_client_service.py
```diff
@@ -64,6 +64,12 @@ def apply_message_order_filter(cls, message: LightClientProtocolMessage, msg_ord
     def get_light_client_payment(cls, payment_id, storage: SerializedSQLiteStorage):
         payment = storage.get_light_client_payment(payment_id)
         if payment:
-            payment = LightClientPayment(payment[3], payment[4], payment[5], int(payment[6]), payment[7],
-                                         payment[8], payment[0], payment[1], payment[2])
+            payment = LightClientPayment(identifier=payment[0],
```
marcosmartinez7, line 67: thanks for this reorg of parameters 🙏

raiden/transfer/node.py
```diff
@@ -90,7 +90,7 @@
     Union,
     Address, AddressHex)
 
-from eth_utils import to_canonical_address
+from eth_utils import to_canonical_address, to_checksum_address
```
mortelli: unused import in `to_checksum_address`

raiden/lightclient/models/light_client_payment.py
```diff
@@ -26,10 +24,7 @@ def __init__(
         created_on: string,
         payment_status: LightClientPaymentStatus,
         identifier: string,
-        # this is only to get easy access to the client, we only query this from DB but is not needed
-        # to create a payment
-        sender_light_client_address: AddressHex = None,
-        receiver_light_client_address: AddressHex = None
+        light_client_address: Address
```
mortelli: can we maybe rename this field or at least add a comment?

it is not initially clear whether this is the originator of the payment or the destination

raiden/lightclient/handlers/light_client_message_handler.py
```diff
@@ -84,111 +97,123 @@ def update_onchain_light_client_protocol_message_set_signed_transaction(
...
-                                              message[1],
-                                              message[4],
-                                              message[0],
-                                              message[5],
-                                              message[2],
-                                              message[3],
-                                              message[6],
-                                              message[7],
-                                              message[8])
+    def map_message_from_result(cls, result: Any) -> Any:
```
mortelli, line 180: superb

raiden/lightclient/handlers/light_client_message_handler.py
```diff
@@ -84,111 +97,123 @@ def update_onchain_light_client_protocol_message_set_signed_transaction(
...
-                                  unsigned_message: Message,
-                                  wal: WriteAheadLog):
+    def get_message_by_content(cls,
+                               light_client_address: AddressHex,
+                               message_type: LightClientProtocolMessageType,
+                               message: Message,
+                               wal: WriteAheadLog):
+
+        if message_type and light_client_address and message:
+            cls.map_message_from_result(wal.storage.get_message_by_content(
```
mortelli: missing a `return` here

raiden/lightclient/handlers/light_client_service.py
```diff
@@ -64,6 +64,12 @@ def apply_message_order_filter(cls, message: LightClientProtocolMessage, msg_ord
     def get_light_client_payment(cls, payment_id, storage: SerializedSQLiteStorage):
         payment = storage.get_light_client_payment(payment_id)
         if payment:
-            payment = LightClientPayment(payment[3], payment[4], payment[5], int(payment[6]), payment[7],
```
mortelli, line 67: could you please briefly explain why `payment[0]` wasn't previously the first field used to instantiate the `LightClientPayment` class?

(i'm guessing the table changed, but can't seem to find this change)

raiden/network/transport/matrix/transport.py
```diff
@@ -1445,6 +1445,11 @@ def _run(self):
             self.stop()  # ensure cleanup and wait on subtasks
             raise
 
+    def send_async(self, queue_identifier: QueueIdentifier, message: Message):
+        self.log.info(f"----------------->>> Sending Message from LC with address {self.get_address()}")
+        self.log.info(f"----------------->>> Message Content {str(message)}")
+        super().send_async(queue_identifier=queue_identifier, message=message)
+
```
mortelli: could you please comment on this change?

raiden/raiden_event_handler.py
```diff
@@ -122,10 +123,10 @@ def unlock_light(raiden: "RaidenService",
         sender=partner,
         merkle_tree_leaves=leaves_packed
     )
-    if not LightClientMessageHandler.is_message_already_stored(
+    if not LightClientMessageHandler.get_message_by_content(
         light_client_address=channel_unlock_event.client,
         message_type=LightClientProtocolMessageType.UnlockLightRequest,
-        unsigned_message=message,
+        message=message,
```
mortelli, line 129: i would maybe explicitly name this `message_content` or `content` (just be consistent with `message_type`)

raiden/raiden_service.py
```diff
@@ -1329,45 +1329,41 @@ def initiate_send_delivered_light(self, sender_address: Address, receiver_addres
                                       message_type: LightClientProtocolMessageType):
         lc_transport = self.get_light_client_transport(to_checksum_address(sender_address))
         # check if receiver is a handled light client too
-        print("ME LLEGO UN DELIVERED CON ORDEN")
```
mortelli, line 1332: :joy:

raiden/raiden_service.py
```diff
@@ -1329,45 +1329,41 @@ def initiate_send_delivered_light(self, sender_address: Address, receiver_addres
...
-                    self.wal,
-                    payment_id
+                    identifier=delivered.delivered_message_identifier,
+                    message=delivered,
+                    signed=True,
+                    light_client_address=sender_address,
+                    order=msg_order,
+                    message_type=message_type,
+                    wal=self.wal,
+                    payment_id=payment_id
```
mortelli, line 1349: thank you for all the named parameters you've added in this PR :pray:

raiden/transfer/node.py
```diff
@@ -327,33 +331,27 @@ def handle_init_unlock_light(
...
-    channel_state = views.get_channelstate_by_canonical_identifier_and_address(
-        chain_state=chain_state,
-        canonical_identifier=CanonicalIdentifier(
-            chain_identifier=chain_state.chain_id,
-            token_network_address=state_change.signed_lock_expired.token_network_address,
-            channel_identifier=state_change.signed_lock_expired.channel_identifier,
-        ),
-        address=state_change.sender
-    )
-
```
mortelli, line 339: could you briefly explain why this isn't needed anymore?

raiden/transfer/node.py
```diff
@@ -922,6 +920,7 @@ def handle_receive_unlock_light(
     chain_state: ChainState, state_change: ReceiveUnlockLight
 ) -> TransitionResult[ChainState]:
     secrethash = state_change.secrethash
+    print("HANDLE RECEIVE UNLOCK LIGHT")
```
mortelli: this can be removed

mortelli (commented):
this looks like a great improvement :clap: 

i didn't verify the low-level methods related to SQL operations, but overall it seems like the code is in good shape.

there is some code pulled from another branch here, right? (other than #143) if so, please specify this in the PR description.

thanks for your hard work on this :heart:

raiden/lightclient/handlers/light_client_service.py
JONAF2103: No the payment table didn't change, the only thing i did here was to put the named parameters, the order then is not important since you use the name of the parameter, actually the identifier is not the first param if you don't use named parameters, is the boolean is_lc_initiator.

raiden/lightclient/models/light_client_payment.py
JONAF2103: We can't because it depends on the situation, if you receive a payment it's saved with the receiver lc address if you are sending is the initiator lc address, so i can't put a name to represent both cases, sender_or_receiver_light_address is not a good name actually so i keep it generic.

raiden/network/transport/matrix/transport.py
JONAF2103: This is only to get a better log of the message being sent out, before this we had a generic print actually that wasn't using log, and also didn't differentiate between LC or FN, we can remove this anyway i just added this to have a better view of the message interaction.

raiden/raiden_event_handler.py
JONAF2103: the thing is that the parameter has the message actually, the content is extracted inside the method

raiden/transfer/node.py
JONAF2103: Will remove it

raiden/transfer/node.py
JONAF2103: We removed the receiver_light_client_address since we are not using both columns anymore so, we were using channel_state to get that, but now we don't so we don't need to get the channel_state anymore

raiden/lightclient/handlers/light_client_message_handler.py
JONAF2103: Good catch

raiden/raiden_event_handler.py
mortelli: i don't actually see this content being extracted, just the entire message being serialized. 

could you please expand on this?

raiden/lightclient/models/light_client_payment.py
mortelli: ok, so we use the same object for sent and received payments? wouldn't the address in that case always be the sender of the payment?

mortelli: i don't think i get it yet :thinking: 

i find these two statements contradictory:
> if you receive a payment it's saved with the receiver lc address 

>if you are sending is the initiator lc address

if the address is set to the initiator when you send the payment, how can it then be set to the receiver when you receive it?

raiden/lightclient/handlers/light_client_service.py
mortelli: cool, thanks.

raiden/network/transport/matrix/transport.py
mortelli: thanks for the clarification

raiden/lightclient/models/light_client_payment.py
JONAF2103: Ok so i got confussed about this, and you are right, it's always the initiator so i will change that

raiden/raiden_event_handler.py
JONAF2103: the serialized message is not the same as the object, the serialized message is a string, if you use this method with a string instead of a message object it will fail, so that's why i didn't use the message_content as a parameter because it could be confusing for who is going to use this

mortelli: it's not a blocker for me. the issue i have here is that if the method is called `get_message_by_content`, it should be clear _what_ the content is. 

the type is one thing, and the serialized message is another, so what is actually the _content_?

raiden/lightclient/models/light_client_payment.py
mortelli: thanks!

mortelli (approved):

raiden/raiden_event_handler.py
JONAF2103: The content should be the serialized message value, the thing is you can't provide that from outside the method, only from the sqlite instance, since we have the serializer there.

raiden/lightclient/models/light_client_payment.py
mauricioirace: If I don't getting it wrong, we are now duplicating the messages: The hub store messages when they are sent, and when they are received, right?

raiden/raiden_event_handler.py
```diff
@@ -237,34 +239,49 @@ def on_raiden_event(self, raiden: "RaidenService", chain_state: ChainState, even
...
+            )
         else:
-            stored_but_unsigned = existing_message.signed_message is None
-            if stored_but_unsigned and store_message_event.is_signed:
+            existing_message = LightClientMessageHandler.get_message_by_content(
+                light_client_address=store_message_event.light_client_address,
+                message_type=store_message_event.message_type,
+                message=store_message_event.message,
+                wal=raiden.wal
+            )
```
mauricioirace: I would extract this into a function `get_message_by_event(raiden: "RaidenService", store_message_event: StoreMessageEvent) -> Optional[LigthClientProtocolMessage]` or something like that.

JONAF2103: I like your suggestion, it's done, thanks

marcosmartinez7 (changes requested):

raiden/lightclient/handlers/light_client_message_handler.py
JONAF2103: Done

marcosmartinez7 (approved):
Lgtm

mauricioirace (approved):
Lgtm

PR #147, opened 2020-10-29
bugfix/RLP-855/settlement-light

mauricioirace:
Add error handling to prevent a race condition ocurring when settling a channel.

### DIFF
--- a/raiden/api/rest.py
+++ b/raiden/api/rest.py
@@ -1733,10 +1733,14 @@ class RestAPI:
                 channel_identifier=channel_identifier,
                 signed_settle_tx=signed_settle_tx
             )
+            return self.update_channel_state(registry_address, channel_state)
+        except ChannelNotFound as e:
+            return api_error(
+                errors="Channel is already unlocked.",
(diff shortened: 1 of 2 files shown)


raiden/api/rest.py
```diff
@@ -1733,11 +1733,15 @@ def settlement_light(self,
                 channel_identifier=channel_identifier,
                 signed_settle_tx=signed_settle_tx
             )
+            return self.update_channel_state(registry_address, channel_state)
+        except ChannelNotFound as e:
+            return api_error(
+                errors="Channel is already unlocked.",
```
mortelli, line 1739: here i would say `Channel is already settled` instead, and omit the `.` to be consistent with the previous error messages.

mortelli (approved):
good catch :clap: we definitely need this. 

question: if i were to attempt to settle a channel which flat-out doesn't exist, would i get the `ChannelNotFound` exception anyway from the API call?

i was thinking that it would be nice to tell an _already_ settled channel apart from one which just wasn't created. can you confirm this is not possible?

in any case, i still prefer this to the alternative of the node crashing :+1:

PR #148, opened 2020-11-05
RLP-863: RIFCommsClient base

marcosmartinez7:
# Description 

This PR introduces a new `RIFCommsClient` class. The purpose of this class is to interact with a [RIF Communication pub sub node](https://github.com/rsksmart/rif-communications-pubsub-node/) and implement the base primitives.

# Changelog 

- added connect, create topic, send message,close topic and disconnect method
- added grpc libs to `requirements-dev`
- added fix for monkey patched gevent and grpc (https://github.com/grpc/grpc/pull/14561)
- initialized rif comms client on transport node with some hardcoded values
- added test_rif_comms_client test with basic usage


transport/rif_comms/client.py
```diff
@@ -0,0 +1,63 @@
+from eth_typing import Address
+
+from transport.message import Message
+from transport.rif_comms.proto.api_pb2 import Notification, Void
+
+
+class RifCommsClient:
+    """
+    SDK like class to connect and operate against a RIF Communications pub-sub node.
```
marcosmartinez7: dont like the ´sdk like class´

transport/rif_comms/client.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+
+    def __init__(self, node_address: Address, grpc_api_endpoint: str):
```
marcosmartinez7: does node_address name reflects the usage of this parameter?

transport/rif_comms/client.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+        Constructs the RifCommsClient
+        :param node_address: address of the node that wants to use the RIF Comms server
+        :param grpc_api_endpoint: http uri of the RIF Communications pub-sub node
+        """
+        self.node_address = node_address
+
+    def connect(self, node_address: Address) -> Notification:
```
marcosmartinez7: does node_address name reflects the usage of this parameter?

transport/rif_comms/client.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+        Connects to RIF Communications Node.
+        Invokes ConnectToCommunicationsNode grpc api endpoint.
+        :param node_address:  address of the client node that is trying to connect.
+        :return: Notification stream
+        """
+        raise NotImplementedError
+
+    def create_topic(self, partner_address: Address) -> Notification:
```
marcosmartinez7: does partner_address name reflects the usage of this parameter?

transport/rif_comms/client.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+        Creates a pub-sub topic between self.node_address and partner_address.
+        Invokes CreateTopicWithRskAddress grpc api endpoint.
+        :param partner_address:
+        :return: Notification stream
+        """
+        raise NotImplementedError
+
+    def send_message(self, receiver: Address, message: Message) -> Void:
```
marcosmartinez7: does receiver name reflects the usage of this parameter?

transport/rif_comms/client.py
```diff
@@ -0,0 +1,82 @@
...
+from grpc import insecure_channel
+
+from transport.message import Message
+from transport.rif_comms.proto.api_pb2 import Notification, Void, PublishPayload, Channel, Msg, RskAddress
+from transport.rif_comms.proto.api_pb2_grpc import CommunicationsApiStub
+
+
+class RifCommsClient:
+    """
+    Client like class to connect and operate against a RIF Communications pub-sub node.
```
mortelli: i don't think this needs to be called client-_like_, it's pretty much a client imo

transport/rif_comms/client.py
mortelli: i think maybe `rsk_address` would be clearer

transport/rif_comms/client.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+
+    def __init__(self, node_address: Address, grpc_api_endpoint: str):
+        """
+        Constructs the RifCommsClient
+        :param node_address: address of the node that wants to use the RIF Comms server
+        :param grpc_api_endpoint: http uri of the RIF Communications pub-sub node
+        """
+        self.node_address = RskAddress(address=node_address)
+        self.grpc_channel = insecure_channel(grpc_api_endpoint)
```
mortelli: i think it is OK to leave these for now, but we might not need to have `grpc_channel` set as an instance variable after all

transport/rif_comms/client.py
```diff
@@ -0,0 +1,82 @@
...
+class RifCommsClient:
+    """
+    Client like class to connect and operate against a RIF Communications pub-sub node.
+    """
+
+    def __init__(self, node_address: Address, grpc_api_endpoint: str):
+        """
+        Constructs the RifCommsClient
+        :param node_address: address of the node that wants to use the RIF Comms server
+        :param grpc_api_endpoint: http uri of the RIF Communications pub-sub node
```
mortelli: it's not HTTP, no?

mortelli: i think it's OK to use either `Communications` or `Comms`, but let's be consistent across the module

transport/rif_comms/client.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+        self.node_address = RskAddress(address=node_address)
+        self.grpc_channel = insecure_channel(grpc_api_endpoint)
+        self.stub = CommunicationsApiStub(self.grpc_channel)
+
+    def connect(self) -> Notification:
```
mortelli, line 24: keep in mind this method _might_ have to be invoked in the constructor.

we don't know if the `ConnectToCommunicationsNode` call will be needed as a pre-requisite for other calls, like creating a topic or sending a message.

another alternative is to invoke this connection logic inside these other actions that might need it.

but in the latter case, we should also look at whether calling `ConnectToCommunicationsNode` multiple times (or maybe even concurrently) could turn out to be problematic.

transport/rif_comms/client.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class RifCommsClient:
+    """
+    Client like class to connect and operate against a RIF Communications pub-sub node.
+    """
+
+    def __init__(self, node_address: Address, grpc_api_endpoint: str):
+        """
+        Constructs the RifCommsClient
```
mortelli: `RifCommsClient` → `RIF Comms Client` / `RIF Communications Client`

transport/rif_comms/client.py
```diff
@@ -0,0 +1,82 @@
...
+        :param node_address: address of the node that wants to use the RIF Comms server
```
mortelli:
```suggestion
        :param rsk_address: RSK address of the RIF Lumino node that will connect to the RIF Comms node.
```

transport/rif_comms/client.py
mortelli: issues here:
- this should probably receive a topic ID or topic name (even if it is of the `Address` type or derived from it).
- this should probably return a topic.

maybe talking about _partner_ here muddies the waters a little bit. we might create a topic for ourselves, and in that case _partner_ would not make as much sense. i would probably go with a parameter name which more clearly conveys how the topic is going to be constructed (like `topic_name`).

in any case i suggest we stand by while we figure out how the `Notification` type works.

transport/rif_comms/client.py
```diff
@@ -0,0 +1,82 @@
...
+    def send_message(self, topic_id: Address, message: Message) -> Void:
+        """
+        Sends a message to receiver node.
+        Invokes the SendMessageToTopic grpc api endpoint
+        :param topic_id: topic identifier
+        :param message: the message data
+        :return: void
+        """
+
+        # TODO message encoding
```
mortelli, line 51: indeed. the `message` should be encoded with something other than `str.encode`, and passed as the payload on line `55`.

transport/rif_comms/client.py
```diff
@@ -0,0 +1,82 @@
...
+        self.stub.SendMessageToTopic(
+            PublishPayload(
+                topic=Channel(channelId=topic_id),
+                message=Msg(payload=str.encode("Test message"))
+            )
+        )
+
+    def close_topic(self, topic_id: str) -> Void:
+        """
+        Closes the topic identified as topic_id.
```
mortelli: i would prefer we talk about _unsubscribing_ from a topic, but the difference is unfortunately not clear yet.

transport/rif_comms/client.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def disconnect(self) -> Void:
+        """
+        Disconnects from RIF Communications Node.
+        Invokes the EndCommunication grpc api endpoint.
+        :return: void
+        """
+        self.grpc_channel.unsubscribe()
+
+    def locate_peer_id(self, node_address: Address) -> str:
```
mortelli: is it worth adding a new type for peer IDs other than `str`?

transport/rif_comms/client.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+        Gets the peer id associated with a node address
+        :param node_address: the node address to locate
+        :return: a string that represents the peer id
```
mortelli: `ID` should be capitalized in the docstring

mortelli (commented):
this looks to be on the right track :+1: 

leaving aside the technical doubts that still need to be dealt with, the main issue i see is related to the return types for the functions that are being added here:
- first of all, i'm not sure we want to use the `Void` type (which seems to come from the bindings). we might want to convert this to something that Lumino can understand, and it's also better to keep these custom types as contained as possible within the client/RIF comms modules.
- even if we go with `Void`, we are missing `return` statements in almost all of the methods defined for the client class.

most of the rest of my comments are about style or grammar, so nothing major there.

transport/rif_comms/client.py
marcosmartinez7: im trying to not use "Lumino" anywhere in order to extract this file as a library in a future

transport/rif_comms/client.py
marcosmartinez7: we need it in order to close the connection right?

transport/rif_comms/client.py
marcosmartinez7: not sure what to do here now

i think its ok right now

1. consturctor creates the stub and requires to create the grpc client for doing that
2. connect initialize the rsk address on the dht
3. disconnect frees the resources of the constructor

what do u think?

transport/rif_comms/client.py
marcosmartinez7: agree, lets  keep this discussion unresolved for now

transport/rif_comms/client.py
marcosmartinez7: if we use the peer id in some point and it adds more information i will do that

transport/rif_comms/client.py
mortelli: you're right, my mistake :+1:

transport/rif_comms/client.py
mortelli: we can also go ahead and capitalize `GRPC` and `URI` :nerd_face:

transport/rif_comms/client.py
mortelli: based on what we discussed on friday, it appears the `ConnectToCommunicationsNode` method is used to register an RSK address on the DHT and link it to a peer ID.

for now, i cannot think of any call to the comms node that we would want to do without having first registered an RSK address (please correct me if you do).

therefore, i would opt to at least invoke `connect` in the class constructor, and possibly give it an internal method name.

transport/rif_comms/client.py
mortelli: fair enough :+1:

transport/rif_comms/client.py
```diff
@@ -0,0 +1,89 @@
...
+class RifCommsClient:
+    """
+    Class to connect and operate against a RIF Communications pub-sub node.
+    """
+
+    def __init__(self, rsk_address: Address, grpc_api_endpoint: str):
+        """
+        Constructs the Rif Communications Client
+        :param rsk_address: address of the node that wants to use the RIF Comms server
+        :param grpc_api_endpoint: grpc uri of the RIF Communications pub-sub node
```
mortelli:
```suggestion
        Constructs the RIF Communications Client.
        :param rsk_address: address of the node that wants to use the RIF Communications server
        :param grpc_api_endpoint: GRPC URI of the RIF Communications pub-sub node
```

mortelli (approved):
i think this is good enough for a foundation

transport/rif_comms/client.py
marcosmartinez7: i have created an issue to review this in a future

PR #151, opened 2020-11-06
Bugfix/rlp 854/invalid close signature

JONAF2103:
**DISCLAIMER**: Re-creating PR due to error merging, this PR is the same as the merged by mistake here https://github.com/rsksmart/lumino/pull/144

****************************************************************************************************************************

When two light clients are handled by the same hub, and a channel is closed, *the Hub crashes showing an invalid balance proof message.*

Indeed, it tries to validate the signature of the balance proof against the receiver's address, instead of the sender's.

This happened due to the coexistence of two LCs with different addresses, under a single **Payment Channel Proxy** (Which holds an address for "our" participant (participant1) an its partner (participant2).

This PR duplicates the **Payment Channel's Proxy** in order to prevent the hub to use the wrong address to validate, both when _closing a channel_ or when sending the _non-closing balance proof._

### DIFF
--- a/raiden/blockchain_events_handler.py
+++ b/raiden/blockchain_events_handler.py
@@ -298,14 +298,16 @@ def handle_channel_closed(raiden: "RaidenService", event: Event):
         else:
             # Must be a light client
             latest_non_closing_balance_proof = LightClientMessageHandler.get_latest_light_client_non_closing_balance_proof(
-                channel_state.identifier, raiden.wal.storage)
+                channel_state.identifier, channel_state.partner_state.address, raiden.wal.storage
+            )
             channel_closed = ContractReceiveChannelClosedLight(
                 transaction_hash=transaction_hash,
                 transaction_from=args["closing_participant"],
                 canonical_identifier=channel_state.canonical_identifier,
                 block_number=block_number,
                 block_hash=block_hash,
-                light_client_address=channel_state.our_state.address,
+                closing_participant=channel_state.our_state.address,
+                non_closing_participant=channel_state.partner_state.address,
                 latest_update_non_closing_balance_proof_data=latest_non_closing_balance_proof
             )
             raiden.handle_and_track_state_change(channel_closed)
--- a/raiden/blockchain_events_handler.py
+++ b/raiden/blockchain_events_handler.py
@@ -1,8 +1,13 @@
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Optional
 
 import gevent
 import structlog
 from eth_utils import to_checksum_address, encode_hex
+from raiden_contracts.constants import (
+    EVENT_SECRET_REVEALED,
+    EVENT_TOKEN_NETWORK_CREATED,
+    ChannelEvent,
+)
 
 from raiden.blockchain.events import Event
 from raiden.blockchain.state import get_channel_state
@@ -14,7 +19,7 @@ from raiden.network.proxies.utils import get_onchain_locksroots
 from raiden.transfer import views
 from raiden.transfer.architecture import StateChange
 from raiden.transfer.identifiers import CanonicalIdentifier
-from raiden.transfer.state import TokenNetworkState, TransactionChannelNewBalance
+from raiden.transfer.state import TokenNetworkState, TransactionChannelNewBalance, ChainState, NettingChannelState
 from raiden.transfer.state_change import (
     ContractReceiveChannelBatchUnlock,
     ContractReceiveChannelClosed,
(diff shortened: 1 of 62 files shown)


raiden/network/transport/matrix/transport.py
```diff
@@ -1447,6 +1447,11 @@ def _run(self):
             self.stop()  # ensure cleanup and wait on subtasks
             raise
 
+    def send_async(self, queue_identifier: QueueIdentifier, message: Message):
```
marcosmartinez7: @JONAF2103 Any ideas of this change?

raiden/transfer/node.py
```diff
@@ -1022,7 +1022,7 @@ def handle_state_change(
         iteration = handle_contract_receive_channel_closed(chain_state, state_change, chain_state.our_address)
     elif type(state_change) == ContractReceiveChannelClosedLight:
         assert isinstance(state_change, ContractReceiveChannelClosedLight), MYPY_ANNOTATION
-        iteration = handle_contract_receive_channel_closed(chain_state, state_change, state_change.light_client_address)
+        iteration = handle_contract_receive_channel_closed(chain_state, state_change, state_change.closing_participant)
```
mortelli: maybe we should also update the signature for `handle_contract_receive_channel_closed` so that it references the closing participant, rather than "our address"

or, 

if it is actually correct to call this "our address" once the method is invoked, then i would add something like:

```python
our_address = state_change.closing_participant
iteration = handle_contract_receive_channel_closed(chain_state, state_change, our_address)
```

raiden/transfer/events.py
```diff
@@ -189,9 +195,11 @@ def __init__(
         expiration: BlockExpiration,
         balance_proof: "BalanceProofSignedState",
         triggered_by_block_hash: BlockHash,
+        our_address: Address
     ) -> None:
         super().__init__(triggered_by_block_hash, expiration)
         self.balance_proof = balance_proof
+        self.our_address = our_address
```
mortelli, line 202: why was this field added to `ContractSendChannelUpdateTransfer` in particular?

raiden/tests/integration/network/proxies/test_payment_channel.py
```diff
@@ -89,6 +90,7 @@ def test_payment_channel_proxy_basics(
     assert new_channel_identifier is not None
 
     channel_proxy_2 = chain.payment_channel(
+        creator_address=privatekey_to_address(private_keys[1]),
```
mortelli, line 93: are you sure both proxies need to use the same private key?

raiden/network/proxies/utils.py
```diff
@@ -72,7 +72,7 @@ def get_onchain_locksroots(
     - When channel is settled A must query the blockchain to figure out which
       locksroot was used.
     """
-    payment_channel = chain.payment_channel(canonical_identifier=canonical_identifier)
+    payment_channel = chain.payment_channel(participant1, canonical_identifier)
```
mortelli, line 75: could the first argument here ever be `participant2`?

raiden/blockchain_events_handler.py
```diff
@@ -91,67 +92,65 @@ def handle_channel_new(raiden: "RaidenService", event: Event):
...
     participant2 = args["participant2"]
-    is_participant = raiden.address in (participant1, participant2)
+
 
     # Check if at least one of the implied participants is a LC handled by the node
-    is_participant1_handled_lc = LightClientService.is_handled_lc(to_checksum_address(encode_hex(participant1)),
-                                                                  raiden.wal)
-    is_participant2_handled_lc = LightClientService.is_handled_lc(to_checksum_address(encode_hex(participant2)),
-                                                                  raiden.wal)
+    is_participant1_handled_lc = LightClientService.is_handled_lc(
```
mortelli, line 98: so, we've talked about this method before (specifically, the issue of it being too lengthy).

having taken another look at it now, i can see that most of the lines are just arguments for `create_channel`. i think this would be best solved by using a class for all the parameters (there's way too many of them in my opinion) but i don't see this as a blocking issue.

so, in summary, i would leave as is for now

mortelli (approved):
LGTM :+1: left some minor questions and remarks.

- please make sure to look at the lgtm bot results: there's stuff like unused imports in some of the code
- (as always) please merge this only when you have made sure that the manual testing turns out OK

raiden/tests/integration/network/proxies/test_payment_channel.py
mauricioirace: Both proxies use the same client (`client = JSONRPCClient(web3, private_keys[1])`), so they already had the same private key.

raiden/transfer/events.py
mauricioirace: It's needed in order to know which channel proxy to use in the event handler.

raiden/network/proxies/utils.py
mauricioirace: No, it couldn't.

marcosmartinez7 (approved):
LGTM

PR #152, opened 2020-11-09
RLP-865: fragment matrix/transport.py

mortelli:
This PR attempts to solve issue [RLP-865](https://jirainfuy.atlassian.net/browse/RLP-865), which is about doing away with the `transport/matrix/transport.py` file, in favor of the `layer.py` + `node.py` file structure used in the `transport/rif_comms` package.

### Changes
- the `transport/matrix/node.py` file is added
  - `MatrixNode` class is moved from `transport/matrix/transport.py` to this new file
  - `MatrixLightClientNode` class is moved from `transport/matrix/transport.py` to this new file
- the `_RetryQueue` class is moved from `transport/matrix/transport.py` to `transport/matrix/utils.py`
- the `transport/matrix/transport.py` file is deleted

### Caveats
- ideally, we'd want the `_RetryQueue` class to be transport-node agnostic, and use a generic transport node rather than a matrix-specific one. this is not possible due to the use that is given to this variable (`self.transport`); calls to this field include the use of `.log`, `._config`, `greenlet`, `._stop_event`, `._prioritize_global_messages`, `._global_send_queue`, `._queueids_to_queues`, `._send_raw`, `._raiden_service` and `._stop_event`, none of which are part of the `Node` abstract class from `transport.node.py`.
  - as a result, `_RetryQueue` was moved to `transport/matrix/utils.py` instead.


marcosmartinez7 (approved):
lgtm

mortelli:
unit tests have turned out OK, as did some basic scenarios in manual testing.

PR #153, opened 2020-11-09
Bug fix/rlp 801/fixing settlement light

JONAF2103:
### Problem
Partner LC can't settle due to channel not found in memory (being deleted because the other side has settled the channel already)
#### Explanation:
When a channel has been settled an event is triggered, the hub detects and handle that event by checking if something is pending for unlock or not, in the case the hub detects that the channel has anything else to unlock the hub deletes the channel from the memory. In this case the LC1 executes the settlement and the event is triggered on both hubs, the hub2 deletes the channel from the memory and the LC2 can't execute the settlement light process since the channel is not there anymore.
#### Solution

- We changed the SDK to fix some bugs too so this PR is also part of the solution provided by [this PR](https://github.com/rsksmart/lumino-light-client-sdk/pull/102)

- We now assume that if the channel is not in memory it was already settled, in the case we detect that the channel is not there but it should be we log a debug log to inform the developer about it.

### Also in this PR

1. Fixing types
2. Adding error codes
3. Fixing .gitignore for endpointConfig.js exclusion


raiden/api/webui/static/endpointConfig.js
```diff
@@ -1,5 +1,5 @@
-const backendUrl='http://localhost:5000'; 
-const nodeAddress = '0x720DD4A77bB906b5a7FC847a6f758ab20b84f214'; 
+const backendUrl='http://localhost:5004'; 
+const nodeAddress = '0x0C46357428Dd9b4A1f951Df8B4868b0D972644AD'; 
 const rnsDomain = null 
 const chainEndpoint = 'http://localhost:4444';
```
mortelli: watch out here

raiden/api/validations/channel_validator.py
```diff
@@ -139,17 +139,28 @@ def validate_and_get_channels_to_settle(token_address: TokenAddress,
...
+                              chain_state=chain_state,
+                              payment_network_id=registry_address,
+                              token_address=token_address,
+                              creator_address=creator_address
+                          ))
+
+        channels_to_settle.extend(channels)
+
+        # we look for the same but with the creator as the partner since we can
+        # be in the other side and the creator of the channel isn't our address
```
mortelli: this comment isn't so clear. could you please write it in spanish so i could understand it better? (and maybe come up with a suggestion)

mortelli (commented):
nice use of lambdas and constants :clap:

raiden/api/webui/static/endpointConfig.js
JONAF2103: that file is really annoying, but i think now it's just spacing, we should add that to the .gitignore file i think

raiden/api/validations/channel_validator.py
JONAF2103: Let me change it to something better

raiden/api/webui/static/endpointConfig.js
mortelli: afaic you can just add this to the .gitignore file in this PR :man_shrugging:

JONAF2103: it was already added but the rule was wrong, now it should work but it will start working after this PR

raiden/api/python.py
```diff
@@ -625,9 +626,14 @@ def settle_light(
                                                         and channel.identifier == channel_identifier, settled_channels)
             filtered_settled_channels_list = list(filtered_settled_channels_iterator)
             if filtered_settled_channels_list:
-                raise RaidenRecoverableError("Failed trying to settle a channel that's already settled")
+                raise RaidenRecoverableError(ErrorCode.CHANNEL_ALREADY_SETTLED)
             else:
-                raise RaidenRecoverableError("Failed trying to settle a channel that's not in waiting_for_settle state")
+                log.debug("Settltment Light: channel is not in waiting_for_settle, probably"
+                          " was removed from memory before this call, channel not found")
```
marcosmartinez7:
```suggestion
                log.debug("Settlement Light: channel is not in waiting_for_settle state, it was probably settled by the counterparty and removed from memory before this call.")
```

raiden/api/webui/static/endpointConfig.js
```diff
@@ -1,13 +1,9 @@
-const backendUrl='http://localhost:5000'; 
-const nodeAddress = '0x720DD4A77bB906b5a7FC847a6f758ab20b84f214'; 
-const rnsDomain = null 
-const chainEndpoint = 'http://localhost:4444'; 
+const backendUrl='http://localhost:5000';
```
marcosmartinez7, line 1: please remove this changes

marcosmartinez7 (changes requested):
lgtm in general, just some minor changes

marcosmartinez7 (approved):
lgtm

PR #155, opened 2020-11-10
RifCommsNode base

marcosmartinez7:
# Description

The objective of this PR is to introduce a first approach to implement the `RifCommsNode` class, the implementation of `TransportNode` that uses RIF Communications pub sub node. 

On this PR, the abstract methods of `TransportNode` are implemented and must be iterated on following PRs.

# Changelog

Implemented:
1. `__init__ `that starts a rif comms client and define the instance attributes of the class: 
  - `_config`
  - `_raiden_service`
  - `_client`
  - `_address_to_message_queue`
  - `_stop_event`
  - `_greenlets`

2. `start` with some pending TODOs
3. `_run` with some pending TODOs
4.  `stop` with some pending TODOs
5. `send_message` (pending `_send_with_retry `and` _send_raw`)
6. no-op `start_health_check` and `whitelist`
7. `link_exception `and `join`
8. `send_raw`, `enqueue_message`, `get_queue` with some pending TODOs
9. `handle_message`, `handle_delivered` with some pending TODOs
10. rest of the relevant methods (to review with @mortelli )
11.  RiffCommsNode listen to received messages on background

# TODO's

1. Global worker and prioritze global msgs? 
2. Join_global_room at start, maybe we can use a bootstrap node for rif comms here?
3. ~~Log class variable?~~
4. what about _run if we dont handle global msgs? 
5. ~~what are self.greenlets used for? Does we really need it? we only need one for leave topics?~~
6. ~~does stop need to invoke the end comms and unsubscribe from grpc channel?~~
7.  ~~self._client.start_listener_thread() ? we need a thread for listen messages~~
8. is spawn needed? it is only used on leave unused rooms, we must unsubscribe when the channels are closed?
9. How does get_all_messagequeues works and how are populated the queues on the raiden service?
10. validate_and_parse_message must be modified to use an address instead of a matrix user

image

### DIFF
--- a/raiden/tests/unit/rif_comms/test_rif_comms_client.py
+++ b/raiden/tests/unit/rif_comms/test_rif_comms_client.py
@@ -44,25 +44,29 @@ def rif_comms_client(request):
 @pytest.mark.usefixtures("rif_comms_client")
 class TestRiffCommsClient(unittest.TestCase):
 
+    @pytest.mark.skip(reason="ignore")
     def test_initialization(self):
         assert self.rif_comms_client is not None
 
+    @pytest.mark.skip(reason="ignore")
     def test_locate_peer_id(self):
         response = self.rif_comms_client.connect()
         peer_id = self.rif_comms_client.get_peer_id(LUMINO_1_ADDRESS)
         print(f"test_locate_peer_id peer_id = {peer_id}")
         assert peer_id is not None
 
+    @pytest.mark.skip(reason="ignore")
     def test_locate_unregistered_peer_id(self):
         self.assertRaises(_InactiveRpcError, lambda: self.rif_comms_client.get_peer_id(UNREGISTERED_ADDRESS))
 
+    @pytest.mark.skip(reason="ignore")
     def test_create_random_topic_id_without_connection(self):
         notification = self.rif_comms_client.connect()
         peer_id = self.rif_comms_client.get_peer_id(LUMINO_1_ADDRESS)
         channel = self.rif_comms_client.subscribe(get_random_address_str())
         peer_id = self.rif_comms_client.get_peer_id(LUMINO_1_ADDRESS)
 
-    #@pytest.mark.skip(reason="ignore")
+    @pytest.mark.skip(reason="ignore")
     def test_subscribe(self):
         channel = grpc.insecure_channel(LUMINO_1_COMMS_API)
         stub = CommunicationsApiStub(channel)
@@ -71,6 +75,7 @@ class TestRiffCommsClient(unittest.TestCase):
         channel = stub.CreateTopicWithRskAddress(rsk_address)
         subscribers = stub.GetSubscribers(Channel(channelId=LUMINO_1_ADDRESS))
 
+    @pytest.mark.skip(reason="ignore")
     def test_has_subscriber(self):
         channel = grpc.insecure_channel(LUMINO_1_COMMS_API)
         stub = CommunicationsApiStub(channel)
@@ -92,3 +97,10 @@ class TestRiffCommsClient(unittest.TestCase):
         )
         print("has_subscriber")
         print(has_subscriber)
+
+    def test_disconnect(self):
+        notification = self.rif_comms_client.connect()
+        peer_id = self.rif_comms_client.get_peer_id(LUMINO_1_ADDRESS)
+        self.rif_comms_client.disconnect()
+
+
--- a/raiden/tests/unit/rif_comms/test_rif_comms_client.py
+++ b/raiden/tests/unit/rif_comms/test_rif_comms_client.py
@@ -115,10 +115,12 @@ class TestRiffCommsClient(unittest.TestCase):
         peer_id = self.rif_comms_client.get_peer_id(LUMINO_1_ADDRESS)
         self.rif_comms_client.disconnect()
 
+    @pytest.mark.skip(reason="ignore")
     def test_send_lumino_message(self):
         channel = grpc.insecure_channel(LUMINO_2_COMMS_API)
         stub = CommunicationsApiStub(channel)
-        channel = stub.Subscribe(Channel(channelId="16Uiu2HAm9otWzXBcFm7WC2Qufp2h1mpRxK1oox289omHTcKgrpRA")) # got this from subscription of lumino node
+        channel = stub.Subscribe(Channel(
+            channelId="16Uiu2HAm9otWzXBcFm7WC2Qufp2h1mpRxK1oox289omHTcKgrpRA"))  # got this from subscription of lumino node
 
         some_raiden_message = {
             'type': 'LockedTransfer',
@@ -152,4 +154,3 @@ class TestRiffCommsClient(unittest.TestCase):
                 message=Msg(payload=str.encode(json.dumps(some_raiden_message)))
             )
         )
-
--- a/raiden/tests/unit/rif_comms/test_rif_comms_client.py
+++ b/raiden/tests/unit/rif_comms/test_rif_comms_client.py
@@ -9,7 +9,7 @@ from grpc._channel import _InactiveRpcError
 from sha3 import keccak_256
 
 from transport.rif_comms.client import RifCommsClient
-from transport.rif_comms.proto.api_pb2 import RskAddress
+from transport.rif_comms.proto.api_pb2 import RskAddress, Channel, Subscriber
 from transport.rif_comms.proto.api_pb2_grpc import CommunicationsApiStub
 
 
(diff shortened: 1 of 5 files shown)


mortelli:
closed in favor of #157

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -37,40 +38,109 @@ def rif_comms_client(request):
     def teardown():
         rif_comms_client.disconnect()
 
-    request.addfinalizer(teardown)
+    #request.addfinalizer(teardown)
```
mortelli: could you please explain what the issue is here?

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -37,40 +38,109 @@ def rif_comms_client(request):
...
+            Channel channel = 2;
+        }"""
+        has_subscriber = stub.HasSubscriber(
+            Subscriber(
+                peerId=peer_id.address,
+                channel=Channel(channelId=LUMINO_1_ADDRESS)
+            )
+        )
+        print("has_subscriber")
+        print(has_subscriber)
```
mortelli:
```suggestion
        print("has_subscriber", has_subscriber)
```

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -37,40 +38,109 @@ def rif_comms_client(request):
...
-        for resp in channel:
-            print(resp)
+        channel = stub.CreateTopicWithPeerId(rsk_address)
+        peer_id = stub.LocatePeerId(rsk_address)
+
+        """
+        message Subscriber {
+            string peerId = 1;
+            Channel channel = 2;
+        }"""
```
mortelli:
```suggestion
            Channel channel = 2;
        }
        """
```

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -1,3 +1,4 @@
+import json
 import secrets
```
mortelli, line 2: shall we add a note to this file? explaining that for the time being a comms node should be started independently from these tests?

transport/rif_comms/layer.py
```diff
@@ -10,7 +10,7 @@ def construct_full_node(self, config):
         return RifCommsNode(config["address"], config["transport"]["rif_comms"])
 
     def construct_light_clients_nodes(self, config):
-        pass
+        return []
```
mortelli, line 13: do you think we could add type hints to both the concrete and abstract versions of this method? (and its equivalent for full nodes)

transport/rif_comms/node.py
```diff
@@ -1,51 +1,339 @@
...
 
     def __init__(self, address: Address, config: dict):
         TransportNode.__init__(self, address)
         Runnable.__init__(self)
         self._config = config
+        self._raiden_service: RaidenService = None
+
+        self._rif_comms_connect_stream: Notification = None
+        self._our_topic: Notification = None
+        self._our_topic_thread: Greenlet = None
```
mortelli: could you please explain the reason behind these `None`s?

transport/rif_comms/node.py
```diff
@@ -1,51 +1,339 @@
...
+from raiden.utils import Address, pex
 from raiden.utils.runnable import Runnable
-from raiden.utils.typing import Address
+from transport.matrix.utils import _RetryQueue
+from transport.message import Message as TransportMessage
 from transport.node import Node as TransportNode
 from transport.rif_comms.client import RifCommsClient
+from transport.rif_comms.proto.api_pb2 import Notification, ChannelNewData
+
+_RoomID = NewType("_RoomID", str)
```
mortelli: don't think we need this one

transport/rif_comms/node.py
```diff
@@ -1,51 +1,339 @@
...
 
     def link_exception(self, callback: Any):
-        raise NotImplementedError
+        self.greenlet.link_exception(callback)
 
     def join(self, timeout=None):
-        raise NotImplementedError
+        self.greenlet.join(timeout)
+
+    def listen_messages(
```
mortelli: for grammatical reasons i suggest we refactor → rename this to `listen_for_messages`

transport/rif_comms/node.py
```diff
@@ -1,51 +1,339 @@
...
+        self
+    ):
+        """
+        Iterate over the Notification stream and blocks thread to receive messages
+        """
+        for notification_stream in self._our_topic:
```
mortelli, line 283: i am not entirely sure how this `for` will work in a greenlet. do you have an inkling of this?

transport/rif_comms/node.py
```diff
@@ -1,51 +1,339 @@
...
+        self.start_listener_thread()
+        self._greenlets = [self._our_topic_thread]
+        self.log.debug("RIF Comms Node started", config=self._config)
+        Runnable.start(self)
+
+    def __repr__(self):
+        if self._raiden_service is not None:
+            node = f" node:{pex(self._raiden_service.address)}"
+        else:
+            node = ""
```
mortelli:
```suggestion
        node = f" node:{pex(self._raiden_service.address)}" if self._raiden_service else ""
```

transport/rif_comms/node.py
```diff
@@ -1,51 +1,339 @@
...
+        """
+        if self._stop_event.ready():
+            return
+        self._stop_event.set()
+
+        for message_queue in self._address_to_message_queue.values():
+            if message_queue:  # if message_queue.greenlet is not None
+                message_queue.notify()  # if we need to send something, this is the time
+
+        self.stop_listener_thread()  # stop sync_thread, wait client's greenlets
```
mortelli:
```suggestion
        self.stop_listener_thread()  # stop sync_thread, wait for client's greenlets
```

transport/rif_comms/node.py
```diff
@@ -1,51 +1,339 @@
...
+        but it should raise any stop-time exception
+
+        Disconnects from RIF Communications node
+        """
+        if self._stop_event.ready():
+            return
+        self._stop_event.set()
+
+        for message_queue in self._address_to_message_queue.values():
+            if message_queue:  # if message_queue.greenlet is not None
```
mortelli:
```suggestion
            if message_queue.greenlet:
```

imo, going all the way from `RifCommsNode` to `Runnable` to `__bool__` is too obscure a path for this line to make sense to the reader

transport/rif_comms/node.py
```diff
@@ -1,51 +1,339 @@
...
+                message_queue.notify()  # if we need to send something, this is the time
+
+        self.stop_listener_thread()  # stop sync_thread, wait client's greenlets
+
+        # wait own greenlets, no need to get on them, exceptions should be raised in _run()
```
mortelli:
```suggestion
        # wait for our own greenlets, no need to get on them, exceptions should be raised in _run()
```

transport/rif_comms/node.py
```diff
@@ -1,51 +1,339 @@
...
+            pass
+        # parent may want to call get() after stop(), to ensure _run errors are re-raised
+        # we don't call it here to avoid deadlock when self crashes and calls stop() on finally
+
+    def send_message(self, message: TransportMessage, recipient: Address):
+        """Queue the message for sending to recipient in the queue_identifier
+
+        It may be called before transport is started, to initialize message queues
+        The actual sending is started only when the transport is started
+        """
```
mortelli: let's replace this docstring with the one present in `transport/node.py` (or something similar)

mortelli (commented):
this looks to be a step in the right direction :+1: 

some general comments first:
- there are some details which are still not so clear to me atm; e.g. some fields set in `RifCommsNode.__init__`, the logic for `RifCommsNode._run`, etc. these should be clarified as we go along in further PRs.
- we have a non-trivial pull from `rif-comms` to sort out together, which might imply some significant changes to the current code.
- please make sure to apply _Reformat Code_ and _Optimize Imports_ to all affected files.

---

now, regarding what is laid out in the PR description:

> [function] with some pending TODOs

this is mentioned multiple times in the changelog. do you think we could document these TODOs somewhere?

> 1. Global worker and prioritze global msgs?

currently, this behavior falls under the `TransportNode.enqueue_global_message` method. we should have a no-op implementation of this method in the RIF Comms node.

> 2. Join_global_room at start, maybe we can use a bootstrap node for rif comms here?

i don't think we need to implement an equivalent for this _yet_.

> 4. what about _run if we dont handle global msgs?

not sure of this yet. i say we leave it as it is and come back to it later.

> 8. is spawn needed? it is only used on leave unused rooms, we must unsubscribe when the channels are closed?

this is a good question for which i don't have a clear answer.

in principle, i don't think we need to spawn a thread (or equivalent) just for unsuscribing. but i also don't know that we've clearly defined a definitive correspondence between threads and transport node tasks/functions/responsibilities.

> 9. How does get_all_messagequeues works and how are populated the queues on the raiden service?

this—as well as the underlying `queueids_to_queues` field—could definitely be better named, but as i understand it, this logic resides in `transfer/node.py`.

it seems that the queues are modified based on the logic written in different handler functions, e.g. `handle_contract_receive_channel_closed`, `handle_processed`, `update_queues`.

> 10. validate_userid_signature must be modified to use an address instead of a matrix user

isn't this used internally? i can only find it in `matrix/utils.py` and tests.

raiden/tests/unit/rif_comms/test_rif_comms_client.py
marcosmartinez7: I commented it because some issues with the rif comms node on the disconnect, but i will leave it uncommented

raiden/tests/unit/rif_comms/test_rif_comms_client.py
marcosmartinez7: i can add that comment for sure

transport/rif_comms/layer.py
marcosmartinez7: in a future yes, but we must create a RifCommsLightClientNode and an abstraction for it and MatrixLightClientNode

maybe we can look on this when we start the hub-lc stuff

transport/rif_comms/node.py
marcosmartinez7: Definition of the variable basically, its ok to use just `self._rif_comms_connect_stream: Notification`  and dont assign it to any value? 

I think python3 supports that

transport/rif_comms/node.py
marcosmartinez7: The for in loop is the standar way to listen to `streams `on GRPC 

- https://stackoverflow.com/questions/47831895/how-do-i-handle-streaming-messages-with-python-grpc
- https://github.com/grpc/grpc/blob/v1.20.0/examples/python/route_guide/route_guide_client.py#L96

Regarding the greenlet, i think the for loop blocks the current thread until a message arrives, therefore we need a greenlet running this background task in order to not block the ppal thread

raiden/tests/unit/rif_comms/test_rif_comms_client.py
marcosmartinez7: Done

raiden/tests/unit/rif_comms/test_rif_comms_client.py
marcosmartinez7: done

transport/rif_comms/node.py
marcosmartinez7: fixed

transport/rif_comms/node.py
marcosmartinez7: I basically put this comment here for that. I found usable to remember that, but i can remove it.

transport/rif_comms/node.py
marcosmartinez7: done

transport/rif_comms/node.py
mortelli: you cannot get away with instance variables type hints without explicit assignments (you can do so only with class variables and it would not be correct here).

i try to avoid `None` assignments as much as possible in constructor methods (i think they're untidy) but i'm not sure if you're only adding these for the type hints.

the problem in this case is that some of these variables are not set until a later point (like the `start` function) so the underlying issue is bigger than this one; it's a design issue.

it's not a blocker for me anyway, so if you just want the type hints i guess this is ok

mortelli (approved):
:ok:

PR #157, opened 2020-11-11
RLP-889: generalize RetryQueue

mortelli:
This PR aims to solve issue [RLP-889](https://jirainfuy.atlassian.net/browse/RLP-889), which is about generalizing the `_RetryQueue` class so that it can be used with any type of transport node, instead of just `MatrixNode`.

## Design overview
The evident path to accomplish this task is to change matrix references from the `_RetryQueue` class to generic transport references (i.e. `MatrixTransportNode` → `TransportNode`). The problem with this approach is that the queue makes references to specific matrix node attributes rather than generic transport node ones. As a result, the abstract and concrete classes had to be modified (details below).

It's worth mentioning that the implementation is far from clean due to the coupling between these classes—both make references to each other; calls in one strict direction would be a more preferable design, but it's too big a task to tackle in the foreseeable future.

`_RetryQueue` **was renamed** to `MessageQueue`, since it better reflects the class purpose.

### Message-sending functions
Up until now, messages were first _enqueued_ through the transport node `send_message` function, but the actual sending of the message was only invoked in the `MessageQueue` class, through the (matrix) transport method `_send_raw`.

For better clarity, `TransportNode.send_message` was renamed to `enqueue_message`, and `_send_raw` was renamed to `send_message` and added to the transport node abstraction. Docstrings were also iterated to clear up how this is supposed to work: external entities should invoke the enqueuing, and only the message queue should invoke the actual sending from the transport node object. 

It would be better to have (the new) `send_message` as an internal method (and not part of the abstraction), but since this function is called by the—now transport-generic—message queue, it must be added to the generic transport node as well.

### `Runnable`
Up until now, `Runnable` was a superclass for `MatrixNode`. Since now `RifCommsNode` is also a subclass of `Runnable`, this inheritance was moved from the concrete nodes to the `TransportNode` abstract class. This is not only a cleaner design, but also helps with accessing transport node fields from the `MessageQueue` class.

As a result, `Runnable.__init__` is now called in the `TransportNode` constructor.

### `TransportNode`
New functions were added so that the `MessageQueue` class can use this abstraction as a type for its transport field:
1. `raiden_service` returns the `RaidenService` instance set in the node.
2. `config` returns the config dictionary set in the node.
3. `log` returns the node logger.
4. `enqueue_global_messages` is a function added only for the purposes of calling matrix-specific behavior in the `MessageQueue` class. As a result, it should be a no-op for `RifCommsNode`.

## Minor changes
- the `Runnable` class was refactored for tidiness purposes; a helper function was added and unnecessary code was removed.
- the `MessageQueue` class was moved out of the `transport.matrix` package, into the generic `transport.utils.py` file.
- the `MessageQueue.transport` field was renamed to `transport_node` for more clarity.
- `receiver` was substituted for `recipient` in some fields and logs for more clarity.
- `MatrixNode._stop_event` field was moved to the `Runnable` class, so that `MessageQueue.transport_node.stop_event` is congruent with the current abstractions (both transport nodes are `Runnables`, so this is a valid call). This field was also renamed so that it is "public".
- _Reformat Code_ and _Optimize Imports_ have been applied to all modified files.

### DIFF
--- a/raiden/utils/runnable.py
+++ b/raiden/utils/runnable.py
@@ -16,8 +16,8 @@ class Runnable:
 
     def __init__(self) -> None:
         self._set_greenlet()
-        self._stop_event = Event()
-        self._stop_event.set()
+        self.stop_event = Event()
+        self.stop_event.set()
 
     def start(self) -> None:
         """ Synchronously start task
--- a/raiden/utils/runnable.py
+++ b/raiden/utils/runnable.py
@@ -13,7 +13,6 @@ class Runnable:
     In the future, when proper restart is implemented, may be replaced by actual greenlet
     """
 
-    greenlet: Greenlet = None
     args: Sequence = tuple()  # args for _run()
     kwargs: dict = dict()  # kwargs for _run()
 
@@ -21,8 +20,7 @@ class Runnable:
         self.args = args
         self.kwargs = kwargs
 
-        self.greenlet = Greenlet(self._run, *self.args, **self.kwargs)
-        self.greenlet.name = f"{self.__class__.__name__}|{self.greenlet.name}"
+        self._set_greenlet()
 
     def start(self) -> None:
         """ Synchronously start task
@@ -38,9 +36,12 @@ class Runnable:
             and self.greenlet.kwargs == self.kwargs
         )
         if not pristine:
-            self.greenlet = Greenlet(self._run, *self.args, **self.kwargs)
-            self.greenlet.name = f"{self.__class__.__name__}|{self.greenlet.name}"
+            self._set_greenlet()
         self.greenlet.start()
+
+    def _set_greenlet(self):
+        self.greenlet = Greenlet(self._run, *self.args, **self.kwargs)
+        self.greenlet.name = f"{self.__class__.__name__}|{self.greenlet.name}"
 
     def _run(self, *args: Any, **kwargs: Any) -> None:
         """ Reimplements in children to busy wait here
@@ -68,9 +69,8 @@ class Runnable:
             subtask=subtask,
             exc=subtask.exception,
         )
-        if not self.greenlet:
-            return
-        self.greenlet.kill(subtask.exception)
+        if self.greenlet:
+            self.greenlet.kill(subtask.exception)
 
     # redirect missing members to underlying greenlet for compatibility
     # but better use greenlet directly for now, to make use of the c extension optimizations
--- a/raiden/utils/runnable.py
+++ b/raiden/utils/runnable.py
@@ -1,4 +1,4 @@
-from typing import Any, Sequence
+from typing import Any
 
 import structlog
 from gevent import Greenlet
--- a/raiden/utils/runnable.py
+++ b/raiden/utils/runnable.py
@@ -2,6 +2,7 @@ from typing import Any
 
 import structlog
 from gevent import Greenlet
+from gevent.event import Event
 
 log = structlog.get_logger(__name__)
 
@@ -15,6 +16,8 @@ class Runnable:
 
     def __init__(self) -> None:
         self._set_greenlet()
+        self._stop_event = Event()
+        self._stop_event.set()
 
     def start(self) -> None:
         """ Synchronously start task
--- a/raiden/utils/runnable.py
+++ b/raiden/utils/runnable.py
@@ -13,13 +13,7 @@ class Runnable:
     In the future, when proper restart is implemented, may be replaced by actual greenlet
     """
 
-    args: Sequence = tuple()  # args for _run()
-    kwargs: dict = dict()  # kwargs for _run()
-
-    def __init__(self, *args: Any, **kwargs: Any) -> None:
-        self.args = args
-        self.kwargs = kwargs
-
+    def __init__(self) -> None:
         self._set_greenlet()
 
     def start(self) -> None:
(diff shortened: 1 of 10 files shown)


raiden/utils/runnable.py
```diff
@@ -68,9 +61,8 @@ def on_error(self, subtask: Greenlet) -> None:
             subtask=subtask,
             exc=subtask.exception,
         )
-        if not self.greenlet:
-            return
-        self.greenlet.kill(subtask.exception)
+        if self.greenlet:
```
marcosmartinez7, line 64: 👍

transport/matrix/node.py
```diff
@@ -101,6 +98,22 @@ def _http_retry_delay() -> Iterable[float]:
...
+    def enqueue_global_messages(self):
+        if self._prioritize_global_messages:
+            self._global_send_queue.join()
+
+    @property
+    def raiden_service(self) -> 'RaidenService':
+        return self._raiden_service
+
+    @property
+    def config(self) -> {}:
```
marcosmartinez7, line 110: why {} instead of dict?

transport/matrix/node.py
```diff
@@ -659,23 +667,23 @@ def _send_with_retry(self, queue_identifier: QueueIdentifier, message: Message):
...
 
-    def _send_raw(self, receiver_address: Address, data: str):
+    def send_message(self, payload: str, recipient: Address):
         with self._getroom_lock:
-            room = self._get_room_for_address(receiver_address)
+            room = self._get_room_for_address(recipient)
         if not room:
             self.log.error(
-                "No room for receiver", receiver=to_normalized_address(receiver_address)
+                "No room for receiver", receiver=to_normalized_address(recipient)
```
marcosmartinez7: lets use recipient on the log string too

transport/matrix/node.py
```diff
@@ -659,23 +667,23 @@ def _send_with_retry(self, queue_identifier: QueueIdentifier, message: Message):
...
             )
             return
         self.log.debug(
-            "Send raw", receiver=pex(receiver_address), room=room, data=data.replace("\n", "\\n")
+            "Send raw", receiver=pex(recipient), room=room, data=payload.replace("\n", "\\n")
```
marcosmartinez7: same

marcosmartinez7: also, use payload insted of data

transport/matrix/node.py
```diff
@@ -1217,30 +1225,30 @@ def _run(self):
...
```
marcosmartinez7: same here

transport/matrix/node.py
```diff
@@ -1217,30 +1225,30 @@ def _run(self):
...
 
-    def _send_raw(self, receiver_address: Address, data: str):
+    def send_message(self, payload: str, recipient: Address):
         with self._getroom_lock:
-            room = self._get_room_for_address(receiver_address)
+            room = self._get_room_for_address(recipient)
         if not room:
             self.log.error(
-                "No room for receiver", receiver=to_normalized_address(receiver_address)
+                "No room for receiver", receiver=to_normalized_address(recipient)
```
marcosmartinez7: same here

transport/node.py
```diff
@@ -35,13 +37,29 @@ def stop(self):
         """
 
     @abstractmethod
-    def send_message(self, message: Message, recipient: Address):
+    def enqueue_message(self, message: Message, recipient: Address):
         """
-        Send a message to the recipient.
+        Enqueue a message to be sent the recipient.
```
marcosmartinez7:
```suggestion
        Enqueue a message to be sent to the recipient.
```

marcosmartinez7: We can add to the docstring something like 

"This method must be called by the raiden_service in order to send a message¨, just like the send_message clarification with the RetryQueue usage

transport/node.py
```diff
@@ -35,13 +37,29 @@ def stop(self):
...
+        Enqueue global messages to be sent.
+        This method might be implemented as a no-op if the concept of global messages does not apply
+        to a specific transport layer implementation.
+        """
+
+    @abstractmethod
+    def send_message(self, payload: str, recipient: Address):
+        """
+        Send a message payload to the recipient.
+        Due to a coupled design, this method should only be called by the _RetryQueue class.
```
marcosmartinez7, line 59: 👍

transport/matrix/node.py
mortelli: > With CPython 2.7, using dict() to create dictionaries takes up to 6 times longer and involves more memory allocation operations than the literal syntax. Use {} to create dictionaries, especially if you are pre-populating them, unless the literal syntax does not work for your case.

[source](https://stackoverflow.com/a/17098017)

mortelli: (in any case, we're just dealing with a type hint here, but i think `{}` is a preferable choice when it can be made in lieu of the slightly more complex `dict`)

transport/matrix/node.py
mortelli: done

transport/matrix/node.py
mortelli: done

transport/matrix/node.py
mortelli: done

transport/matrix/node.py
mortelli: done

transport/node.py
mortelli: > We can add to the docstring something like
> 
> "This method must be called by the raiden_service in order to send a message¨, just like the send_message clarification with the RetryQueue usage

the way i see it: there's a difference with `send_message`; while the latter should _only_ be called by the `_RetryQueue` class, this one can arbitrarily be called from any other entity.

in our case, both the `RaidenService` and the `RaidenEventHandler` classes make use of this function (through reference to transport nodes)

marcosmartinez7 (approved):
lgtm

mortelli:
- unit tests seem OK compared to `develop`
- manually verified sending tokens in the following scenario: `FN` ↔ `FN` ↔ `LC`

PR #158, opened 2020-11-12
RLP-850 Fixing insufficient funds error

JONAF2103:
### Problem
When we send a payment from an LC and we don't wait and send again another payment on the same channel that has an amount greater than the total available balance on the channel the hub fails and stop working.

### Cause
This was because we have an assert on the state machine that checks the balances before starting a payment.

### Solution
We have added some checks before creating and initiating the payment to avoid that error

### Changes

* Adding control before creating payment
* Adding control before initiating payment

* This affects the payments light

### DIFF
--- a/raiden/api/python.py
+++ b/raiden/api/python.py
@@ -57,6 +57,7 @@ from raiden.messages import RequestMonitoring, LockedTransfer, RevealSecret, Unl
 from raiden.settings import DEFAULT_RETRY_TIMEOUT, DEVELOPMENT_CONTRACT_VERSION
 
 from raiden.transfer import architecture, views, routes
+from raiden.transfer.channel import get_distributable
 from raiden.transfer.events import (
     EventPaymentReceivedSuccess,
     EventPaymentSentFailed,
@@ -1235,6 +1236,12 @@ class RaidenAPI:
             payment_network_id=payment_network_identifier,
             token_address=token_address,
         )
+
+        # checking the balance before doing the payment to avoid problems on the state machine
+        chain_state = views.state_from_raiden(self.raiden)
+
+
+
         self.raiden.mediated_transfer_async_light(
             token_network_identifier=token_network_identifier,
             amount=amount,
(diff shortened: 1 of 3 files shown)


raiden/api/python.py
```diff
@@ -1235,6 +1236,12 @@ def transfer_async_light(
             payment_network_id=payment_network_identifier,
             token_address=token_address,
         )
+
+        # checking the balance before doing the payment to avoid problems on the state machine
+        chain_state = views.state_from_raiden(self.raiden)
```
marcosmartinez7: this is not needed as lgtm bot is reporting

marcosmartinez7 (changes requested):
lgtm but we have an unused var

raiden/api/python.py
JONAF2103: sorry about that my mistake

JONAF2103: done

raiden/api/python.py
```diff
@@ -1856,6 +1858,10 @@ def create_light_client_payment(
 
         if channel_state:
 
+            # checking the balance before creating the payment
+            if amount > get_distributable(channel_state.our_state, channel_state.partner_state):
+                raise InsufficientFunds("Insufficient funds to create payment")
```
mortelli, line 1863: are you sure we need this text given the exception type?

raiden/raiden_service.py
```diff
@@ -157,6 +159,10 @@ def initiator_init_light(
     current_channel = views.get_channelstate_by_canonical_identifier_and_address(chain_state, canonical_identifier,
                                                                                  creator_address)
 
+    # checking the balance before initiating the payment
+    if transfer_amount > get_distributable(current_channel.our_state, current_channel.partner_state):
+        raise InsufficientFunds("Insufficient funds to initiate payment")
```
mortelli, line 164: same here

mortelli (approved):

raiden/api/python.py
JONAF2103: Yes because if you don't put text there the errors array at the api response will be empty

raiden/raiden_service.py
JONAF2103: I answered above

mauricioirace (approved):
LGTM

marcosmartinez7 (approved):
lgtm now

PR #159, opened 2020-11-13
Prevent crash when deleting payment task

mauricioirace:
Elements from a dictionary were being deleted while iterating over it, forcing the node to crash.

Changes:
- We now create a list from the dict's elements in order to prevent the concurrent modification problem


marcosmartinez7 (approved):
lgtm, thanks for the exhaustive use of the labels :)

JONAF2103 (approved):
LGTM

PR #160, opened 2020-11-16
RLP-885 test_mediatedtransfer_invalid.py

JONAF2103:
## Problem
The test seems to be flaky in this branch because some times it fails on an assert check.

## Cause
The assert fails because it's checking the channel balance after triggering some operations. The problem is that sometimes those operations take too long and the assert it's being checking the condition before the data can be valid.

## Solution
The problem was the way to call this function `assert_synced_channel_state` where it was directly called instead of using the helper function for tests called `wait_assert`. That other function waits for the value to be ready to assert and then it checks the value. Also it has a gevent timeout to make sure that if something is not ok it will fail anyway.

## Why it works on develop?
Probably because with all the changes that we are introducing on this branch the test here is taking more time to complete compared to develop branch.

### DIFF
--- a/raiden/tests/integration/transfer/test_mediatedtransfer_invalid.py
+++ b/raiden/tests/integration/transfer/test_mediatedtransfer_invalid.py
@@ -27,20 +27,19 @@ from raiden.utils.signer import LocalSigner
 
 @pytest.mark.parametrize("channels_per_node", [1])
 @pytest.mark.parametrize("number_of_nodes", [2])
-def test_failsfast_lockedtransfer_exceeding_distributable(
-    raiden_network, token_addresses, deposit
-):
+def test_failsfast_lockedtransfer_exceeding_distributable(raiden_network, token_addresses, deposit, network_wait):
     raise_on_failure(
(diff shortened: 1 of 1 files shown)


marcosmartinez7 (approved):
lgtm

mortelli (approved):
this seems OK, but the `wait_assert` calls are looking a bit hard to read with the given parameters (e.g. `[]`)

have you tried doing these calls with named parameters? i see that the function can receive keyword arguments, so i think it might be possible.

JONAF2103:
@mortelli the args for `wait_assert` are dynamic since it's a function that calls another function, here the definition `def wait_assert(func: Callable, *args, **kwargs) -> None:` basically the args are the same args for the called function, in this case is 
```python
def assert_synced_channel_state(
    token_network_identifier: TokenNetworkID,
    app0: App,
    balance0: Balance,
    pending_locks0: List[HashTimeLockState],
    app1: App,
    balance1: Balance,
    pending_locks1: List[HashTimeLockState],
) -> None:
```

PR #161, opened 2020-11-18
RLP-892: RIF Comms improvements & bugfixes

mortelli:
This PR attempts to solve issue [RLP-892](https://jirainfuy.atlassian.net/browse/RLP-892), which ultimately turned out to be about applying bugfixes to and improving the RIF Comms transport node and client (in the process of investigating the workings of the transport node and message queues).

#### `RIFCommsTransportNode`
This class has been significantly iterated upon.

##### Fields
1. the `_client` field has been renamed to `_comms_client` for greater clarity.
2. ~~the `_our_topic` field has been renamed to `_our_topic_stream`, to better differentiate between this field and `our_topic_thread`.~~  this has now been removed, `.greenlet` already serves this purpose.
3. the `_greenlets` field has been removed, since it served no purpose.
4. the `_stop_event` field initialization has been removed, since the `Runnable` class already does this. references to this now-removed variable have been replaced with references to `stop_event` (which is the `Runnable` field).

##### Methods
1. the `_queueids_to_queues` method has been removed, since it served no purpose.
2. ~~a new `_start_message_listener` function has been added and invoked in the `start` method. its purpose is to start a greenlet which will listen for incoming messages (it also replaces the old `start_listener_thread` function).~~ this is now done inside the `_run` method.
3. `listen_for_messages` has been replaced with `_receive_messages`. this new function now calls the previously unused `_handle_message` function.
4. `_handle_message` has been overhauled:
    - its signature was corrected, since it received the wrong parameters.
    - it now deals with individual messages rather than assuming that a collection of messages is received.
    - the individual `_receive` functions have been replaced for the `_ack_message` function, which sends back a `Delivered` message (unless the received message is also `Delivered`-typed). this also simplifies the previous `if` statement which checked received message types.
    - useless code has been removed.
5. ~~`stop_listener_thread` has been renamed to `_stop_message_listener`, which is symmetric with `_start_message_listener`. its code has also been fixed, which called greenlet functions on the wrong type of variables.~~ this is now done within the `stop` method
6. the `_enqueue_message` code was moved to `enqueue_message` since it was 2 lines long, and only used there.
7. `_parse_topic_new_data` has been renamed to `_notification_to_message` for hopefully greater clarity. it's also now a static method, since it does not use `self`. finally, some of its internal variables have been renamed for clarity's sake.
8. the order of the methods for this class has been changed to match the abstract transport node class, as well as its own internal order of use.

#### `RIFCommsClient`
Some of the methods for this class have had their signatures and/or bodies changed:
  1. `subscribe_to` now receives an `Address` parameter instead of a `str` topic ID, since this conversion is done within the comms client code.
  2. `is_subscribed_to` now returns a `bool` instead of a `BooleanResponse`, which is simpler. it also uses the given parameter differently (more details in the diff).
  3. `send_message` now:
      - receives a `message_payload` first, and then an `rsk_address`. this is changed so that it is more similar to the transport node `send_message` method. 
      - uses the given `payload` parameter to put the `PublishPayload` object together (instead of a test string). 
      - `topic_id`—which was previously a function param—is now internally derived from the given address.
      - returns nothing instead of a `Void`-typed variable.
  4. `unsubscribe_from` now receives an `Address` instead of a `str`. the conversion of this address to a topic ID is done internally within the client. it also returns nothing instead of a `Void`-typed variable.
  5. `disconnect` now has no `return` statement (instead of returning a `Void`-typed variable).

### Minor
- for consistency's sake:
  1. `rif_comms/layer.RifCommsLayer` has been renamed to `rif_comms/layer.Layer`.
  2. `rif_comms/node.RifCommsNode` has been renamed to `rif_comms/node.Node`.
  3. `rif_comms/client.RifCommsClient` has been renamed to `rif_comms/client.Client`.
  (named imports are still used)
- the type hint for the `message` param in `transport/node.py`'s `Node.send_message` method was fixed (`RaidenMessage` → `TransportMessage`).
- truthiness for `Runnable` variables in both the Matrix and RIF Comms transport nodes has been made more explicit: `if queue` → `if queue.greenlet` (`if queue` works due to the `Runnable.__bool__` function, but it's quite obscure).
- RIF Comms docstrings were added and iterated upon, both for the node as well as the client.
- some log texts have been (hopefully) improved.
- some TODOs have been added.
- some `RIFCommsClient` methods have been renamed:
  1. `subscribe` has been renamed to `subscribe_to` for more clarity.
  2. `get_peer_id` has been renamed to be internal, as it is currently used outside of the client code only due to a bug.
  3. `has_subscription` has been renamed to `is_subscribed_to` for more clarity.
- `SignedRetrieableMessage` has been added to the the `__all__` variable in `raiden/messages.py` (a rightful IDE complaint).
- _Reformat Code_ and _Optimize Imports_ have been applied to all modified files.

## TODOs
- [x] verify unit test suite results
- [x] create issues for all the code TODOs (if they don't exist yet)
- [x] manually verify rif comms unit tests

### DIFF
--- a/transport/matrix/node.py
+++ b/transport/matrix/node.py
@@ -310,7 +310,7 @@ class MatrixNode(TransportNode):
             )
 
         self.log.info(
-            "Send message",
+            "Enqueue message",
             recipient=pex(recipient),
             message=raiden_message,
             queue_identifier=queue_identifier,
--- a/transport/matrix/node.py
+++ b/transport/matrix/node.py
@@ -11,7 +11,6 @@ from gevent.event import Event
 from gevent.queue import JoinableQueue
 from matrix_client.errors import MatrixRequestError
 from matrix_client.user import User
-
 from raiden.constants import DISCOVERY_DEFAULT_ROOM
 from raiden.exceptions import InvalidAddress, UnknownAddress, UnknownTokenAddress
 from raiden.message_handler import MessageHandler
@@ -179,7 +178,7 @@ class MatrixNode(TransportNode):
 
         # (re)start any _RetryQueue which was initialized before start
         for retrier in self._address_to_retrier.values():
-            if not retrier:
+            if not retrier.greenlet:
                 self.log.debug("Starting retrier", retrier=retrier)
                 retrier.start()
 
--- a/transport/matrix/node.py
+++ b/transport/matrix/node.py
@@ -194,7 +194,7 @@ class MatrixNode(TransportNode):
         self.greenlet.name = f"MatrixTransport._run node:{pex(self._raiden_service.address)}"
         self._raiden_service.handle_and_track_state_change(state_change)
         try:
-            # waits on _stop_event.ready()
+            # waits on stop_event.ready()
             self._global_send_worker()
             # children crashes should throw an exception here
         except gevent.GreenletExit:  # killed without exception
@@ -1216,7 +1216,7 @@ class MatrixLightClientNode(MatrixNode):
         self.greenlet.name = f"MatrixLightClientTransport._run light_client:{to_canonical_address(self.address)}"
         self._raiden_service.handle_and_track_state_change(state_change)
         try:
-            # waits on _stop_event.ready()
+            # waits on stop_event.ready()
             self._global_send_worker()
             # children crashes should throw an exception here
         except gevent.GreenletExit:  # killed without exception
--- a/transport/rif_comms/client.py
+++ b/transport/rif_comms/client.py
@@ -37,7 +37,7 @@ class RifCommsClient:
         """
         return self.stub.ConnectToCommunicationsNode(self.rsk_address)
 
-    def subscribe(self, rsk_address: Address) -> Notification:
+    def subscribe_to(self, rsk_address: Address) -> Notification:
         """
         Subscribes to a pub-sub topic in order to send messages to or receive messages from an address.
         Invokes CreateTopicWithRskAddress GRPC API endpoint.
--- a/transport/rif_comms/client.py
+++ b/transport/rif_comms/client.py
@@ -65,11 +65,11 @@ class Client:
             )
         ).value
 
-    def send_message(self, message_payload: str, rsk_address: Address):
+    def send_message(self, payload: str, rsk_address: Address):
         """
         Sends a message to a destination RSK address.
         Invokes the SendMessageToTopic GRPC API endpoint.
-        :param message_payload: the message data to be sent
+        :param payload: the message data to be sent
         :param rsk_address: the destination for the message to be sent to
         """
         topic_id = self._get_peer_id(rsk_address)
@@ -77,7 +77,7 @@ class Client:
         self.stub.SendMessageToTopic(
             PublishPayload(
                 topic=Channel(channelId=topic_id),
-                message=Msg(payload=str.encode(message_payload))
+                message=Msg(payload=str.encode(payload))
             )
         )
 
--- a/transport/rif_comms/client.py
+++ b/transport/rif_comms/client.py
@@ -8,8 +8,7 @@ from transport.rif_comms.proto.api_pb2 import (
     Msg,
     RskAddress,
     Void,
-    Subscriber,
-    BooleanResponse
+    Subscriber
 )
 from transport.rif_comms.proto.api_pb2_grpc import CommunicationsApiStub
 
@@ -22,7 +21,7 @@ class RifCommsClient:
     def __init__(self, rsk_address: Address, grpc_api_endpoint: str):
         """
         Constructs the RIF Communications Client.
-        :param rsk_address: address of the node that wants to use the RIF Communications server
+        :param rsk_address: RSK address of the node that wants to use the RIF Communications server
         :param grpc_api_endpoint: GRPC URI of the RIF Communications pub-sub node
         """
         self.rsk_address = RskAddress(address=rsk_address)
@@ -33,8 +32,8 @@ class RifCommsClient:
         """
         Connects to RIF Communications Node.
         Invokes ConnectToCommunicationsNode GRPC API endpoint.
-        Adds the client RSK Address under the RIF Communications node Peer ID.
-        :return: Notification stream
+        Adds the client RSK address under the RIF Communications node peer ID.
+        :return: notification stream
         """
         return self.stub.ConnectToCommunicationsNode(self.rsk_address)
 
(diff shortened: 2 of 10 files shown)


transport/rif_comms/client.py
```diff
@@ -8,95 +8,101 @@
...
+        topic_id = self._get_peer_id(rsk_address)
         return self.stub.HasSubscriber(
             Subscriber(
-                peerId=peer_id,
-                channel=Channel(channelId=rsk_address)
+                peerId=our_peer_id,
+                channel=Channel(channelId=topic_id)
             )
-        )
+        ).value
```
mortelli: i'm not sure if this is correct. i _think_ that this will check if `peerId` is subscribed to `channel`. hence, we should supply our own ID, rather than the `rsk_address` param.

transport/rif_comms/node.py
```diff
@@ -1,135 +1,228 @@
...
+        # set instance variables
         self._config = config
         self._raiden_service: RaidenService = None
 
+        # set up comms node
         self._rif_comms_connect_stream: Notification = None
-        self._our_topic: Notification = None
+        self._our_topic_stream: Notification = None
         self._our_topic_thread: Greenlet = None
+        self._comms_client = RIFCommsClient(to_checksum_address(address), self._config["grpc_endpoint"])
```
mortelli: @marcosmartinez7 do we really need to call `to_checksum_address` here?

transport/rif_comms/node.py
```diff
@@ -1,135 +1,228 @@
...
+        # start greenlet through the Runnable class; this will eventually call _run
         Runnable.start(self)
 
-    def __repr__(self):
-        node = f" node:{pex(self._raiden_service.address)}" if self._raiden_service else ""
+    def _start_message_listener(self):
+        """
+        Start a listener greenlet to listen for received messages in the background.
+        """
+        our_address = to_checksum_address(self.raiden_service.address)
```
mortelli: @marcosmartinez7 same here about the `to_checksum_address` call

transport/matrix/node.py
```diff
@@ -179,7 +178,7 @@ def on_success(greenlet):
 
         # (re)start any _RetryQueue which was initialized before start
         for retrier in self._address_to_retrier.values():
-            if not retrier:
+            if not retrier.greenlet:
```
marcosmartinez7, line 181: is this added for clarification?

transport/rif_comms/client.py
```diff
@@ -8,95 +8,101 @@
...
     def __init__(self, rsk_address: Address, grpc_api_endpoint: str):
         """
         Constructs the RIF Communications Client.
-        :param rsk_address: address of the node that wants to use the RIF Communications server
+        :param rsk_address: RSK address of the node that wants to use the RIF Communications server
         :param grpc_api_endpoint: GRPC URI of the RIF Communications pub-sub node
         """
         self.rsk_address = RskAddress(address=rsk_address)
-        self.grpc_channel = insecure_channel(grpc_api_endpoint)
+        self.grpc_channel = insecure_channel(grpc_api_endpoint)  # TODO: how to make this secure?
```
marcosmartinez7, line 29: We might take a look at this https://grpc.io/docs/guides/auth/

i have created this issue for the future https://jirainfuy.atlassian.net/browse/RLP-896

transport/rif_comms/client.py
marcosmartinez7: I think that using our_peer_id instead of retrieving the peer id from the rsk_address is ok

You can receive an rsk address that isnt "our" rsk address and try to check with HasSubscriber with a relation of peerid-address that doesnt make sense

also not sure about the channelId=topic_id (maybe it expects the rsk address here)

this must be double checked with rif comms team i guess

transport/rif_comms/client.py
```diff
@@ -8,95 +8,101 @@
...
+        :param rsk_address: the destination for the message to be sent to
         """
-
-        # TODO message encoding
-        return self.stub.SendMessageToTopic(
+        topic_id = self._get_peer_id(rsk_address)
+        # TODO: message encoding
+        self.stub.SendMessageToTopic(
             PublishPayload(
                 topic=Channel(channelId=topic_id),
```
marcosmartinez7, line 80: same here, im confussed about channelId must be an address or a peer id

i think that peer id is something that we must not know from lumino, but maybe the comms implementation need this

transport/rif_comms/client.py
```diff
@@ -8,95 +8,101 @@
...
         """
-        This unsubscribe the node from the topic.
-        Invokes the CloseTopic grpc api endpoint.
-        :param topic_id: topic identifier
-        :return: void
+        Unsubscribes from a topic which corresponds to the given RSK address.
+        Invokes the CloseTopic GRPC API endpoint.
+        :param rsk_address: RSK address which corresponds to the topic which the client is unsubscribing from.
         """
+        topic_id = self._get_peer_id(rsk_address)
```
marcosmartinez7: same

transport/rif_comms/client.py
marcosmartinez7: btw, did you check that the test_rif_comms_client works with those modifications?

transport/rif_comms/node.py
marcosmartinez7: hmm, i dont think so

transport/rif_comms/node.py
```diff
@@ -1,135 +1,228 @@
...
-        self._client.get_peer_id(
-            to_checksum_address(raiden_service.address))  # TODO remove this after grpc api request blocking is fixed
+        # subscribe to our own topic to receive messages
+        self._start_message_listener()
+
+        # start pre-loaded message queues
         for message_queue in self._address_to_message_queue.values():
-            if not message_queue:
-                self.log.debug("Starting message_queue", message_queue=message_queue)
+            if not message_queue.greenlet:
```
marcosmartinez7, line 72: i dont understand why checking the greenlet attribute

transport/rif_comms/node.py
marcosmartinez7: same, i dont think so, i think that addresses are already checksumed from when you select one on the cli

transport/rif_comms/node.py
```diff
@@ -1,135 +1,228 @@
...
+                type=type(message),
+                node=pex(self._raiden_service.address),
+                message=message,
+                sender=pex(message.sender),
+            )
 
-        return f"<{self.__class__.__name__}{node} id:{id(self)}>"
+            try:
+                # acknowledge to sender that their message was received
+                self._ack_message(message)
```
marcosmartinez7: i dont think this is ok, a Delivered message has not a response

Thats why we have `_receive_delivered_to_lc` on matrix impl

transport/rif_comms/node.py
```diff
@@ -1,135 +1,228 @@
...
+                node=pex(self._raiden_service.address),
+                message=message,
+            )
+
+    def _ack_message(self, message: (Delivered, Processed, SignedRetrieableMessage)):
+        """
+        Acknowledge a received Raiden message by sending a Delivered-type message back.
+        If the received Raiden message is of the Delivered type, no action is taken.
+        """
+        if not isinstance(message, Delivered):
```
marcosmartinez7: ok, so we call _ack_message from delivered but skip the response here..

im ok with this

marcosmartinez7 (changes requested):
Great, there are some things to fix regarding the delivered msgs but in general i think this is a great improvement

Also i would like to test the test_rif_comms_client again on this branch

transport/rif_comms/client.py
mortelli: > You can receive an rsk address that isnt "our" rsk address and try to check with HasSubscriber with a relation of peerid-address that doesnt make sense

would we ever have such a use case? as far as i'm concerned, the node only cares about what the client is subscribed to, not other peers. i might be wrong though.

> also not sure about the channelId=topic_id (maybe it expects the rsk address here)

> this must be double checked with rif comms team i guess

will get back to you on this :+1:

transport/rif_comms/node.py
mortelli: removed

transport/rif_comms/node.py
mortelli: removed

transport/matrix/node.py
mortelli: yes:

> - truthiness for `Runnable` variables in both the Matrix and RIF Comms transport nodes has been made more explicit: `if queue` → `if queue.greenlet` (`if queue` works due to the `Runnable.__bool__` function, but it's quite obscure).

transport/rif_comms/client.py
mortelli: thanks!

transport/rif_comms/client.py
mortelli: the RIF comms team has confirmed that this call will check whether `peer_id` is subscribed to `channel_id`, and that it doesn't specifically work with RSK addresses.

transport/rif_comms/client.py
mortelli: > same here, im confussed about channelId must be an address or a peer id
> 
> i think that peer id is something that we must not know from lumino, but maybe the comms implementation need this

see [previous reply](https://github.com/rsksmart/lumino/pull/161#discussion_r527727683).

> btw, did you check that the test_rif_comms_client works with those modifications?

no, on it.

transport/rif_comms/client.py
mortelli: resolving this thread so we keep the discussion in one place

transport/rif_comms/node.py
mortelli: see [here](https://github.com/rsksmart/lumino/pull/161#discussion_r527697926)

mortelli: otherwise;

```python
for message_queue in values:
    if not message_queue:
```

makes no sense

transport/rif_comms/node.py
mortelli: we could take the `if` outside of the `_ack_message` function if you prefer, i'm ok with it either way

marcosmartinez7: agree

transport/rif_comms/node.py
marcosmartinez7: lets move this to handle_message

transport/rif_comms/node.py
mortelli: done

transport/rif_comms/client.py
mortelli: @marcosmartinez7 is going to push the fixes based on the rif comms unit test results

marcosmartinez7 (approved):
Lgtm, good work!

mortelli:
based on manually running the unit tests, @marcosmartinez7 has pushed some fixes and the PR description has been updated.

PR #162, opened 2020-11-18
Feature/rlp 849/receive lc messages design

marcosmartinez7:
This PR intends to be the first Implementation of the Rif Comms `LightClientNode`

It also fixes some tests to use the current implementation of rif comms pubsub

Manual testing Steps: 
- A Full node transport and a Light client transport are initialized (addresses 0x1 and 0x2) 
- full node transport subscribes to 0x1
- light transport subscribes to 0x2
- the subscription shows the same peer id for each one (See *1)
- the sender.py script sends a message to 0x1
- the message is received sometimes by full node transport and some times by light transport 
- the sender.py script sends a message to 0x2 and the same occurs

This table represents the behaviour vaguely

|  sender calls with | first time  | second time   |  restart all  | restart lumino (subscribe again)   |
|---|---|---|---|---|
| lc  | lc   | lc  | lc   | fn  |
| fn  | lc   | lc  | lc   | fn  |

Notes that represent incorrect behaviour: 

- each message is received twice
- each message is received on the wrong topic, for example, if sender sends a message with fn address, it doesnt make sense to receive it on the lc topic

(*1)

```
createTopicWithRskAddress {"address":"0x138aF366e0ED7Cc4b9747a935D1b5F75A86b9D83","exclusive":false} 
locatePeerID "0x138aF366e0ED7Cc4b9747a935D1b5F75A86b9D83" 
connectToCommunicationsNode {"address":"0x138aF366e0ED7Cc4b9747a935D1b5F75A86b9D83","exclusive":false}
createTopicWithRskAddress {"address":"0x2Cd53d83E5ca1F3c66EDC7bf307139c6B1989C9a","exclusive":false} 
locatePeerID "0x2Cd53d83E5ca1F3c66EDC7bf307139c6B1989C9a" 
connectToCommunicationsNode {"address":"0x2Cd53d83E5ca1F3c66EDC7bf307139c6B1989C9a","exclusive":false}
Adding RSKADDRESS PEER= 16Uiu2HAm9otWzXBcFm7WC2Qufp2h1mpRxK1oox289omHTcKgrpRA  : RSKADDRESS= 0x138aF366e0ED7Cc4b9747a935D1b5F75A86b9D83
Adding RSKADDRESS PEER= 16Uiu2HAm9otWzXBcFm7WC2Qufp2h1mpRxK1oox289omHTcKgrpRA  : RSKADDRESS= 0x2Cd53d83E5ca1F3c66EDC7bf307139c6B1989C9a
address 16Uiu2HAm9otWzXBcFm7WC2Qufp2h1mpRxK1oox289omHTcKgrpRA
 - New subscription to 16Uiu2HAm9otWzXBcFm7WC2Qufp2h1mpRxK1oox289omHTcKgrpRA
address 16Uiu2HAm9otWzXBcFm7WC2Qufp2h1mpRxK1oox289omHTcKgrpRA
 - New subscription to 16Uiu2HAm9otWzXBcFm7WC2Qufp2h1mpRxK1oox289omHTcKgrpRA

```


transport/rif_comms/node.py
```diff
@@ -177,7 +177,7 @@ def _run(self, *args: Any, **kwargs: Any) -> None:
             self.log.info("RIF Comms Node _run. Listening for messages.")
         except GreenletExit:  # killed without exception
             self.stop_event.set()
-            killall(self.greenlet)  # kill children
+            killall([self.greenlet])  # kill comms listener thread
```
marcosmartinez7, line 157: this was causing an error cause greenlet is not iterable and the method killall expects a list

transport/rif_comms/node.py
```diff
@@ -204,7 +202,6 @@ def stop(self):
         if self.greenlet:
             self.greenlet.kill()
             self.greenlet.get()
-        self.greenlet = None
```
marcosmartinez7, line 207: i have removed this because on the line 210 we wait for the greenlet and it was none

i dont think the assignment make any sense since the node is shutting down anyway (and the memory is released)

transport/rif_comms/node.py
```diff
@@ -303,11 +299,38 @@ def log(self):
         return self._log
 
     def __repr__(self):
-        node = f" node:{pex(self._raiden_service.address)}" if self._raiden_service else ""
+        node = f" RIF Comms Transport node:{pex(self.address)}"
```
marcosmartinez7, line 275: i have changed this because the raiden_service.address is the hub address always and we use this method for ligth client node representation too

transport/rif_comms/node.py
```diff
@@ -303,11 +299,38 @@ def log(self):
...
 
 class LightClientNode(Node):
 
-    def __init__(self, address: Address, config: dict, auth_params: dict):
+    def __init__(self, address: Address, config: dict):
         Node.__init__(self, address, config)
+
+    def _handle_message(self, message: RaidenMessage):
+        """
+        Handle received Raiden message.
```
marcosmartinez7, line 286: this method is exactly the same as the base one, but, it does not ack messages and calls the raiden service with the is_light_client flag on True

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -26,19 +28,24 @@ def get_random_address_str() -> str:
 LUMINO_2_COMMS_API = "localhost:5016"
 LUMINO_2_ADDRESS = to_canonical_address("0xeBfF0EEe8E2b6952E589B0475e3F0E34dA0655B1")
 
+LUMINO_3_ADDRESS = "0x636BA79E46E0594ECbbEBb4F74B9336Fd4454442"
+
```
mortelli: we can structure these, like so:

```python
test_nodes = [
    {
        "address": to_canonical_address("0x8cb891510dF75C223C53f910A98c3b61B9083c3B"),
        "comms_api": "localhost:5013",
    },
    {
        "address": to_canonical_address("0xeBfF0EEe8E2b6952E589B0475e3F0E34dA0655B1"),
        "comms_api": "localhost:5016",
    },
    {
        "address": to_canonical_address("0x636BA79E46E0594ECbbEBb4F74B9336Fd4454442"),
        "comms_api": "localhost:5019",
    },
]
```

(not a blocker)

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -26,19 +28,24 @@ def get_random_address_str() -> str:
...
 UNREGISTERED_ADDRESS = get_random_address_str()
```
mortelli: i would get rid of this variable and call the `get_random_address_str` function directly on each test, so that it is different for each test in the suite.

also not a blocker

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -26,19 +28,24 @@ def get_random_address_str() -> str:
...
 
 
 @pytest.mark.usefixtures("rif_comms_client")
 @pytest.fixture(scope="class")
 def rif_comms_client(request):
-    rif_comms_client = RIFCommsClient(LUMINO_1_ADDRESS, LUMINO_1_COMMS_API)
+    rif_comms_client1 = RIFCommsClient(LUMINO_1_ADDRESS, LUMINO_1_COMMS_API)
+    rif_comms_client2 = RIFCommsClient(LUMINO_2_ADDRESS, LUMINO_2_COMMS_API)
```
mortelli: with the suggested change, this would be something like:

```python
rif_comms_client1 = RIFCommsClient(test_nodes[0]["address"], test_nodes[0]["comms_api"])
rif_comms_client2 = RIFCommsClient(test_nodes[1]["address"], test_nodes[1]["comms_api"])
```

mortelli: or

```python
node_1 = test_nodes[0]
node_2 = test_nodes[1]
```
(these could even be module variables)

```python
rif_comms_client1 = RIFCommsClient(node_1["address"], node_1["comms_api"])
rif_comms_client2 = RIFCommsClient(node_2["address"], node_2["comms_api"])
```

you could also use `.get()` for the params if you'd like

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: instead of a `list`, the outer variable could also be a `dict`, but you'd need a name for each entry (which could be numbers starting at `1`)

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -26,19 +28,24 @@ def get_random_address_str() -> str:
...
 
     def teardown():
-        rif_comms_client.disconnect()
+        rif_comms_client1.disconnect()
+        rif_comms_client2.disconnect()
 
-    request.addfinalizer(teardown)
-    request.cls.rif_comms_client = rif_comms_client
+    #request.addfinalizer(teardown)
```
mortelli: this needs to be fixed

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -53,70 +60,57 @@ class TestRiffCommsClient(unittest.TestCase):
     - Others uses both LUMINO_1_ADDRESS and LUMINO_2_ADDRESS to represent two separeted peers
     - ALl the tests assumes that there is a RIF COMMS node already running
     """
-
     @pytest.mark.skip(reason="ignore")
     def test_initialization(self):
-        assert self.rif_comms_client is not None
+        response = self.rif_comms_client1.connect()
+        assert self.rif_comms_client1 is not None
+        time.sleep(5)
```
mortelli: do we still need this `sleep` call? why?

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -53,70 +60,57 @@ class TestRiffCommsClient(unittest.TestCase):
...
         assert peer_id is not None
 
     @pytest.mark.skip(reason="ignore")
-    def test_locate_unregistered_peer_id(self):
-        # TODO: this should be fixed with the next RIF Comms pub-sub release
-        self.assertRaises(_InactiveRpcError, lambda: self.rif_comms_client._get_peer_id(UNREGISTERED_ADDRESS))
+    def test_locate_other_peer_id(self):
+        response = self.rif_comms_client1.connect()
+        peer_id = self.rif_comms_client1._get_peer_id(LUMINO_3_ADDRESS)
+        print(f"test_locate_peer_id peer_id = {peer_id}")
```
mortelli: this should be removed before merging

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -53,70 +60,57 @@ class TestRiffCommsClient(unittest.TestCase):
     - Others uses both LUMINO_1_ADDRESS and LUMINO_2_ADDRESS to represent two separeted peers
     - ALl the tests assumes that there is a RIF COMMS node already running
     """
-
     @pytest.mark.skip(reason="ignore")
     def test_initialization(self):
```
mortelli:
```suggestion
    def test_connect(self):
```

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -53,70 +60,57 @@ class TestRiffCommsClient(unittest.TestCase):
...
-        assert self.rif_comms_client is not None
+        response = self.rif_comms_client1.connect()
+        assert self.rif_comms_client1 is not None
+        time.sleep(5)
 
     @pytest.mark.skip(reason="ignore")
     def test_locate_peer_id(self):
```
mortelli:
```suggestion
    def test_locate_own_peer_id(self):
```

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -53,70 +60,57 @@ class TestRiffCommsClient(unittest.TestCase):
...
+        response = self.rif_comms_client1.connect()
+        peer_id = self.rif_comms_client1._get_peer_id(LUMINO_1_ADDRESS)
         print(f"test_locate_peer_id peer_id = {peer_id}")
         assert peer_id is not None
 
     @pytest.mark.skip(reason="ignore")
-    def test_locate_unregistered_peer_id(self):
-        # TODO: this should be fixed with the next RIF Comms pub-sub release
-        self.assertRaises(_InactiveRpcError, lambda: self.rif_comms_client._get_peer_id(UNREGISTERED_ADDRESS))
+    def test_locate_other_peer_id(self):
```
mortelli:
```suggestion
    def test_locate_unregistered_peer_id(self):
```

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -53,70 +60,57 @@ class TestRiffCommsClient(unittest.TestCase):
...
+        response = self.rif_comms_client1.connect()
+        peer_id = self.rif_comms_client1._get_peer_id(LUMINO_3_ADDRESS)
+        print(f"test_locate_peer_id peer_id = {peer_id}")
+        assert peer_id is ""
 
-    @pytest.mark.skip(reason="ignore")
+    @pytest.mark.skipif(reason="rif comms issue infinite loop")
```
mortelli:
```suggestion
    @pytest.mark.skip(reason="ignore")
```

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -53,70 +60,57 @@ class TestRiffCommsClient(unittest.TestCase):
...
     def test_create_random_topic_id_without_connection(self):
-        # TODO: this should be fixed with the next RIF Comms pub-sub release
-        notification = self.rif_comms_client.connect()
-        peer_id = self.rif_comms_client._get_peer_id(LUMINO_1_ADDRESS)
-        channel = self.rif_comms_client.subscribe_to(get_random_address_str())
-        peer_id = self.rif_comms_client._get_peer_id(LUMINO_1_ADDRESS)
+        notification = self.rif_comms_client1.connect()
+        channel = self.rif_comms_client1.subscribe_to(get_random_address_str())
+        peer_id = self.rif_comms_client1._get_peer_id(LUMINO_3_ADDRESS)
```
mortelli: why are we subscribing to a random address and then finding the peer id for test node # 3?

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -53,70 +60,57 @@ class TestRiffCommsClient(unittest.TestCase):
...
-        self.assertRaises(_InactiveRpcError, lambda: self.rif_comms_client._get_peer_id(UNREGISTERED_ADDRESS))
+    def test_locate_other_peer_id(self):
+        response = self.rif_comms_client1.connect()
+        peer_id = self.rif_comms_client1._get_peer_id(LUMINO_3_ADDRESS)
+        print(f"test_locate_peer_id peer_id = {peer_id}")
+        assert peer_id is ""
 
-    @pytest.mark.skip(reason="ignore")
+    @pytest.mark.skipif(reason="rif comms issue infinite loop")
     def test_create_random_topic_id_without_connection(self):
```
mortelli: i'll suggest a better name for this one once i understand the function body a bit better

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -53,70 +60,57 @@ class TestRiffCommsClient(unittest.TestCase):
...
-        # TODO: this is failing with "peer is not subscribed to"
-        channel = grpc.insecure_channel(LUMINO_1_COMMS_API)
-        stub = CommunicationsApiStub(channel)
-        rsk_address = RskAddress(address=LUMINO_1_ADDRESS)
-        notification = stub.ConnectToCommunicationsNode(rsk_address)
-        channel = stub.CreateTopicWithRskAddress(rsk_address)
-        subscribers = stub.GetSubscribers(Channel(channelId=LUMINO_1_ADDRESS))
 
-    @pytest.mark.skip(reason="ignore")
+    @pytest.mark.skip(reason="rif comms issue infinite loop")
```
mortelli:
```suggestion
    @pytest.mark.skip(reason="ignore")
```

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -53,70 +60,57 @@ class TestRiffCommsClient(unittest.TestCase):
...
-            Subscriber(
-                peerId=peer_id.address,
-                channel=Channel(channelId=LUMINO_1_ADDRESS)
-            )
-        )
-        print("has_subscriber", has_subscriber)
+        notification = self.rif_comms_client1.connect()
+        notification2 = self.rif_comms_client2.connect()
+        channel = self.rif_comms_client1.subscribe_to(LUMINO_2_ADDRESS)
+        time.sleep(10)
```
mortelli: if the comms node blocks properly now, we need to remember to remove this one

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -53,70 +60,57 @@ class TestRiffCommsClient(unittest.TestCase):
...
+        subscribed = self.rif_comms_client1.is_subscribed_to(LUMINO_2_ADDRESS)
+        print("Subscribed: ", subscribed)
```
mortelli:
```suggestion
        assert subscribed is True
```

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -53,70 +60,57 @@ class TestRiffCommsClient(unittest.TestCase):
...
+
+
+
+    @pytest.mark.skip(reason="works but subscribed equals False")
+    def test_has_subscriber_self(self):
+        notification = self.rif_comms_client1.connect()
+        channel = self.rif_comms_client1.subscribe_to(LUMINO_1_ADDRESS)
+        subscribed = self.rif_comms_client1.is_subscribed_to(LUMINO_1_ADDRESS)
+        print("Subscribed: ", subscribed)
```
mortelli:
```suggestion
        assert subscribed is True
```

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -53,70 +60,57 @@ class TestRiffCommsClient(unittest.TestCase):
...
 
     @pytest.mark.skip(reason="ignore")
     def test_disconnect(self):
         notification = self.rif_comms_client.connect()
         peer_id = self.rif_comms_client._get_peer_id(LUMINO_1_ADDRESS)
         self.rif_comms_client.disconnect()
+        # TODO check if end comms deletes topics
```
mortelli, line 119: let's create an issue for this if needed and inexistent

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -53,70 +60,57 @@ class TestRiffCommsClient(unittest.TestCase):
...
+
+    @pytest.mark.skip(reason="works but subscribed equals False")
+    def test_has_subscriber_self(self):
+        notification = self.rif_comms_client1.connect()
+        channel = self.rif_comms_client1.subscribe_to(LUMINO_1_ADDRESS)
+        subscribed = self.rif_comms_client1.is_subscribed_to(LUMINO_1_ADDRESS)
+        print("Subscribed: ", subscribed)
 
     @pytest.mark.skip(reason="ignore")
     def test_disconnect(self):
```
mortelli, line 114: not sure yet what we expect to happen after `disconnect` is called :thinking: 

we have to think about what the point of this test is

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -53,70 +60,57 @@ class TestRiffCommsClient(unittest.TestCase):
...
         notification = self.rif_comms_client.connect()
         peer_id = self.rif_comms_client._get_peer_id(LUMINO_1_ADDRESS)
```
mortelli: we should verify whether this one can be removed

transport/rif_comms/node.py
mortelli: whoops, my bad. thanks!

transport/rif_comms/node.py
```diff
@@ -186,11 +185,9 @@ def _run(self, *args: Any, **kwargs: Any) -> None:
     def stop(self):
         """
         Try to gracefully stop the underlying greenlet synchronously.
-
         Stop isn't expected to re-raise greenlet _run exception
         (use self.greenlet.get() for that),
         but it should raise any stop-time exception.
-
```
mortelli, line 193: are you not a fan of new lines in docstrings? :stuck_out_tongue:

transport/rif_comms/node.py
mortelli: i'm not sure where we got this from. i'm positive it was copied over from somewhere else, but i can't seem to find it anymore :man_shrugging:

transport/rif_comms/node.py
mortelli: coolness :+1:

transport/rif_comms/node.py
```diff
@@ -303,11 +298,44 @@ def log(self):
...
+                node=pex(self.address),
+                message=message,
+            )
+
+    def _ack_message(self, message: (Processed, SignedRetrieableMessage)):
+        """
+        Acks must be signed by the Light client first, therefore the transport is not in charge to create and send
+        the Delivered messages.
+        """
+        raise Exception("Do not use _ack_message for light client transport")
```
mortelli, line 315: this is nice and makes sense, but i am assuming the signed `Delivered` message will be sent somewhere after the  `self._raiden_service.on_message` call on line `327`?

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -159,3 +153,36 @@ def test_send_lumino_message(self):
...
+
+        notification_stream_2 = self.rif_comms_client2.connect()
+
+        one_own_sub = self.rif_comms_client1.subscribe_to(LUMINO_1_ADDRESS)
+        one_two_sub = self.rif_comms_client1.subscribe_to(LUMINO_2_ADDRESS)
+
+        two_own_sub = self.rif_comms_client2.subscribe_to(LUMINO_2_ADDRESS)
+        two_one_sub = self.rif_comms_client2.subscribe_to(LUMINO_1_ADDRESS)
+
+        peer_id_1 = self.rif_comms_client2._get_peer_id(LUMINO_2_ADDRESS)
```
mortelli: this seems unfinished; maybe we should replace this `_get_peer_id` call (once the code no longer hangs without it) and check for both subscriptions on each peer instead

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -159,3 +153,36 @@ def test_send_lumino_message(self):
...
+    def test_two_clients_listen_same_topic(self):
+
+        notification_stream_1 = self.rif_comms_client1.connect()
+
+        notification_stream_2 = self.rif_comms_client2.connect()
+
+        one_own_sub = self.rif_comms_client1.subscribe_to(LUMINO_1_ADDRESS)
+        two_one_sub = self.rif_comms_client2.subscribe_to(LUMINO_1_ADDRESS)
+
+        peer_id_1 = self.rif_comms_client2._get_peer_id(LUMINO_2_ADDRESS)
```
mortelli: let's remember to remove this once the code no longer hangs without it

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -159,3 +153,36 @@ def test_send_lumino_message(self):
...
+
+        for resp in one_own_sub:
+            print("Respone for 1: ", resp)
+
+        for resp in two_one_sub:
+            print("Respone for 2: ", resp)
```
mortelli: this will hang indefinitely.

with the current RIF Comms implementation, we are expecting an initial response which is the result of the subscription call (this contains no data). after it, we can expect messages to be received in this `for` loop.

what we can do here is count the amount of messages we expect to receive, and `break` out of this loop once we reach that amount.

in this case there's 0 data messages; there's just 1 message for the subscription result. if this works, i would extend this test to send 1 message to the topic, and exit each loop after 2 iterations.

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -53,70 +60,57 @@ class TestRiffCommsClient(unittest.TestCase):
...
 
     @pytest.mark.skip(reason="ignore")
     def test_disconnect(self):
         notification = self.rif_comms_client.connect()
         peer_id = self.rif_comms_client._get_peer_id(LUMINO_1_ADDRESS)
         self.rif_comms_client.disconnect()
+        # TODO check if end comms deletes topics
 
     @pytest.mark.skip(reason="ignore")
     def test_send_lumino_message(self):
```
mortelli: would you agree on not having a test which requires an entire lumino node to be fired up?

if so, i'll create an issue to mock or somehow reduce this

mortelli (commented):
superb work :clap: 

i've left some comments, but i see no major issue. 

the whole thing isn't functional yet, but it doesn't need to be for this PR to be merged (it's also getting there!)

some more thoughts:
- we have to start thinking about how to have these RIF Comms unit/integration tests executed through `pytest` without having to run a RIF Comms node manually (lmk if you agree on creating an issue for this and i'll document it). otherwise we will have to run and verify them manually each time, which almost defeats the purpose of unit tests
- we have a new upcoming version for the comms node. we can either merge this PR first, then branch out and make changes based on the new comms code, _or_ verify this new code directly on this branch/PR. up to you, both are acceptable afaic
- once we know that the registration for multiple RSK addresses on the same peer ID is—in theory—functional, we'll have our work cut out for us 
- it'd be nice to have the `teardown` code working :nerd_face: hmu if you need help with this one

marcosmartinez7:
> superb work 👏
> 
> i've left some comments, but i see no major issue.
> 
> the whole thing isn't functional yet, but it doesn't need to be for this PR to be merged (it's also getting there!)
> 
> some more thoughts:
> 
> * we have to start thinking about how to have these RIF Comms unit/integration tests executed through `pytest` without having to run a RIF Comms node manually (lmk if you agree on creating an issue for this and i'll document it). otherwise we will have to run and verify them manually each time, which almost defeats the purpose of unit tests
> * we have a new upcoming version for the comms node. we can either merge this PR first, then branch out and make changes based on the new comms code, _or_ verify this new code directly on this branch/PR. up to you, both are acceptable afaic
> * once we know that the registration for multiple RSK addresses on the same peer ID is—in theory—functional, we'll have our work cut out for us
> * it'd be nice to have the `teardown` code working 🤓 hmu if you need help with this one

Thanks for the review

- I think is better to iterate this PR using the new version of rif comms node and fix something if needed before merge. 
- Lets improve the test cases too and try to make them automated
- I will take a look to the teardown issue

raiden/tests/unit/rif_comms/test_rif_comms_client.py
marcosmartinez7: i think we have time now to improve this

raiden/tests/unit/rif_comms/test_rif_comms_client.py
marcosmartinez7: lets do this too

raiden/tests/unit/rif_comms/test_rif_comms_client.py
marcosmartinez7: lets do this too

raiden/tests/unit/rif_comms/test_rif_comms_client.py
marcosmartinez7: Lets take a look at this on this iteration

raiden/tests/unit/rif_comms/test_rif_comms_client.py
marcosmartinez7: it was a workarround for the teardown of disconnect, but it didnt work

raiden/tests/unit/rif_comms/test_rif_comms_client.py
marcosmartinez7: this test must fail, therefore it must be changed to an assertion of exception raising

raiden/tests/unit/rif_comms/test_rif_comms_client.py
marcosmartinez7: yeah this doesnt make sense, i will remove this test

raiden/tests/unit/rif_comms/test_rif_comms_client.py
marcosmartinez7: agree, lets test that

raiden/tests/unit/rif_comms/test_rif_comms_client.py
marcosmartinez7: https://jirainfuy.atlassian.net/browse/RLP-909

raiden/tests/unit/rif_comms/test_rif_comms_client.py
marcosmartinez7: it must not fail, but, we have to check the subscriptions are discarded

transport/rif_comms/node.py
marcosmartinez7: Exactly, the LCSDK has the responsability of sending the acks

raiden/tests/unit/rif_comms/test_rif_comms_client.py
marcosmartinez7: yep, this test was for checking that two nodes can subscribe to each other topic and receive messages, but we need to iterate over this

right now the test that ensures this works is to run two lumino nodes and send messages between each other, but is not something automatic

raiden/tests/unit/rif_comms/test_rif_comms_client.py
marcosmartinez7: agree with the count approach for now

raiden/tests/unit/rif_comms/test_rif_comms_client.py
marcosmartinez7: the purpose was to test that a running node of lumino can receive and parse correctly a message sent to a rif comms node

i agree we need to mock this or create an integration test that really runs a lumino node, but for now i fund this test useful

transport/rif_comms/node.py
marcosmartinez7: this was on the matrix client

mortelli: > this was on the matrix client

you're right, but it doesn't really add up then.

the matrix node, when stopped, calls a function for its matrix client, which kills its thread.

shouldn't the rif comms node do something analogous?

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: we can leave it so we have:
1. test locate invalid RSK address (although maybe we don't need this one with the latest comms version)
2. test subscribe to invalid RSK address

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: thanks

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: https://jirainfuy.atlassian.net/browse/RLP-910

transport/rif_comms/node.py
mortelli: @marcosmartinez7 can you confirm whether this is happening or not?

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: done through #166

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: completed through #166

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: completed through #166

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: removed through #166

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: still pending

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: resolved through #166

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: resolved through #166

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: number 2 is still pending

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: removed through #166

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: removed through #166

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: removed through #166

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: resolved through #166

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: removed through #166

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: resolved through #166

transport/rif_comms/node.py
mortelli: the client is not running on a thread so we don't actually need this, resolving

mortelli (approved):

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: [RLP-937](https://jirainfuy.atlassian.net/browse/RLP-937)

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: removed in #170

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: done through #170

PR #164, opened 2020-11-30
Bugfix/rlp 802/register secret lc

mauricioirace:
## What does this commit solves?
On chain unlock in mediated payments was not working in a Taringa-like architecture. 

## How does it solves it?

If applied, this commit will:
- **Exchange** closing/non-closing participant addresses in the channel closed event handler in order to send the right balance proof
- **Remove** channel restoration when unlocking, to prevent wrong participant from unlocking locked transfers
- **Check** if the light client will have any gain unlocking, instead of always unlocking 
- **Add** light client handler for on chain secret reveal, as the gain of a participant is not the same if a secret is registered or not
- **Add** the balance proof as a property to SendLockedTransferLight class to prevent a crash when settling a channel from a initiator. hwn the payment was not completed off chain

### DIFF
--- a/raiden/api/webui/static/endpointConfig.js
+++ b/raiden/api/webui/static/endpointConfig.js
@@ -1,5 +1,5 @@
 const backendUrl='http://localhost:5002'; 
-const nodeAddress = '0x2Cd53d83E5ca1F3c66EDC7bf307139c6B1989C9a'; 
+const nodeAddress = '0x6dd4d8005E5D2DEA818ae4A8d57c8FA1B5c492C8'; 
 const rnsDomain = null 
 const chainEndpoint = 'http://localhost:4444'; 
 
--- a/raiden/api/webui/static/endpointConfig.js
+++ b/raiden/api/webui/static/endpointConfig.js
@@ -1,7 +1,7 @@
-const backendUrl='http://localhost:5002';
-const nodeAddress = '0x62Aa1173561C4E92876B3D073B9767A2d07496f8';
-const rnsDomain = null;
-const chainEndpoint = 'http://localhost:4444';
+const backendUrl='http://localhost:5002'; 
+const nodeAddress = '0x2Cd53d83E5ca1F3c66EDC7bf307139c6B1989C9a'; 
+const rnsDomain = null 
+const chainEndpoint = 'http://localhost:4444'; 
 
 window.luminoUrl = backendUrl;
 window.nodeAddress= nodeAddress;
--- a/raiden/api/webui/static/endpointConfig.js
+++ b/raiden/api/webui/static/endpointConfig.js
@@ -1,7 +1,7 @@
-const backendUrl='http://localhost:5002'; 
-const nodeAddress = '0x6dd4d8005E5D2DEA818ae4A8d57c8FA1B5c492C8'; 
-const rnsDomain = null 
-const chainEndpoint = 'http://localhost:4444'; 
+const backendUrl='http://localhost:5002';
+const nodeAddress = '0x62Aa1173561C4E92876B3D073B9767A2d07496f8';
+const rnsDomain = null;
+const chainEndpoint = 'http://localhost:4444';
 
 window.luminoUrl = backendUrl;
 window.nodeAddress= nodeAddress;
--- a/raiden/blockchain_events_handler.py
+++ b/raiden/blockchain_events_handler.py
@@ -33,7 +33,7 @@ from raiden.transfer.state_change import (
     ContractReceiveSecretReveal,
     ContractReceiveUpdateTransfer,
     ContractReceiveChannelClosedLight,
-    ContractReceiveChannelSettledLight
+    ContractReceiveChannelSettledLight, ContractReceiveSecretRevealLight
 )
 from raiden.transfer.utils import (
     get_event_with_balance_proof_by_locksroot,
@@ -588,16 +588,32 @@ def handle_secret_revealed(raiden: "RaidenService", event: Event):
     block_number = data["block_number"]
     block_hash = data["block_hash"]
     transaction_hash = data["transaction_hash"]
-    registeredsecret_state_change = ContractReceiveSecretReveal(
-        transaction_hash=transaction_hash,
-        secret_registry_address=secret_registry_address,
-        secrethash=args["secrethash"],
-        secret=args["secret"],
-        block_number=block_number,
-        block_hash=block_hash,
-    )
-
-    raiden.handle_and_track_state_change(registeredsecret_state_change)
+    secret=args["secret"]
+    secrethash=args["secrethash"]
+    chain_state = views.state_from_raiden(raiden)
+    if chain_state.get_payment_task(raiden.address, secrethash):
+        registeredsecret_state_change = ContractReceiveSecretReveal(
+            transaction_hash=transaction_hash,
+            secret_registry_address=secret_registry_address,
+            secrethash=secrethash,
+            secret=secret,
+            block_number=block_number,
+            block_hash=block_hash,
+        )
+        raiden.handle_and_track_state_change(registeredsecret_state_change)
+    else:
+        for lc_address, payment_state in chain_state.get_payment_states_by_address():
+            if secrethash in set(payment_state.secrethashes_to_task):
+                registeredsecret_state_change = ContractReceiveSecretRevealLight(
+                    transaction_hash=transaction_hash,
+                    secret_registry_address=secret_registry_address,
+                    secrethash=secrethash,
+                    secret=secret,
+                    block_number=block_number,
+                    block_hash=block_hash,
+                    lc_address=lc_address
+                )
+                raiden.handle_and_track_state_change(registeredsecret_state_change)
 
 
 def on_blockchain_event(raiden: "RaidenService", event: Event):
--- a/raiden/blockchain_events_handler.py
+++ b/raiden/blockchain_events_handler.py
@@ -326,17 +326,20 @@ def handle_channel_closed(raiden: "RaidenService", event: Event):
             raiden.handle_and_track_state_change(channel_closed)
         else:
             # Must be a light client
+            closing_participant = args["closing_participant"]
+            non_closing_participant = channel_state.partner_state.address \
+                if channel_state.our_state.address == closing_participant \
+                else channel_state.our_state.address
             latest_non_closing_balance_proof = LightClientMessageHandler.get_latest_light_client_non_closing_balance_proof(
-                channel_state.identifier, channel_state.our_state.address, raiden.wal.storage
+                channel_state.identifier, non_closing_participant, raiden.wal.storage
             )
             channel_closed = ContractReceiveChannelClosedLight(
                 transaction_hash=transaction_hash,
-                transaction_from=args["closing_participant"],
+                transaction_from=closing_participant,
                 canonical_identifier=channel_state.canonical_identifier,
                 block_number=block_number,
                 block_hash=block_hash,
-                closing_participant=channel_state.partner_state.address,
-                non_closing_participant=channel_state.our_state.address,
+                non_closing_participant=non_closing_participant,
                 latest_update_non_closing_balance_proof_data=latest_non_closing_balance_proof
             )
             raiden.handle_and_track_state_change(channel_closed)
--- a/raiden/blockchain_events_handler.py
+++ b/raiden/blockchain_events_handler.py
@@ -327,7 +327,7 @@ def handle_channel_closed(raiden: "RaidenService", event: Event):
         else:
             # Must be a light client
             latest_non_closing_balance_proof = LightClientMessageHandler.get_latest_light_client_non_closing_balance_proof(
-                channel_state.identifier, channel_state.partner_state.address, raiden.wal.storage
+                channel_state.identifier, channel_state.our_state.address, raiden.wal.storage
             )
             channel_closed = ContractReceiveChannelClosedLight(
                 transaction_hash=transaction_hash,
@@ -335,8 +335,8 @@ def handle_channel_closed(raiden: "RaidenService", event: Event):
                 canonical_identifier=channel_state.canonical_identifier,
                 block_number=block_number,
                 block_hash=block_hash,
-                closing_participant=channel_state.our_state.address,
-                non_closing_participant=channel_state.partner_state.address,
+                closing_participant=channel_state.partner_state.address,
+                non_closing_participant=channel_state.our_state.address,
                 latest_update_non_closing_balance_proof_data=latest_non_closing_balance_proof
             )
             raiden.handle_and_track_state_change(channel_closed)
--- a/raiden/blockchain_events_handler.py
+++ b/raiden/blockchain_events_handler.py
@@ -33,7 +33,8 @@ from raiden.transfer.state_change import (
     ContractReceiveSecretReveal,
     ContractReceiveUpdateTransfer,
     ContractReceiveChannelClosedLight,
-    ContractReceiveChannelSettledLight, ContractReceiveSecretRevealLight
+    ContractReceiveChannelSettledLight,
+    ContractReceiveSecretRevealLight
 )
 from raiden.transfer.utils import (
     get_event_with_balance_proof_by_locksroot,
--- a/raiden/transfer/mediated_transfer/events.py
+++ b/raiden/transfer/mediated_transfer/events.py
@@ -2,9 +2,10 @@
 from eth_utils import to_canonical_address, to_checksum_address
 
 from raiden.lightclient.models.light_client_protocol_message import LightClientProtocolMessageType
-from raiden.messages import RevealSecret, Unlock, Message, SecretRequest, LockExpired
+from raiden.messages import RevealSecret, Unlock, Message, SecretRequest, LockExpired, LockedTransfer
 
 from raiden.transfer.architecture import Event, SendMessageEvent
+from raiden.transfer.identifiers import CanonicalIdentifier
 from raiden.transfer.mediated_transfer.state import LockedTransferUnsignedState
 from raiden.transfer.state import BalanceProofUnsignedState
 from raiden.utils import pex, sha3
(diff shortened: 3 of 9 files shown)


raiden/api/webui/static/endpointConfig.js
```diff
@@ -1,7 +1,7 @@
-const backendUrl='http://localhost:5002';
-const nodeAddress = '0x62Aa1173561C4E92876B3D073B9767A2d07496f8';
-const rnsDomain = null;
-const chainEndpoint = 'http://localhost:4444';
+const backendUrl='http://localhost:5002'; 
+const nodeAddress = '0x6dd4d8005E5D2DEA818ae4A8d57c8FA1B5c492C8'; 
+const rnsDomain = null 
+const chainEndpoint = 'http://localhost:4444';
```
mauricioirace: Wasn't this ignored?

raiden/blockchain_events_handler.py
```diff
@@ -588,16 +588,32 @@ def handle_secret_revealed(raiden: "RaidenService", event: Event):
...
+                registeredsecret_state_change = ContractReceiveSecretRevealLight(
+                    transaction_hash=transaction_hash,
+                    secret_registry_address=secret_registry_address,
+                    secrethash=secrethash,
+                    secret=secret,
+                    block_number=block_number,
+                    block_hash=block_hash,
+                    lc_address=lc_address
+                )
+                raiden.handle_and_track_state_change(registeredsecret_state_change)
```
mauricioirace, line 620: Here we are checking if the node is a hub, and sending the event to the light clients if so. I don't really like the _for_, but can't think right now  a better way

raiden/blockchain_events_handler.py
```diff
@@ -33,7 +33,7 @@
     ContractReceiveSecretReveal,
     ContractReceiveUpdateTransfer,
     ContractReceiveChannelClosedLight,
-    ContractReceiveChannelSettledLight
+    ContractReceiveChannelSettledLight, ContractReceiveSecretRevealLight
```
mortelli: nl here

raiden/blockchain_events_handler.py
```diff
@@ -327,16 +327,16 @@ def handle_channel_closed(raiden: "RaidenService", event: Event):
         else:
             # Must be a light client
             latest_non_closing_balance_proof = LightClientMessageHandler.get_latest_light_client_non_closing_balance_proof(
-                channel_state.identifier, channel_state.partner_state.address, raiden.wal.storage
+                channel_state.identifier, channel_state.our_state.address, raiden.wal.storage
```
marcosmartinez7: im a little bit concern about how this was working before for LC - FN scenarios

i mean, this change could break the different topologies? 

- LC - FN
- LC - FN - LC (same hub)
- LC - FN -LC (different hub)

raiden/blockchain_events_handler.py
```diff
@@ -327,16 +327,16 @@ def handle_channel_closed(raiden: "RaidenService", event: Event):
...
             )
             channel_closed = ContractReceiveChannelClosedLight(
                 transaction_hash=transaction_hash,
                 transaction_from=args["closing_participant"],
                 canonical_identifier=channel_state.canonical_identifier,
                 block_number=block_number,
                 block_hash=block_hash,
-                closing_participant=channel_state.our_state.address,
-                non_closing_participant=channel_state.partner_state.address,
+                closing_participant=channel_state.partner_state.address,
```
marcosmartinez7: same here

raiden/blockchain_events_handler.py
marcosmartinez7: why not accessing to chain_state.get_payment_states_by_address()[lc_address] here?

youre iterating over all the payment tasks right? even the tasks that are not related with the lc

mauricioirace: The problem with that is we don't know which lc is related to the payment task. Our only input is the payment's secrethash

raiden/api/webui/static/endpointConfig.js
mauricioirace: (I forgot this, again)

raiden/transfer/mediated_transfer/initiator.py
```diff
@@ -4,7 +4,7 @@
 
 from raiden.constants import MAXIMUM_PENDING_TRANSFERS
 from raiden.lightclient.models.light_client_protocol_message import LightClientProtocolMessageType
-from raiden.messages import LockedTransfer, Unlock
+from raiden.messages import LockedTransfer, Unlock, RevealSecret
```
mortelli: unused import here

raiden/blockchain_events_handler.py
marcosmartinez7: got it, lets keep it like this then

what do you guys think? @mortelli @JONAF2103

raiden/blockchain_events_handler.py
```diff
@@ -326,17 +326,20 @@ def handle_channel_closed(raiden: "RaidenService", event: Event):
             raiden.handle_and_track_state_change(channel_closed)
         else:
             # Must be a light client
```
marcosmartinez7, line 329: could you explain why did you put this docstring "Must be a light client"? 

This handler function is both for managing a closed channel for a full node and a light client

raiden/blockchain_events_handler.py
```diff
@@ -326,17 +326,20 @@ def handle_channel_closed(raiden: "RaidenService", event: Event):
...
+            closing_participant = args["closing_participant"]
+            non_closing_participant = channel_state.partner_state.address \
+                if channel_state.our_state.address == closing_participant \
+                else channel_state.our_state.address
             latest_non_closing_balance_proof = LightClientMessageHandler.get_latest_light_client_non_closing_balance_proof(
-                channel_state.identifier, channel_state.partner_state.address, raiden.wal.storage
+                channel_state.identifier, non_closing_participant, raiden.wal.storage
```
marcosmartinez7, line 335: previously this was channel_state.our_state.address

then we used channel_state.partner_state.address 

did you find wich scenarios (toppologies) were causing an issue when using our state?

i mean, i found it ok now, but we should have clearly identified the wrong behaviour when making a critical change like this

what do you guys think? @mortelli @JONAF2103

raiden/blockchain_events_handler.py
mauricioirace: if the nodes address (raiden.address) is not the same as the chanel_state.our_state.address, then, necessarilly, the node is a light client

raiden/blockchain_events_handler.py
mauricioirace: channel_state.our_state.address worked fine with lc-lc same hub, but failed in lc-fn-lc.

channel_state.partner_state.address worked well in any situation, but just because some checks that were done in the ContractReceiveChannelClosedLight handler, which prevented the wrong participant from sending the non closis bp.

I changed it to the current code in order to make it cleaner, less confussing, and much less error prone in the future.

mortelli: > previously this was channel_state.our_state.address
>
> then we used channel_state.partner_state.address 

those 2 options are alternatives for the `if` on line `331`.

seems OK to me as long as it appears to work

JONAF2103: My question is why handle_channel_closed is being triggered by both participants, and if that's the correct behaviour then what should be latest_non_closing_balance_proof? if that is the balance proof for the non closing participant then the new logic make sense to me, since you always will need the address for the non closing participant, if not then i miss understood the purpose of this logic.

mauricioirace: I guess the closing participant needs to get notified that the transaction it sent was successful, in order to update update the channel state accordingly.

marcosmartinez7: yeah basically both sides must update the channel status

but the hub must send the latest non closing balance proof for the non closing participant (acting as a watchtower) in order to prevent the closer to get all the funds

raiden/blockchain_events_handler.py
mortelli: sorry, i don't understand enough to have an opinion here (regarding the lc/addresses stuff).

i also don't like the `for`, but are there any alternatives that don't involve a significant redesign?

not worth it atm if you ask me

raiden/transfer/state_change.py
```diff
@@ -1083,6 +1085,83 @@ def from_dict(cls, data: Dict[str, Any]) -> "ContractReceiveSecretReveal":
...
+        block_number: BlockNumber,
+        block_hash: BlockHash,
+        lc_address: Address
+    ) -> None:
+        if not isinstance(secret_registry_address, T_SecretRegistryAddress):
+            raise ValueError("secret_registry_address must be of type SecretRegistryAddress")
+        if not isinstance(secrethash, T_SecretHash):
+            raise ValueError("secrethash must be of type SecretHash")
+        if not isinstance(secret, T_Secret):
+            raise ValueError("secret must be of type Secret")
```
mortelli, line 1106: why are these compared to `T_Type` instead of `Type`?

raiden/transfer/node.py
```diff
@@ -762,11 +764,17 @@ def handle_secret_reveal(
 
 
 def handle_contract_secret_reveal(
-    chain_state: ChainState, state_change: ContractReceiveSecretReveal
+    chain_state: ChainState, state_change: ContractReceiveSecretReveal, storage=None
```
mortelli: we don't need this change, do we?

raiden/transfer/node.py
```diff
@@ -762,11 +764,17 @@ def handle_secret_reveal(
...
 ) -> TransitionResult[ChainState]:
     return subdispatch_to_paymenttask(chain_state, state_change, chain_state.our_address, state_change.secrethash)
 
 
+def handle_contract_secret_reveal_light(
+    chain_state: ChainState, state_change: ContractReceiveSecretRevealLight, storage=None
```
mortelli: `storage` also does not seem to be used here (please correct me if i am wrong)

raiden/transfer/mediated_transfer/target.py
```diff
@@ -756,6 +757,11 @@ def state_transition(
         msg = "ContractReceiveSecretReveal should be accompanied by a valid target state"
         assert target_state, msg
         iteration = handle_onchain_secretreveal(target_state, state_change, channel_state)
+    elif type(state_change) == ContractReceiveSecretRevealLight:
+        assert isinstance(state_change, ContractReceiveSecretRevealLight), MYPY_ANNOTATION
+        msg = "ContractReceiveSecretReveal should be accompanied by a valid target state"
+        assert target_state, msg
```
mortelli:
```suggestion
        assert target_state, "ContractReceiveSecretReveal should be accompanied by a valid target state"
```

raiden/transfer/mediated_transfer/target.py
```diff
@@ -756,6 +757,11 @@ def state_transition(
         msg = "ContractReceiveSecretReveal should be accompanied by a valid target state"
         assert target_state, msg
         iteration = handle_onchain_secretreveal(target_state, state_change, channel_state)
+    elif type(state_change) == ContractReceiveSecretRevealLight:
+        assert isinstance(state_change, ContractReceiveSecretRevealLight), MYPY_ANNOTATION
```
mortelli: how could this `assert` fail after the `elif` on line `760` is `True`?

raiden/transfer/mediated_transfer/initiator_manager.py
```diff
@@ -38,7 +39,7 @@
     Optional,
     SecretHash,
     TokenNetworkID,
-    cast,
+    cast, Union,
```
mortelli: nl here

raiden/transfer/mediated_transfer/initiator_manager.py
```diff
@@ -752,6 +753,26 @@ def state_transition(
...
+    elif type(state_change) == ContractReceiveSecretReveal:
+        assert isinstance(state_change, ContractReceiveSecretReveal), MYPY_ANNOTATION
+        msg = "ContractReceiveSecretReveal should be accompanied by a valid payment state"
+        assert payment_state, msg
+        iteration = handle_onchain_secretreveal(
+            payment_state=payment_state,
+            state_change=state_change,
+            channelidentifiers_to_channels=channelidentifiers_to_channels,
+            pseudo_random_generator=pseudo_random_generator,
+        )
```
mortelli: isn't this entire block of code duplicated?

raiden/transfer/mediated_transfer/initiator_manager.py
```diff
@@ -752,6 +753,26 @@ def state_transition(
...
+    elif type(state_change) == ContractReceiveSecretRevealLight:
+        assert isinstance(state_change, ContractReceiveSecretRevealLight), MYPY_ANNOTATION
+        msg = "ContractReceiveSecretRevealLight should be accompanied by a valid payment state"
+        assert payment_state, msg
+        iteration = handle_onchain_secretreveal(
+            payment_state=payment_state,
+            state_change=state_change,
+            channelidentifiers_to_channels=channelidentifiers_to_channels,
+            pseudo_random_generator=pseudo_random_generator,
+        )
```
mortelli: why not do this as an `or` on the previous block?

it seems that the only difference between these 2 is the `assert isinstance` (the purpose of which i am unsure) and the `msg` (which could be parametrized)

raiden/transfer/mediated_transfer/initiator.py
```diff
@@ -721,6 +722,40 @@ def handle_onchain_secretreveal(
...
 
+def handle_onchain_secretreveal_light(
+    initiator_state: InitiatorTransferState,
+    state_change: ContractReceiveSecretRevealLight,
+    channel_state: NettingChannelState,
+    pseudo_random_generator: random.Random,
+) -> TransitionResult[InitiatorTransferState]:
+    """ When a secret is revealed on-chain all nodes learn the secret.
+
+    This check the on-chain secret corresponds to the one used by the
```
mortelli:
```suggestion
    This checks that the on-chain secret corresponds to the one used by the
```

raiden/transfer/mediated_transfer/initiator.py
```diff
@@ -721,6 +722,40 @@ def handle_onchain_secretreveal(
     return iteration
 
 
+def handle_onchain_secretreveal_light(
```
mortelli, line 725: comparing this function to `handle_onchain_secretreveal`, 2 things:
1. could you briefly explain why this function does not care about events? (so that i can maybe understand the logic here a bit better)
2. should the docstrings be the same if the functions are different?

raiden/transfer/mediated_transfer/events.py
```diff
@@ -292,12 +293,25 @@ def __init__(
...
         super().__init__(recipient, channel_identifier, message_identifier)
 
         self.signed_locked_transfer = signed_locked_transfer
 
+        canonical_identifier = CanonicalIdentifier(
+            channel_identifier=channel_identifier,
+            token_network_address=self.signed_locked_transfer.token_network_address,
+            chain_identifier=self.signed_locked_transfer.chain_id
+        )
+        self.balance_proof = BalanceProofUnsignedState(
```
mortelli: i find this a bit confusing: why is this `Unsigned` if it uses a `signed_locked_transfer`?

mortelli (commented):
i think i've managed to more or less grasp everything except for the changes to `handle_contract_send_channelunlock`, i find these are going a bit over my head :sweat_smile:

raiden/transfer/mediated_transfer/events.py
mauricioirace: I changed the signed_locked_transfer type to LockedTransfer, as that was what it actually received from the caller. LockedTransfer doesn't have enought information to create a signed balance proof. I think we could change signed_locked_transfer's name to locked_transfer. in order to prevent confusions.

raiden/transfer/mediated_transfer/initiator.py
mauricioirace: 1. It just don't send the balance proof, as it would have been needed to modify the light client, and that would be too much time expensive without debray.

mauricioirace: 1. It just don't send the balance proof, as it would have been needed to modify the light client, and that would be too much time expensive without debray. Anyway it can be unlocked using the unlock mechanism.
2. No, they shouldn't. I'll update them

raiden/transfer/state_change.py
mauricioirace: Because some weirds things were happening

mortelli: yeah, this is not worth looking into any further. leave as is

raiden/transfer/mediated_transfer/initiator.py
marcosmartinez7: 1. basically it only update the local offchain state of the channel in order to allow the unlock to be fired right?

but the initiator (the lc) didnt handle this, therefore the balance proof isnt sent to the partner

raiden/transfer/mediated_transfer/events.py
mortelli: @mauricioirace was the type hint in the signature for this function wrong?

raiden/transfer/mediated_transfer/initiator.py
mauricioirace: exactly what marcos says

raiden/transfer/mediated_transfer/events.py
mauricioirace: Yes, both worked. But this one is more more correct, as the locked transfer and the bp are signed.

mortelli (approved):
LGTM as far as my understanding goes, just please make sure this is working as intended before merging (as always :stuck_out_tongue: )

marcosmartinez7 (approved):

raiden/blockchain_events_handler.py
JONAF2103: I don't know about this but ok if you approve it @marcosmartinez7 then i think is ok, but why is only the secret hash available here? i mean if we have that we should have the address too

JONAF2103 (approved):
LGTM

PR #166, opened 2020-12-01
RLP-849: iterate RIF comms tests

mortelli:
This PR:
1. iterates upon the RIF comms tests (this module is by no means final, though)
    1. adds the `test_nodes` variable for tidiness and removes random addresses.
    2. expands the `rif_comms_client` fixture with RSK and API addresses.
    3. renames some tests for clarity's sake.
    4. updates some test logic based on the latest behavior of the RIF comms node.
    5. removes some tests which were useless (or, in one case, a bit of an overkill for `pytest`) and adds other needed ones (like cross messaging).
    6. clarifies `reason` fields for skipped tests plus some comments.
2. extracts the static method `_notification_to_message` into `rif_comms/utils.py`, and makes it more generic.
    - instead of returning a `RaidenMessage`, now a `str` payload is returned. this `str` is then processed as a raiden message in the RIF comms transport node code.

There are still some `TODOs` to be resolved on the base branch or later down the road. This is meant to be an improvement rather than a final implementation.

builds on #162

### DIFF
--- a/raiden/tests/unit/rif_comms/test_rif_comms_client.py
+++ b/raiden/tests/unit/rif_comms/test_rif_comms_client.py
@@ -1,13 +1,9 @@
-import json
 import unittest
 
-import grpc
 import pytest
 from eth_utils import to_canonical_address
 from transport.rif_comms.client import Client as RIFCommsClient
 from transport.rif_comms.node import Node as RIFCommsNode
-from transport.rif_comms.proto.api_pb2 import Channel, PublishPayload, Msg
-from transport.rif_comms.proto.api_pb2_grpc import CommunicationsApiStub
 
 test_nodes = dict([
     (1, {
@@ -53,14 +49,13 @@ def rif_comms_client(request):
 @pytest.mark.usefixtures("rif_comms_client")
 class TestRIFCommsClient(unittest.TestCase):
     """
-    Test for RIFCommsClient. This class is for test the basic operations of the client.
+    Test class for RIFCommsClient. It covers the basic operations of the client.
 
-    How to use it:
-
-    - modify the test_nodes address and rif_comms_api fields
+    How to use:
+    - modify the `address and `comms-api` fields for the `test_nodes` variable
     - addresses are used as representations of the Lumino keystore
     - each of the nodes is used as a separate peer
-    - all tests assume that there is a RIF COMMS node already running
+    - all tests assume that there is a RIF Comms node already running
     """
 
     @pytest.mark.skip(reason="succeeds no matter what")
@@ -122,51 +117,6 @@ class TestRIFCommsClient(unittest.TestCase):
         peer_id = self.client_1._get_peer_id(self.address_1)
         self.client_1.disconnect()
         # TODO check if end comms deletes topics
-
-    @pytest.mark.skip(reason="fails due to as-of-yet undetermined reason, used to work")
-    def test_send_lumino_message(self):
-        # this test requires a lumino node started with a comms api matching test_nodes[1]["comms_api"]
-        stub = CommunicationsApiStub(grpc.insecure_channel(self.api_1))
-
-        # TODO: obtain this programmatically
-        channel_id = "16Uiu2HAmQmHFGepf3eaptGCkPhhw8ggipsrJzZyLTrnd5TJKWDHk"
-        # got this from subscription of lumino node
-
-        sub = stub.Subscribe(Channel(channelId=channel_id))
-        print("sub")
-
-        some_raiden_message = {
-            'type': 'LockedTransfer',
-            'chain_id': 33,
-            'message_identifier': 9074731958492744333,
-            'payment_identifier': 2958725218135700941,
-            'payment_hash_invoice': '0x0000000000000000000000000000000000000000000000000000000000000000',
-            'nonce': 45,
-            'token_network_address': '0xd548700d98f32f83b3c88756cf340b7f61877d75',
-            'token': '0xf563b16dc42d9cb6d7ca31793f2d62131d586d05',
-            'channel_identifier': 12,
-            'transferred_amount': 17000000000000000000,
-            'locked_amount': 1000000000000000000,
-            'recipient': '0x00e8249ee607ea67127c4add69291a6c412603c5',
-            'locksroot': '0x043c092c72059c4c154cb342409d95364888a98fa87efadece5add9c255dce9a',
-            'lock': {
-                'type': 'Lock',
-                'amount': 1000000000000000000,
-                'expiration': 1165251,
-                'secrethash': '0x2dc5a7ff26be395d443200db5d16b5d2aadae3a83836be648cd4cba8e2e555fe'
-            },
-            'target': '0x00e8249ee607ea67127c4add69291a6c412603c5',
-            'initiator': '0x27633dc87378a551f09f2fcf43a48fc2b3425d43',
-            'fee': 0,
-            'signature': '0xbaa4df61ab23ab8fdddf4a90fd5db7dd04da364795ebb717426bdd1fcefb411759e1b41d71c3c54b2a75c3eba0f2a2c54c846a0139120ee794f51b0e2a0d954d1c'
-        }
-
-        stub.SendMessageToTopic(
-            PublishPayload(
-                topic=Channel(channelId=channel_id),
-                message=Msg(payload=str.encode(json.dumps(some_raiden_message)))
-            )
-        )
 
     @pytest.mark.skip(reason="hangs when attempting to sub to a node without it having subbed to itself first")
     def test_two_clients_cross_messaging_same_topic(self):
--- a/raiden/tests/unit/rif_comms/test_rif_comms_client.py
+++ b/raiden/tests/unit/rif_comms/test_rif_comms_client.py
@@ -3,7 +3,7 @@ import unittest
 import pytest
 from eth_utils import to_canonical_address
 from transport.rif_comms.client import Client as RIFCommsClient
-from transport.rif_comms.node import Node as RIFCommsNode
+from transport.rif_comms.utils import notification_to_payload
 
 test_nodes = dict([
     (1, {
@@ -135,11 +135,11 @@ class TestRIFCommsClient(unittest.TestCase):
         self.client_2.send_message(payload_2, self.address_1)
 
         for resp in one_sub_two:
-            received_message = RIFCommsNode._notification_to_message(resp)
+            received_message = notification_to_payload(resp)
             assert received_message == payload_1
             break  # only 1 message is expected
 
         for resp in two_sub_one:
-            received_message = RIFCommsNode._notification_to_message(resp)
+            received_message = notification_to_payload(resp)
             assert received_message == payload_2
             break  # only 1 message is expected
--- a/raiden/tests/unit/rif_comms/test_rif_comms_client.py
+++ b/raiden/tests/unit/rif_comms/test_rif_comms_client.py
@@ -99,7 +99,7 @@ class TestRIFCommsClient(unittest.TestCase):
         # register node 1, get own peer id, disconnect
         notification = self.client_1.connect()
         peer_id = self.client_1._get_peer_id(self.address_1)
-        self.rif_comms_client.disconnect()
+        self.client_1.disconnect()
         # TODO check if end comms deletes topics
 
     @pytest.mark.skip(reason="ignore")
--- a/raiden/tests/unit/rif_comms/test_rif_comms_client.py
+++ b/raiden/tests/unit/rif_comms/test_rif_comms_client.py
@@ -1,119 +1,118 @@
 import json
-import secrets
-import time
 import unittest
 
 import grpc
 import pytest
-from coincurve import PublicKey
-from eth_utils import to_checksum_address, to_canonical_address
-from sha3 import keccak_256
+from eth_utils import to_canonical_address
 from transport.rif_comms.client import Client as RIFCommsClient
 from transport.rif_comms.proto.api_pb2 import Channel, PublishPayload, Msg
 from transport.rif_comms.proto.api_pb2_grpc import CommunicationsApiStub
 
-
-def get_random_address_str() -> str:
-    private_key = keccak_256(secrets.token_bytes(32)).digest()
-    public_key = PublicKey.from_valid_secret(private_key).format(compressed=False)[1:]
-    addr = keccak_256(public_key).digest()[-20:]
-    return to_checksum_address(addr)
-
-
-LUMINO_1_ADDRESS = to_canonical_address("0x8cb891510dF75C223C53f910A98c3b61B9083c3B")
-LUMINO_1_COMMS_API = "localhost:5013"
-
-LUMINO_2_COMMS_API = "localhost:5016"
-LUMINO_2_ADDRESS = to_canonical_address("0xeBfF0EEe8E2b6952E589B0475e3F0E34dA0655B1")
-
-LUMINO_3_ADDRESS = "0x636BA79E46E0594ECbbEBb4F74B9336Fd4454442"
+test_nodes = [
+    {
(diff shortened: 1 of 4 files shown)


transport/rif_comms/node.py
```diff
@@ -281,9 +248,9 @@ def send_message(self, payload: str, recipient: Address):
         is_subscribed_to_receiver_topic = self._comms_client.is_subscribed_to(recipient)
         # if not, create the topic subscription
         if not is_subscribed_to_receiver_topic:
-            topic_id, _ = self._comms_client.subscribe_to(recipient)
+            self._comms_client.subscribe_to(recipient)
         # send the message
-        self._comms_client.send_message(payload, topic_id)  # TODO: exception handling for RIF Comms client
```
mortelli, line 286: `topic_id` might be not set here, but we always have the recipient

transport/rif_comms/utils.py
```diff
@@ -0,0 +1,24 @@
+import json
+
+from transport.rif_comms.proto.api_pb2 import ChannelNewData
+
+
+def notification_to_payload(notification_data: ChannelNewData) -> str:
```
mortelli, line 6: returning a `str` payload is arguably tidier due to the fact that (like in some of the new test cases) messages could be just any string, instead of specifically raiden messages.

the `str` → `RaidenMessage` transformation is now done in the `RIFCommsNode._receive_messages` function.

transport/rif_comms/node.py
```diff
@@ -169,21 +150,7 @@ def _run(self, *args: Any, **kwargs: Any) -> None:
...
-        time.sleep(10)
-
-        # subscribe to the other node topic and wait
-        peer_id, other_node_topic_stream = self._comms_client.subscribe_to(
-            to_canonical_address("0x138aF366e0ED7Cc4b9747a935D1b5F75A86b9D83"))
-
-        # now send a message to the other node that is already subscribed
-        self._comms_client.send_message("hello from lumino node",
-                                        peer_id)
-        # END POC
```
mortelli, line 186: this should be removed in the base branch

transport/rif_comms/node.py
```diff
@@ -83,36 +83,17 @@ def _receive_messages(self):
...
         """
         for notification in self._our_topic_stream:
-            raiden_message = self._notification_to_message(notification.channelNewData)
-            if raiden_message:
-                self.log.info("incoming message", message=raiden_message)
-                self._handle_message(raiden_message)
+            payload = notification_to_payload(notification)
 
-    @staticmethod
-    def _notification_to_message(notification_data: ChannelNewData) -> RaidenMessage:
```
mortelli, line 92: partly extracted to `rif_comms/utils.py`

transport/rif_comms/node.py
```diff
@@ -83,36 +83,17 @@ def _receive_messages(self):
...
-        return None
+            try:
+                message_dict = json.loads(payload)
+                raiden_message = message_from_dict(message_dict)
+
+                if raiden_message:
+                    self.log.info("incoming message", message=raiden_message)
+                    self._handle_message(raiden_message)
+            except (TypeError, InvalidProtocolMessage):
+                self.log.error("incoming message could not be processed", payload=payload)
```
mortelli, line 96: once the payload is extracted, all this code is specific to processing raiden messages in a RIF comms transport node

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -1,181 +1,145 @@
...
-        one_own_sub = self.rif_comms_client1.subscribe_to(LUMINO_1_ADDRESS)
-        two_one_sub = self.rif_comms_client2.subscribe_to(LUMINO_1_ADDRESS)
-
-        peer_id_1 = self.rif_comms_client2._get_peer_id(LUMINO_2_ADDRESS)
-
-        for resp in one_own_sub:
-            print("Respone for 1: ", resp)
-
-        for resp in two_one_sub:
-            print("Respone for 2: ", resp)
```
mortelli, line 181: this didn't match the new signature for rif comms subscription calls

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -1,181 +1,145 @@
...
-
-        stub.SendMessageToTopic(
-            PublishPayload(
-                topic=Channel(channelId="16Uiu2HAm9otWzXBcFm7WC2Qufp2h1mpRxK1oox289omHTcKgrpRA"),
-                message=Msg(payload=str.encode(json.dumps(some_raiden_message)))
-            )
-        )
-
-    @pytest.mark.skip(reason="ignore")
-    def test_two_clients_topic_subscription(self):
```
mortelli, line 152: the idea for this test is now in `test_two_clients_cross_subscription`

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -1,181 +1,145 @@
...
-        peer_id = self.rif_comms_client._get_peer_id(LUMINO_1_ADDRESS)
-        self.rif_comms_client.disconnect()
+        # register node 1, get own peer id, disconnect
+        notification = self.client_1.connect()
+        peer_id = self.client_1._get_peer_id(self.address_1)
+        self.client_1.disconnect()
         # TODO check if end comms deletes topics
 
-    @pytest.mark.skip(reason="ignore")
-    def test_send_lumino_message(self):
```
mortelli, line 110: this test was meant to be temporary. 

it's already heavy enough to have to start up a comms node through python (which we plan to do soon); a lumino node on top of that is not worth the hassle.

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -1,181 +1,145 @@
...
+        self.client_1.subscribe_to(self.address_2)
+        self.client_2.subscribe_to(self.address_1)
+        self.client_2.subscribe_to(self.address_2)
+
+        assert self.client_1.is_subscribed_to(self.address_1) is True
+        assert self.client_1.is_subscribed_to(self.address_2) is True
+        assert self.client_2.is_subscribed_to(self.address_1) is True
+        assert self.client_2.is_subscribed_to(self.address_2) is True
+
+    @pytest.mark.skip(reason="incomplete test")
```
mortelli, line 113: this still doesn't make much sense

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -1,181 +1,145 @@
...
+    api_2 = test_nodes[2]["comms-api"]
+    client_2 = RIFCommsClient(test_nodes[2]["address"], test_nodes[2]["comms-api"])
 
     def teardown():
-        rif_comms_client1.disconnect()
-        rif_comms_client2.disconnect()
+        client_1.disconnect()
+        client_2.disconnect()
 
     # request.addfinalizer(teardown)
```
mortelli, line 39: still needs to be resolved

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -1,181 +1,145 @@
...
-LUMINO_1_ADDRESS = to_canonical_address("0x8cb891510dF75C223C53f910A98c3b61B9083c3B")
-LUMINO_1_COMMS_API = "localhost:5013"
-
-LUMINO_2_COMMS_API = "localhost:5016"
-LUMINO_2_ADDRESS = to_canonical_address("0xeBfF0EEe8E2b6952E589B0475e3F0E34dA0655B1")
-
-LUMINO_3_ADDRESS = "0x636BA79E46E0594ECbbEBb4F74B9336Fd4454442"
+from transport.rif_comms.utils import notification_to_payload
+
+test_nodes = dict([
```
mortelli: `dict` that can be accessed by index, and also starts at index 1 (better matches the variable names)

transport/rif_comms/client.py
```diff
@@ -72,13 +71,14 @@ def is_subscribed_to(self, rsk_address: Address) -> bool:
...
-    def send_message(self, payload: str, topic_id: str):
+    def send_message(self, payload: str, rsk_address: Address):
         """
         Sends a message to a destination RSK address.
         Invokes the SendMessageToTopic GRPC API endpoint.
         :param payload: the message data to be sent
-        :param topic_id: the topic identifier to be sent to
+        :param rsk_address: the destination for the message to be sent to
         """
+        topic_id = self._get_peer_id(to_checksum_address(rsk_address))
```
mortelli, line 81: stuff like this should be looked at [here](https://jirainfuy.atlassian.net/browse/RLP-924)

raiden/tests/unit/rif_comms/test_rif_comms_client.py
marcosmartinez7: yes, i have created an issue for this and we can check them in a future, i didnt found this really important now

https://jirainfuy.atlassian.net/browse/RLP-925

raiden/tests/unit/rif_comms/test_rif_comms_client.py
marcosmartinez7: yeah, the idea was to connect again and check that the subscriptions are gone, but we must add to this tets:

- subscribe to a topic
- connect again after the disconnect
- check for subscriptions

marcosmartinez7: im ok with keeping the TODO here and creating an issue for this with really low priority


https://jirainfuy.atlassian.net/browse/RLP-926

raiden/tests/unit/rif_comms/test_rif_comms_client.py
marcosmartinez7: yeah, this test was just a script for test lumino that doesnt make sense now

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -1,181 +1,145 @@
...
+        self.client_1.send_message(payload_1, self.address_2)
+        self.client_2.send_message(payload_2, self.address_1)
+
+        for resp in one_sub_two:
+            received_message = notification_to_payload(resp)
+            assert received_message == payload_1
+            break  # only 1 message is expected
+
+        for resp in two_sub_one:
+            received_message = notification_to_payload(resp)
```
marcosmartinez7, line 143: I think this will not work without breaking the for loop somehow

raiden/tests/unit/rif_comms/test_rif_comms_client.py
marcosmartinez7: this line?        `  for resp in two_one_sub: `

why?

transport/rif_comms/client.py
marcosmartinez7: we must fully review the test cases after the new proto is done and remove this kind of invocations

transport/rif_comms/node.py
marcosmartinez7: we must add the validate and parse msg here, maybe we can do that before merging the pr of p2p payments

transport/rif_comms/node.py
marcosmartinez7: i have removed this on the p2p payments poc so we can take care of this on the merge of that pr

transport/rif_comms/node.py
marcosmartinez7: but this assumes that the sendMessageToTopic can work with an rsk address right? 

is that working? 

i was using the topic id here because of that

transport/rif_comms/utils.py
marcosmartinez7: agree, but we must add the validate and parse method for getting the list of messages

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: thanks!

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: not specifically that line, but the whole test.

for starters, `subscribe_to` now outputs 2 variables instead of one, which didn't match the code.

but also: unless a message was sent—which wasn't happening here—these `for` statements would just hang, since the `subscribe_to` code now fetches the first notification which contains the topic ID.

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: sorry, what do you mean by this? there is a `break` on line `145`

transport/rif_comms/node.py
mortelli: > i was using the topic id here because of that

we can't do this because `if not is_subscribed_to_receiver_topic` is `False`, then `topic_id` will raise a `NameError` exception, as it is undefined.

i _think_ the **only** reason why this is working so far is because the `is_subscribed_to` calls are always (wrongly) returning `False`—this is a known bug—and so you always fetched the `topic_id`.

> but this assumes that the sendMessageToTopic can work with an rsk address right?
> 
> is that working?

yes, i've verified this through the tests.

note that:
1. now, `_get_peer_id` is being called inside the `send_message` function, which allows us to get the topic id.
2. `sendMessageToTopic` still uses a topic ID (same as before).

see here:
https://github.com/rsksmart/lumino/blob/a24ced2059d6512e3a47e1beba7a11cd3f143e8b/transport/rif_comms/client.py#L74-L88

---

another reason for this change is these signatures:

https://github.com/rsksmart/lumino/blob/a24ced2059d6512e3a47e1beba7a11cd3f143e8b/transport/node.py#L55-L56

https://github.com/rsksmart/lumino/blob/a24ced2059d6512e3a47e1beba7a11cd3f143e8b/transport/rif_comms/node.py#L243

it seems preferable to me that the client matches these 2.

marcosmartinez7: Lgtm then

raiden/tests/unit/rif_comms/test_rif_comms_client.py
marcosmartinez7: My bad

im courious how it works anyway, since we break at the first response, what about the empty response of the subscribe?

marcosmartinez7 (approved):

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: the client's `subscribe_to` method already does this:

https://github.com/rsksmart/lumino/blob/a24ced2059d6512e3a47e1beba7a11cd3f143e8b/transport/rif_comms/client.py#L41-L55

this doesn't solve the underlying issue that the `for` inside that method could hang, as it is currently doing when subscribing to a topic to which its owner has not subscribed to yet.

i've created an issue to eventually resolve this, see [here](https://jirainfuy.atlassian.net/browse/RLP-929)

marcosmartinez7: Great!

raiden/tests/unit/rif_comms/test_rif_comms_client.py
mortelli: i've simplified this structure (seems like i overcomplicated it in the first place)

raiden/tests/unit/rif_comms/test_rif_comms_client.py
```diff
@@ -1,181 +1,145 @@
...
+    3: {
+        "address": to_canonical_address("0x636BA79E46E0594ECbbEBb4F74B9336Fd4454442"),
+        "comms-api": "localhost:7013",
+    },
+}
 
 
 @pytest.mark.usefixtures("rif_comms_client")
 @pytest.fixture(scope="class")
 def rif_comms_client(request):
```
mortelli, line 26: i don't particularly like the way this fixture is written, but changing it is out of scope for now

PR #168, opened 2020-12-03
Feature/rlp 847/start comms and lumino

marcosmartinez7:
# Description:

This PR include scripts to run both a rif comms node and a lumino node

# Changelog 

Two scripts were added: 

**startLumino**: minimal modifications to the old script (i have removed the --address option and password file because that approach is not compatible with production env imo)

**startLuminoWithTransport**:  (probably startLuminoWIthComms is better since matrix is another transport layer that is not relevant here)
- start rif comms with a lot of assumptions
- logs the rif comms pub sub bootnode std output to a file
- start lumino if wanted

builds on open pr #162


startLuminoWithTransport.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
```
mortelli: i agree with the pr description, name this something rif comms-specific

startLumino.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
+
+# How it works
+# ./startLumino <ENVIRONMENT_NAME>
+# This script is to start lumino using an environment called <ENVIRONMENT_NAME>.
+# ENVIRONMENT_NAME defaults to luminoNode
```
mortelli: i would just call the default `lumino`

startLumino.sh
```diff
@@ -0,0 +1,83 @@
...
+
+# Here some explanation of the parameters:
```
mortelli:
```suggestion
# Here is an explanation of the parameters:
```

startLumino.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
+
+# How it works
+# ./startLumino <ENVIRONMENT_NAME>
+# This script is to start lumino using an environment called <ENVIRONMENT_NAME>.
```
mortelli:
```suggestion
# This script starts Lumino using an environment called <ENVIRONMENT_NAME>.
```

startLumino.sh
```diff
@@ -0,0 +1,83 @@
...
+# HUB_MODE: Sets if the node will be a hub or not
+# RSK_ENDPOINT: The RSK endpoint url
+# EXPOSE_ENDPOINT: The expose url to use by the lumino node
+# PORT: The port number to setup for the lumino node
+# NETWORK_ID: The network id to connect to
+
+ENV_NAME=$1
+
+if [ "" == "${ENV_NAME}" ]; then
+  ENV_NAME=clientEnv
```
mortelli: this doesn't match what line `6` says

startLumino.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
+
+# How it works
+# ./startLumino <ENVIRONMENT_NAME>
+# This script is to start lumino using an environment called <ENVIRONMENT_NAME>.
+# ENVIRONMENT_NAME defaults to luminoNode
+
+# Here some explanation of the parameters:
+# ADDRESS_TO_USE: The address to set to be used by the node
+# PASSWORD_FILE_CONTENT: The password for the private key of the ADDRESS_TO_USE address.
```
mortelli: just call this `PASSWORD`

startLuminoWithTransport.sh
```diff
@@ -0,0 +1,52 @@
...
+RIF_COMMS_NODE_CONFIGURATION=$1
+RIF_COMMS_NODE_PORT=$2
+
+
+# move to the previous folder where is the rif comms node
+cd ..
+cd rif-communications-pubsub-bootnode/
+
+# load nvm
+export NVM_DIR="$([ -z "${XDG_CONFIG_HOME-}" ] && printf %s "${HOME}/.nvm" || printf %s "${XDG_CONFIG_HOME}/nvm")"
```
JONAF2103: If you put NVM 14.13.1 as a pre-requirement you don't need to do this.

startLumino.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
+
+# How it works
+# ./startLumino <ENVIRONMENT_NAME>
+# This script is to start lumino using an environment called <ENVIRONMENT_NAME>.
+# ENVIRONMENT_NAME defaults to luminoNode
+
+# Here some explanation of the parameters:
+# ADDRESS_TO_USE: The address to set to be used by the node
```
mortelli:
```suggestion
# RSK_ADDRESS: The RSK address to be used by the node
```

startLumino.sh
```diff
@@ -0,0 +1,83 @@
...
+# PASSWORD_FILE_CONTENT: The password for the private key of the ADDRESS_TO_USE address.
+# TOKEN_NETWORK_REGISTRY: The token network registry contract address to be used by the node.
+# SECRET_REGISTRY: The secret registry contract address to be used by the node.
+# ENDPOINT_REGISTRY: The endpoint registry contract address to be used by the node.
+# HUB_MODE: Sets if the node will be a hub or not
```
mortelli:
```suggestion
# HUB_MODE: If set, the node will work as a hub
```

startLumino.sh
```diff
@@ -0,0 +1,83 @@
...
+# RSK_ENDPOINT: The RSK endpoint url
```
mortelli:
```suggestion
# RSK_ENDPOINT: The RSK node endpoint URL
```

startLumino.sh
```diff
@@ -0,0 +1,83 @@
...
+# EXPOSE_ENDPOINT: The expose url to use by the lumino node
+# PORT: The port number to setup for the lumino node
```
mortelli: i would just merge these 2 into 1, we're already using just 1 value for the RSK node URL, for example

startLuminoWithTransport.sh
```diff
@@ -0,0 +1,52 @@
...
+[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh" # This loads nvm
+
+# Set node version 14.13.1
+nvm use v14.13.1
+
+# Start rif comms node
+export NODE_ENV=$RIF_COMMS_NODE_CONFIGURATION
+npm run api-server > commsNodeLog.txt &
```
JONAF2103: You should use this `npm run api-server > commsNodeLog.log 2>&1` instead, 2 and 1 are error and stdout, if you only put & the command will run async but all the logs are going to be out on the same screen as the script runner and the commsNodeLog will have only stdout or even will be empty.

startLumino.sh
```diff
@@ -0,0 +1,83 @@
...
+# NETWORK_ID: The network id to connect to
```
mortelli:
```suggestion
# NETWORK_ID: The blockchain network ID to connect to
```

startLuminoWithTransport.sh
```diff
@@ -0,0 +1,52 @@
...
+
+echo "Rif comms running, logs at "$(pwd)"/commsNodeLog.txt"
+
+# Start lumino node
+
+read -p "Press y to start lumino " -n 1 -r
+if [[ ! $REPLY =~ ^[Yy]$ ]]
+then
+    COMMS_PID=$(lsof -sTCP:LISTEN -ti :$RIF_COMMS_NODE_PORT)
```
JONAF2103: I think you could get RIF_COMMS_NODE_PORT from the config file directly instead of using a parameter, since your first param is the node server config you already have the port there

startLuminoWithTransport.sh
```diff
@@ -0,0 +1,52 @@
...
+    echo "Killing rif comms process: "$COMMS_PID
+    kill -9 $COMMS_PID
+else
+    cd ..
```
JONAF2103: If you take my advice [earlier](https://github.com/rsksmart/lumino/pull/168/files#r536071256) then here you don't have to move anywhere since you will always be in the same folder.

startLuminoWithTransport.sh
```diff
@@ -0,0 +1,52 @@
...
+# Run this script into the directory in which Lumino was git cloned
+# Specify the node configuration and the port that its already configured: i.e ./startLuminoWithTransport.sh nodeA 5013
+# The output log for rif comms node is on rif-communications-pub-sub-bootnode under the commsNodeLog.txt name
+
+RIF_COMMS_NODE_CONFIGURATION=$1
+RIF_COMMS_NODE_PORT=$2
+
+
+# move to the previous folder where is the rif comms node
+cd ..
```
JONAF2103: This can be replaced by something like this:
```bash
# Set node version 14.13.1
nvm use v14.13.1
npm run api-server --prefix ../rif-communications-pubsub-bootnode > commsNodeLog.log 2>&1

```

JONAF2103 (commented):
I think you can improve this script a little, i leave a bunch of comments there, but it looks good to me in general anyway.

startLumino.sh
```diff
@@ -0,0 +1,83 @@
...
+
+ENV_NAME=$1
+
+if [ "" == "${ENV_NAME}" ]; then
+  ENV_NAME=clientEnv
+fi
+
+KETYSTORE_PATH="/home/marcos/rsk/keystore"
```
mortelli: this doesn't look right

startLumino.sh
```diff
@@ -0,0 +1,83 @@
...
+TOKEN_NETWORK_REGISTRY="0x07e1CD1ea2123e3a0f624553761EaD1c1e150CC3"
+SECRET_REGISTRY="0xfA8Bc06C815BD4C4bB641ccb00EBF9CB8BEB2d67"
+ENDPOINT_REGISTRY="0xC14E9A67A87949bC131b9f22a4EB5cd9d9A6728e"
```
mortelli: these should probably be params, no?

startLumino.sh
```diff
@@ -0,0 +1,83 @@
...
+HUB_MODE=0
+RSK_ENDPOINT=http://localhost:4444
+EXPOSE_ENDPOINT=http://localhost
```
mortelli: i would maybe call this `LUMINO_API_ENDPOINT`

startLuminoWithTransport.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+
+### Usage
+
+## This script assumes:
+
+# You have NVM installed
+# You have NVM 14.13.1 installed
```
mortelli: isn't this `npm`?

startLuminoWithTransport.sh
```diff
@@ -0,0 +1,52 @@
...
+# You have ran npm install on the rif communications pub sub bootnode project
+# You have created the config files for the rif communications pub sub bootnode (i.e /config/nodeA and keys/nodeA)
+# You already know and configured the PORT in which rif comms node is going to run
+
+## Running
+
+# Run this script into the directory in which Lumino was git cloned
```
mortelli:
```suggestion
# Run this script in the directory in which Lumino was git cloned
```

mortelli (commented):
i know this was probably not the purpose of the task, but it would be nice if we could start shaping this up to be scripts to eventually be run in a testing or production environment.

some default values go against this idea (like regtest network ID, no sync flag, etc.)

with this in mind, maybe we can explicitly name these scripts with something like `development` in their name.

all in all LG :+1:

startLuminoWithTransport.sh
marcosmartinez7: im not sure about this, i mean, at least this implies that the dev must not load nvm.. but.. maybe is not a big feature

startLuminoWithTransport.sh
marcosmartinez7: thanks! i have tested that but it was throwing me an error, im going to try again

startLuminoWithTransport.sh
marcosmartinez7: yeah, its a little bit hard to parse the json5 file i think, but i will try

startLuminoWithTransport.sh
marcosmartinez7: thxs!

startLumino.sh
marcosmartinez7: this is the venv, it must be clientEnv imo

startLuminoWithTransport.sh
marcosmartinez7: i think it should be` npm run api-server > commsNodeLog.log 2>&1 &`
(with the trailing &)
in order to return the control

marcosmartinez7: yep, it works, thanks!!

startLuminoWithTransport.sh
marcosmartinez7: done

startLumino.sh
marcosmartinez7: done

startLumino.sh
marcosmartinez7: fixed

startLumino.sh
marcosmartinez7: I have removed for now since we dont know how to secure this on production
also, i wasnt using it

startLumino.sh
marcosmartinez7: agree, and done

startLumino.sh
marcosmartinez7: this must be a script variable

startLumino.sh
marcosmartinez7: we have a lot of params, im not a fan of reading config files from bash but maybe is better

startLumino.sh
marcosmartinez7: done

startLuminoWithTransport.sh
marcosmartinez7: in fact is node..

maybe i can remove the nvm stuff and just assume that the user uses node 14.13.1

startLuminoWithTransport.sh
marcosmartinez7: i decided to remove this and the script just assumes a version of node

nvm should not be a requirement

startLuminoWithTransport.sh
mortelli: latest version for `nvm` is `v0.37.2` so this can't be right

startLumino.sh
mortelli: how would hard-coded contract addresses ever work on different machines?

startLuminoWithTransport.sh
marcosmartinez7: its node version not nvm nor npm

mortelli: > its node version not nvm nor npm

it says `You have NVM 14.13.1 installed`; if we need node 14.13, just say `node 14.13`, no?

startLuminoWithTransport.sh
marcosmartinez7: out of scope for now

startLuminoWithTransport.sh
marcosmartinez7: done

startLuminoWithTransport.sh
marcosmartinez7: done

startLumino.sh
marcosmartinez7: now the script assumes an already created virtualenv

startLumino.sh
marcosmartinez7: deprecated

startLumino.sh
marcosmartinez7: deprecated

startLumino.sh
marcosmartinez7: deprecated

startLumino.sh
marcosmartinez7: deprecated

startLumino.sh
marcosmartinez7: done

startLumino.sh
marcosmartinez7: done

startLumino.sh
marcosmartinez7: done

startLumino.sh
marcosmartinez7: done

startLumino.sh
marcosmartinez7: moved to script variables

startLuminoWithTransport.sh
marcosmartinez7: yep, fixed

mauricioirace (approved):
LGTM! Anyway I would wait for Master Jona(@JONAF2103 )

startLuminoWithRIFComms.sh
```diff
@@ -0,0 +1,130 @@
...
+opts=$(getopt \
+    --longoptions "$(printf "%s:," "${ARGUMENT_LIST[@]}")" \
+    --name "$(basename "$0")" \
+    --options "" \
+    -- "$@"
+)
+
+eval set --$opts
+
+while [[ $# -gt 0 ]]; do
```
JONAF2103, line 59: If this are the same as the child script you can skip this and send the params to the child directly from this parent, something like this should be ok ```./startLumino.sh $@```, that will pass the same params from this script ot the child script and you don't need to validate them all here again.

startLuminoWithRIFComms.sh
```diff
@@ -0,0 +1,130 @@
...
+            shift 2
+            ;;
+        *)
+            break
+            ;;
+    esac
+done
+
+# Start rif comms node
+export NODE_ENV=$RIF_COMMS_NODE_CONFIGURATION
```
JONAF2103, line 112: As the comment above you can just validate the rif comms params and then send all the other params to startLumino since that script already validate the others.

JONAF2103 (commented):
it looks good but you should remove the duplicated code to validate params in the parent since the child script already does that

startLuminoWithRIFComms.sh
marcosmartinez7: oh thats nice! i will take a look at this

marcosmartinez7: The thing is, that the specific variables to rif comms, for example: RIF_COMMS_NODE_PORT, must be validated on the parent script, and, since we are using $0, we are passing all the params to the child.

The child, doesnt recognize the RIF_COMMS_NODE_PORT flag... 

we must have a way to send a subset of $0

what do u think?

startLumino.sh
mortelli: so the user must input their password if their account has one, no? that's fine by me

startLumino.sh
mortelli: it's good now. i would just leave names (e.g. `marcos`) out of the example but it is up to you

mortelli (approved):
i haven't tested this, but it looks good as a first implementation.

2 additional thoughts:
- we definitely should look into what @JONAF2103 is saying about removing duplicated code and do so if possible.
- we have to start thinking about how much Lumino knows about RIF Comms. 

regarding the second point: maybe it's fine to start both programs through the shell (rather than having Lumino start RIF Comms) but after that, we should probably store information about the processes or configurations if both were started together. this will probably come in handy when having to restart Comms from Lumino.

i am working on something _somewhat_ similar in #170. we should put our heads together here and see if we can come up with improvements for any or both of these tasks.

startLuminoWithRIFComms.sh
JONAF2103: Yeah i get that, you can use a slice in that case if you want, be careful because @0 doesn't do anything, i know you mean to put $@, anyway using the slice should work, here some examples:

```bash
A=( foo bar "a  b c" 42 )
B=("${A[@]:1:2}")
C=("${A[@]:1}")       # slice to the end of the array
echo "${B[@]}"        # bar a  b c
echo "${B[1]}"        # a  b c
echo "${C[@]}"        # bar a  b c 42
echo "${C[@]: -2:2}"  # a  b c 42 # The space before the - is necesssary

```

you can see it [here](https://stackoverflow.com/questions/1335815/how-to-slice-an-array-in-bash)

PR #169, opened 2020-12-04
bugfix/RLP-828/channel-not-deleted

mauricioirace:
### changes
- `raiden/blockchain_events_handler.py`: check if participant is current node, or a handled LC,  instsead of checking only if it's a full node
- Simplify `handle_batch_unlock logic`, as it was duplicating channel deletion
- `handle_batch_unlock` now uses `subdispatch_to_channels_by_participant_address`, using the nodes address if it's payee or payer (it didn't work with payers before) or a lc address if corresponds
- Add `channel_identifier` property to `ContractReceiveChannelBatchUnlock`, as it's needed by `subdispatch_to_channels_by_participant_address`

### DIFF
--- a/raiden/blockchain_events_handler.py
+++ b/raiden/blockchain_events_handler.py
@@ -519,9 +519,9 @@ def handle_channel_batch_unlock(raiden: "RaidenService", event: Event):
     )
     assert token_network_state is not None
 
-    if participant1 == raiden.address:
+    if participant1 == raiden.address or participant1 in token_network_state.channelidentifiers_to_channels:
         partner = participant2
-    elif participant2 == raiden.address:
+    elif participant2 == raiden.address or participant2 in token_network_state.channelidentifiers_to_channels:
         partner = participant1
     else:
         log.debug(
--- a/raiden/transfer/state_change.py
+++ b/raiden/transfer/state_change.py
@@ -1206,6 +1206,10 @@ class ContractReceiveChannelBatchUnlock(ContractReceiveStateChange):
     def token_network_identifier(self) -> TokenNetworkAddress:
         return TokenNetworkAddress(self.canonical_identifier.token_network_address)
 
+    @property
+    def channel_identifier(self) -> ChannelID:
+        return self.canonical_identifier.channel_identifier
+
     def __repr__(self) -> str:
         return (
             "<ContractReceiveChannelBatchUnlock "
--- a/raiden/transfer/token_network.py
+++ b/raiden/transfer/token_network.py
@@ -27,6 +27,7 @@ StateChangeWithChannelID = Union[
     ContractReceiveChannelNewBalance,
     ContractReceiveChannelSettled,
     ContractReceiveUpdateTransfer,
+    ContractReceiveChannelBatchUnlock
 ]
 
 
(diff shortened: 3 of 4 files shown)


raiden/transfer/token_network.py
```diff
@@ -254,30 +255,19 @@ def handle_batch_unlock(
...
-            token_network_state.partneraddresses_to_channelidentifiers[
-                channel_state.partner_state.address
-            ].remove(channel_state.identifier)
-
-            del token_network_state.channelidentifiers_to_channels[state_change.participant][channel_state.identifier]
+    participant = None
+    if state_change.participant in token_network_state.channelidentifiers_to_channels:
+        participant = state_change.participant
+    elif state_change.partner in token_network_state.channelidentifiers_to_channels:
+        participant = state_change.partner
```
mortelli, line 262: are you aware that if both of these equal `False`, participant will be `None`?

this will apparently not cause an exception in `subdispatch_to_channels_by_participant_address`.

are we OK with this? (honestly don't know)

raiden/blockchain_events_handler.py
```diff
@@ -519,9 +519,9 @@ def handle_channel_batch_unlock(raiden: "RaidenService", event: Event):
     )
     assert token_network_state is not None
```
marcosmartinez7, line 520: is the handle_channel_batch_unlock used both for light and full channels?

raiden/blockchain_events_handler.py
```diff
@@ -519,9 +519,9 @@ def handle_channel_batch_unlock(raiden: "RaidenService", event: Event):
...
 
-    if participant1 == raiden.address:
+    if participant1 == raiden.address or participant1 in token_network_state.channelidentifiers_to_channels:
```
marcosmartinez7, line 522: i dont understand the `participant1 in token_network_state.channelidentifiers_to_channels:` part here

if the two light clientes are into the same hub, both are going to be present on  `token_network_state.channelidentifiers_to_channels`: right?

could you explain how this works?

raiden/transfer/state_change.py
```diff
@@ -1206,6 +1206,10 @@ def __init__(
     def token_network_identifier(self) -> TokenNetworkAddress:
         return TokenNetworkAddress(self.canonical_identifier.token_network_address)
 
+    @property
+    def channel_identifier(self) -> ChannelID:
```
marcosmartinez7, line 1210: purpose and usage?

raiden/transfer/token_network.py
```diff
@@ -254,30 +255,19 @@ def handle_batch_unlock(
     block_number: BlockNumber,
     block_hash: BlockHash,
 ) -> TransitionResult:
-    events = list()
```
marcosmartinez7, line 257: thats a lot of business logic removed

could you explain why? 

note that, really important stuff like the removal of a channel is gone

`             del token_network_state.channelidentifiers_to_channels[state_change.participant][channel_state.identifier] `

raiden/transfer/token_network.py
marcosmartinez7: yeah, this is important, because before this change it was not calling the state transition function for None channels

marcosmartinez7:
Lets discuss the conversation above, also, did you:

- test it manually on each possible scenario
-  ran the unit tests

?

raiden/blockchain_events_handler.py
mauricioirace: Yes. I don't why all the functions here are both for light clients and full nodes

raiden/blockchain_events_handler.py
mauricioirace: Yes. This codes just checks if the node/lc is a participant of the channel, what is needed in order to know if it must dispatch a state_change

raiden/transfer/state_change.py
mauricioirace: as I explained in description, subdispatch_to_channels_by_participant_address needs this

raiden/transfer/token_network.py
mauricioirace: subdispatch_to_channels_by_participant_address calls subdispatch_to_channels, which already does that if the channel must be deleted.

mauricioirace: What I mean is, the deleted code was duplicated

raiden/transfer/token_network.py
mauricioirace: when the participant address is None, the subdispatch_to_channels_by_participant_address does nothing. Maybe we should remove the optionallity of that param someday

raiden/blockchain_events_handler.py
marcosmartinez7: but suposse two light clients into the same hub, this will assign arbitrary..

JONAF2103 (commented):
I think it looks ok but we should make sure the baseline for this branch runs ok and all the tests pass before merging a change like this one, i'm just thinking about the removed logic scope that's all

raiden/blockchain_events_handler.py
mauricioirace: This participant is only used to fetch the channel state from db (something like what was done on unlock), and check its existence. So for this purpose is not relevant. 

Anyway, I agree its not very clean, but I didn't want to touch this part to much.

marcosmartinez7 (approved):
lgtm, lets see the manual test and unit test results in order to merge

JONAF2103 (approved):
LGTM

PR #170, opened 2020-12-07
RLP-919: automate comms tests

mortelli:
This PR is the first implementation for [RLP-919](https://jirainfuy.atlassian.net/browse/RLP-919), which is about automating RIF Comms test in lieu of executing them manually.

## Overview
The solution is a middle-ground between completely autonomous tests and an implementation that could be achieved within the time-frame and with the priorities that we have. 

The main objective here is not to provide a perfect solution, but rather a way to run these tests through the command-line without having to manually start RIF Comms processes. While a better, possibly fully independent version of these tests can be programmed, its development time is too costly to go through with it at the moment.

As a result, multiple issues have been created (and referenced in this PR) in order to later iterate on them, should we have the need and capability to do so.

### Fixtures
All RIF comms tests now use the new fixture `comms_nodes` and are parametrized with `amount_of_nodes`. This means that every test will generate a dynamic amount of `CommsNode` instances, which is a new class that represents a running comms client and process. Each of these are in charge of booting up the RIF Comms pub-sub node process at the OS level and then connecting to it.

This new fixture allows each test to do a proper set-up and tear-down, [which we couldn't do before](https://github.com/rsksmart/lumino/pull/162#discussion_r536151033). 

### New classes
The `CommsNode` is initialized by a `CommsConfig` object, which generates the attributes necessary to boot up the comms node (API endpoint URL, RSK address, port, etc.). The config object will attempt to load a configuration file from disk based on the number of the generated comms node (`1`, `2`, and so on) to obtain the attributes that it cannot generate by itself.

The `CommsNode` class also contains a `RIFCommsClient` object used by the tests, and internally stores other data such as a reference to the running comms process.

The `CommsProcess` class contains static methods to start up a non-blocking RIF Comms node process, as well as end it based on its process ID (PID).

## Limitations
- there are some new requirements for the tests to make up for the elements still not automated:
  - the [RIF Comms repo](https://github.com/rsksmart/rif-communications-pubsub-bootnode) must be cloned at the same level as the project which contains the tests being executed. the folder name must match the default repo name. eventually, this should be replaced by a dependency for the testing environment.
  - a valid configuration for each RIF Comms process to be run must exist. these are to be placed under the comms repo `config` folder, and named `testing_1.json5`, `testing_2.json5`, etc. at some point, code for automatically generating these node environment configurations could be written, but this is not a trivial task as it requires the generation of valid keys among other things.
- `CommsNode` implements a dictionary in the form of a class variable which stores the result of every `connect` call executed by any RIF comms client. this is a known bug; if these calls are not assigned the client & comms node will not behave correctly.
- `CommsProcess.stop` is a crude implementation for stopping all spawned processes. a better alternative would be asking the comms node to shut itself down rather than doing it at the OS level in this brute manner.
- a 5-second `sleep` is forced after starting any comms process. this is done to give the node time to finish booting up before calling operations on it. it should be replaced by a ping call or a synchronous boot-up command.
- nothing is being done with the RIF Comms process logs. these should ideally be written to files at some point in order to correctly debug a failing test.

## Technical details
- the original tests were moved to an `integration` folder, since they are not unit tests.
- ~~there are 2 tests which are expected to fail still, due to the unexpected behavior of the underlying `hasSubscriber` calls.~~ the failure has been fixed, although there are still documented pending issues related to this functionality.
- the `generate_address` method has been added back. ideally, i'd prefer no random addresses to be used on tests. however, since `n` comms nodes are booted up dynamically now, it's needed.
- `test_connect` and `test_disconnect` have been removed since they served no purpose.
- `json5` has been added as a dependency to the testing environment in order to parse environment configurations for the comms nodes. these are used to get the port for the comms API endpoints.

## Documentation
A PR to `lumino-docs` [has been submitted](https://github.com/anarancio/lumino-docs/pull/30) with updated information on programmatic testing.

## New and related issues
1. [RLP-897](https://jirainfuy.atlassian.net/browse/RLP-897): fix `connect` calls so that they don't need variable assignments 
2. [RLP-925](https://jirainfuy.atlassian.net/browse/RLP-925) & [RLP-926](https://jirainfuy.atlassian.net/browse/RLP-926): test `disconnect`
3. [RLP-936](https://jirainfuy.atlassian.net/browse/RLP-936): fix flaky tests
4. [RLP-937](https://jirainfuy.atlassian.net/browse/RLP-937): improve test coverage
5. [RLP-938](https://jirainfuy.atlassian.net/browse/RLP-938): generate node env config files for tests
6. [RLP-939](https://jirainfuy.atlassian.net/browse/RLP-939): add rif comms dependency
7. [RLP-940](https://jirainfuy.atlassian.net/browse/RLP-940): write comms test node logs to file
8. [RLP-941](https://jirainfuy.atlassian.net/browse/RLP-941): properly implement comms test nodes start-up
9. [RLP-942](https://jirainfuy.atlassian.net/browse/RLP-942): properly implement comms test nodes shutdown


raiden/tests/integration/network/transport/rif_comms/process.py
```diff
@@ -0,0 +1,50 @@
...
+import subprocess
+import time
+from pathlib import Path
+from subprocess import Popen
+
+import psutil
+
+# FIXME: this should be a dependency within the project
+# path to the RIF Comms Node repo
+RIF_COMMS_PATH = Path(__file__).parents[7].joinpath('rif-communications-pubsub-node')
```
JONAF2103, line 12: This is relative path? i didn't understood this but we should read that from an environment variable and if doesn't exists assume a relative or absolute path right?

mortelli: > This is relative path? i didn't understood this but we should read that from an environment variable and if doesn't exists assume a relative or absolute path right?

yes, it's a relative path. 

this means: "from the current file, move up a dir 7 times". this path matches the containing folder of the `lumino` repo, which _should_ be the same folder that contains the `rif-comms` repo (see _Limitations_ section).

i hope this clears it up.

raiden/tests/integration/network/transport/utils.py
```diff
@@ -0,0 +1,14 @@
+import secrets
+
+from coincurve import PublicKey
+from eth_typing import Address
+from eth_utils import to_canonical_address, to_checksum_address
+from sha3 import keccak_256
+
+
+def generate_address() -> Address:
```
JONAF2103, line 9: Awesome, maybe we could return the private key too to use this outside comms also.

JONAF2103 (approved):
LGTM in general, i put some comments there but is ok to me

raiden/tests/integration/network/transport/rif_comms/process.py
JONAF2103: So we never want to change lumino comms to another path? i'm just asking because maybe for deployment this could change and be customized that's all

mortelli: > So we never want to change lumino comms to another path? i'm just asking because maybe for deployment this could change and be customized that's all

of course. as it says in the `FIXME`, comms should be a test env dependency installed through `pip` (although i'm not 100% sure on how this would work in terms of building the project) and then located through environment variables.

raiden/tests/integration/network/transport/rif_comms/node.py
```diff
@@ -0,0 +1,67 @@
...
+    api_endpoint_prefix = "localhost"
+    env_file_prefix = "testing_"
+
+    def __init__(self, node_number: int):
+        """
+        Load and set a configuration attributes for a RIF Comms node.
+        A valid configuration file must exist at RIF_COMMS_PATH/config/testing_<node_number>.json5
+        """
+        self.env_name = self.env_file_prefix + str(node_number)
+        # TODO: generate these files if needed
```
mortelli, line 24: [RLP-938](https://jirainfuy.atlassian.net/browse/RLP-938)

raiden/tests/integration/network/transport/rif_comms/node.py
```diff
@@ -0,0 +1,67 @@
...
+
+        self.address = generate_address()
+        self.api_endpoint = self.api_endpoint_prefix + ":" + str(self._listening_port)
+
+
+class Node:
+    """
+    Class for RIF Comms node program.
+    """
+    connections = {}  # hack to get around the fact that each connect() call needs to be assigned
```
mortelli, line 40: [RLP-897](https://jirainfuy.atlassian.net/browse/RLP-897)

raiden/tests/integration/network/transport/rif_comms/node.py
```diff
@@ -0,0 +1,67 @@
...
+
+        self.client = RIFCommsClient(rsk_address=self.address, grpc_api_endpoint=self._api_endpoint)
+        self._process = self.start()
+
+    def start(self) -> Popen:
+        """
+        Start a RIF Comms node process and connect to it.
+        """
+        process = CommsProcess.start(env_name=self._env_name)
+        # FIXME: client.connect() calls should not need assignment
```
mortelli, line 55: [RLP-897](https://jirainfuy.atlassian.net/browse/RLP-897)

raiden/tests/integration/network/transport/rif_comms/node.py
```diff
@@ -0,0 +1,67 @@
...
+        self.connections[self.address] = self.client.connect()
+        return process
+
+    def stop(self):
+        """
+        Disconnect from RIF Comms node program, and stop its process.
+        """
+        try:
+            # FIXME: deleting entries in the connections dictionary is causing non-crashing thread exceptions
```
mortelli, line 64: if [RLP-897](https://jirainfuy.atlassian.net/browse/RLP-897) is solved, we won't need the `dict` in the first place

raiden/tests/integration/network/transport/rif_comms/process.py
```diff
@@ -0,0 +1,50 @@
+import os
+import signal
+import subprocess
+import time
+from pathlib import Path
+from subprocess import Popen
+
+import psutil
+
+# FIXME: this should be a dependency within the project
```
mortelli, line 10: [RLP-939](https://jirainfuy.atlassian.net/browse/RLP-939)

raiden/tests/integration/network/transport/rif_comms/process.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+
+    @staticmethod
+    def start(env_name: str) -> Popen:
+        """
+        Start a RIF Comms node process through npm, with the given environment name.
+        The RIF Comms Pub-Sub Bootnode repo must be at RIF_COMMS_PATH.
+        """
+        # TODO: look into using shell=False
+        # FIXME: write output to memory or disk
```
mortelli, line 27: [RLP-940](https://jirainfuy.atlassian.net/browse/RLP-940)

raiden/tests/integration/network/transport/rif_comms/process.py
```diff
@@ -0,0 +1,50 @@
...
+        process = subprocess.Popen(
+            "NODE_ENV=" + env_name + " npm run api-server",
+            cwd=RIF_COMMS_PATH,
+            shell=True,
+            preexec_fn=os.setsid,  # set to later kill process group
+        )
+
+        # FIXME: we need some sort of ping call, this will sometimes not be enough
```
mortelli, line 35: [RLP-941](https://jirainfuy.atlassian.net/browse/RLP-941)

raiden/tests/integration/network/transport/rif_comms/process.py
```diff
@@ -0,0 +1,50 @@
...
+        time.sleep(5)  # hack to get around calling the comms node before it is ready
+
+        return process
+
+    @staticmethod
+    def stop(process: Popen):
+        """
+        Kill a RIF Comms node process with the given process ID (PID),
+        """
+        # FIXME: find a better way to stop the comms node process
```
mortelli, line 45: [RLP-942](https://jirainfuy.atlassian.net/browse/RLP-942)

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -0,0 +1,123 @@
...
+
+@pytest.mark.parametrize("amount_of_nodes", [1])
+def test_locate_own_peer_id(comms_nodes):
+    comms_node = comms_nodes[1]
+
+    # attempt to locate self
+    assert comms_node.client._get_peer_id(comms_node.address) is not ""
+
+
+# FIXME: causes ERR_NO_PEERS_IN_ROUTING_TABLE in comms node although it passes
```
mortelli, line 33: documented in _RIF Comms: pending issues_

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -0,0 +1,123 @@
...
+@pytest.mark.parametrize("amount_of_nodes", [1])
+def test_locate_unregistered_peer_id(comms_nodes):
+    comms_node = comms_nodes[1]
+
+    # attempt to locate an unregistered peer
+    assert comms_node.client._get_peer_id(generate_address()) is ""
+
+
+# FIXME: comms node prints strange ServerUnaryCall message
```
mortelli, line 42: documented in _RIF Comms: pending issues_

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -0,0 +1,123 @@
...
+    assert client.is_subscribed_to(address) is True
+
+
+@pytest.mark.parametrize("amount_of_nodes", [2])
+def test_has_subscriber(comms_nodes):
+    comms_node_1, comms_node_2 = comms_nodes[1], comms_nodes[2]
+    client_1, address_1 = comms_node_1.client, comms_node_1.address
+    client_2, address_2 = comms_node_2.client, comms_node_2.address
+
+    # FIXME: nodes shouldn't have to subscribe to themselves for this to work
```
mortelli, line 60: documented in _RIF Comms: pending issues_

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -0,0 +1,123 @@
...
+    assert client_2.is_subscribed_to(address_2) is True
+
+
+@pytest.mark.parametrize("amount_of_nodes", [2])
+def test_two_clients_cross_subscription(comms_nodes):
+    comms_node_1, comms_node_2 = comms_nodes[1], comms_nodes[2]
+    client_1, address_1 = comms_node_1.client, comms_node_1.address
+    client_2, address_2 = comms_node_2.client, comms_node_2.address
+
+    # FIXME: nodes shouldn't have to subscribe to themselves for this to work
```
mortelli, line 80: documented in _RIF Comms: pending issues_

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -0,0 +1,123 @@
...
+    assert client_2.is_subscribed_to(address_2) is True
+
+
+@pytest.mark.parametrize("amount_of_nodes", [2])
+def test_two_clients_cross_messaging_same_topic(comms_nodes):
+    comms_node_1, comms_node_2 = comms_nodes[1], comms_nodes[2]
+    client_1, address_1 = comms_node_1.client, comms_node_1.address
+    client_2, address_2 = comms_node_2.client, comms_node_2.address
+
+    # FIXME: nodes shouldn't have to subscribe to themselves for this to work
```
mortelli, line 101: documented in _RIF Comms: pending issues_

raiden/tests/integration/network/transport/rif_comms/node.py
marcosmartinez7: does this implies any incorrect behaviour on the tests? 

or just throws some exceptions that doesnt affect the test resutls?

raiden/tests/integration/network/transport/rif_comms/test_client.py
marcosmartinez7: this causes the test to fail? or the assertion is correct?

raiden/tests/integration/network/transport/rif_comms/test_client.py
marcosmartinez7: that print is the result of the hasSubscriber grpc endpoint

i think is ok

marcosmartinez7 (approved):
Great work! 

Im ok with postpone all the TODOs here.

raiden/tests/integration/network/transport/rif_comms/node.py
mortelli: > does this implies any incorrect behaviour on the tests?
> 
> or just throws some exceptions that doesnt affect the test resutls?

i've seen some exceptions pop up in the case of deleting these entries during the execution of the teardown code, and they look like this:
> python grpcio 
> RuntimeError: cannot exit context: thread state references a different context object

since this occurs _after_ the tests are executed, it does not affect their result (be it success or failure). 

having said this, i would still rather not have these exceptions, so this is why i did not touch the dictionary at this point.

you can take a look at this for yourself by adding `del self.connections[self.address]` on line `65` and then running `pytest -s`

raiden/tests/integration/network/transport/rif_comms/test_client.py
mortelli: the test still succeeds. i suspect something funky is going on in the comms node: although it outputs the expected result, it might not have arrived at it through the correct logic.

i.e. "i have no peers in my routing table, therefore i cannot locate this random peer ID you're giving me", should actually be "i do not have this random peer ID in my routing table, so i cannot locate it"

raiden/tests/integration/network/transport/rif_comms/test_client.py
mortelli: it's a minor squabble on my part. this print is humanly unreadable, so i am not sure what purpose it serves

PR #172, opened 2020-12-08
RLP-932/add-subscription-check-to-payments

marcosmartinez7:
## Description

This pr modifies the rif comms node `send_message` in order to check for subscriptions on message sending.

## Changelog: 

- modified params for hasSubscriber call on rif comms client, now we check for `peerId=topic_id, channel=Channel(channelId=topic_id)` where `topic_id` = recipient peer id

- add is subscribed to call on send message

## Manually tested: 

p2p payment continue working

### DIFF
--- a/transport/rif_comms/client.py
+++ b/transport/rif_comms/client.py
@@ -62,11 +62,10 @@ class Client:
         :param rsk_address: RSK address which corresponds to the topic which is being checked for subscription
         :return: boolean value indicating whether the client is subscribed or not
         """
-        our_peer_id = self._get_peer_id(self.rsk_address.address)
-        topic_id = self._get_peer_id(to_checksum_address(rsk_address))
+        topic_id = self._get_peer_id(rsk_address)
         return self.stub.HasSubscriber(
             Subscriber(
-                peerId=our_peer_id,
+                peerId=topic_id,
                 channel=Channel(channelId=topic_id)
             )
         ).value
--- a/transport/rif_comms/node.py
+++ b/transport/rif_comms/node.py
@@ -240,7 +240,14 @@ class Node(TransportNode):
         """
         Send text message through the RIF Comms client.
         """
-        self._comms_client.subscribe_to(recipient)
+        self.log.info(
+            "RIF Comms sending message", message_payload=payload.replace("\n", "\\n"), recipient=pex(recipient)
+        )
+        if not self._comms_client.is_subscribed_to(recipient):
+            self.log.info(
+                "RIF Comms sending message not subscribed to recipient. Subscribing...", recipient=pex(recipient)
+            )
+            self._comms_client.subscribe_to(recipient)
         # send the message
         self._comms_client.send_message(payload, recipient)  # TODO: exception handling for RIF Comms client
         self.log.info(
--- a/transport/rif_comms/node.py
+++ b/transport/rif_comms/node.py
@@ -241,11 +241,11 @@ class Node(TransportNode):
         Send text message through the RIF Comms client.
         """
         self.log.info(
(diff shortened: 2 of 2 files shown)


transport/rif_comms/node.py
```diff
@@ -240,7 +240,14 @@ def send_message(self, payload: str, recipient: Address):
         """
         Send text message through the RIF Comms client.
         """
-        self._comms_client.subscribe_to(recipient)
+        self.log.info(
+            "RIF Comms sending message", message_payload=payload.replace("\n", "\\n"), recipient=pex(recipient)
+        )
+        if not self._comms_client.is_subscribed_to(recipient):
+            self.log.info(
+                "RIF Comms sending message not subscribed to recipient. Subscribing...", recipient=pex(recipient)
```
mortelli: log messages should not be capitalized and use as little punctuation as possible.

imo, they should also avoid mentioning specifics about rif comms node logic (such as: "i need to be subscribed from the comms node to send this message")
```suggestion
                "subscribing to recipient", transport="rif_comms", recipient=pex(recipient)
```

transport/rif_comms/node.py
```diff
@@ -240,7 +240,14 @@ def send_message(self, payload: str, recipient: Address):
         """
         Send text message through the RIF Comms client.
         """
-        self._comms_client.subscribe_to(recipient)
+        self.log.info(
+            "RIF Comms sending message", message_payload=payload.replace("\n", "\\n"), recipient=pex(recipient)
```
mortelli: per [my other comment](https://github.com/rsksmart/lumino/pull/172#discussion_r538736556):
```suggestion
            "sending message", message_payload=payload.replace("\n", "\\n"), transport="rif_comms", recipient=pex(recipient)
```

transport/rif_comms/client.py
```diff
@@ -62,11 +62,10 @@ def is_subscribed_to(self, rsk_address: Address) -> bool:
...
         """
-        our_peer_id = self._get_peer_id(self.rsk_address.address)
-        topic_id = self._get_peer_id(to_checksum_address(rsk_address))
+        topic_id = self._get_peer_id(rsk_address)
         return self.stub.HasSubscriber(
             Subscriber(
-                peerId=our_peer_id,
+                peerId=topic_id,
                 channel=Channel(channelId=topic_id)
             )
```
mortelli, line 70: i still don't understand this.

i interpret the previous code as:

> check if `our_peer_id` is subscribed to this `topic_id`

where:
- `our_peer_id` is the result of looking up our own RSK address in the DHT
- `topic_id` is the result of looking up the destination RSK address in the DHT

this made sense to me :man_shrugging: 

now, this looks like: 
> check if the destination peer is subscribed to this `channelId` 

because:
- `peerId` is the result of looking up the **recipient** RSK address in the DHT
- `channelId` is also the same as the previous point, which is confusing :thinking: 

can you please explain this further?

mortelli (commented):
i understand the comms code has changed, but i am not yet convinced about the new logic for `is_subscribed_to`

transport/rif_comms/client.py
marcosmartinez7: AFAIK even its redundant to call with the same parameter twice, maybe on other scenarios you want to check if another peer is subscribed to the same channel

on this scenario, is the node behind the client that wants to check if it is subscribed to the peer id of the partner on that same channel

not too happy with this

mortelli: ok, i suppose this is good enough for now

mortelli (approved):
i expect this to be iterated upon further

PR #173, opened 2020-12-08
RLP-793 Fixing send signed close tx error

JONAF2103:
### Problem

When we had a pending close transaction and we pull down the node, the next time we raise lumino again it fails trying to execute the operation again.

### Cause

When we start lumino a process get's the latest snapshot from database to populate the information on the chain_state. There we have a list of pending_transactions. That list contains the pending to send transactions (like close channel, open channel, etc). So if those were not sent out already they will. The problem with this is that for close channel it was failing since it wasn't checking if the operation was still pending after the time the node was down. Basically we had a bad check for pending transactions after being added to the list.

### Solution

Fix the pending transaction check before adding the transaction to the list to be processed.

### TODO

* Run Tests :heavy_check_mark:, [here](https://github.com/rsksmart/lumino/files/5675867/test.results.log) the results compared to [this](https://github.com/anarancio/lumino-docs/blob/master/pytest/results/develop/2020-11-06/summary.txt) base line.

### DIFF
--- a/raiden/transfer/node.py
+++ b/raiden/transfer/node.py
@@ -1148,7 +1148,13 @@ def is_transaction_effect_satisfied(
         and isinstance(transaction, ContractSendChannelClose)
         and state_change.token_network_identifier == transaction.token_network_identifier
         and state_change.channel_identifier == transaction.channel_identifier
+    ) or (
+        isinstance(state_change, ContractReceiveChannelClosedLight)
+        and isinstance(transaction, ContractSendChannelClose)
+        and state_change.token_network_identifier == transaction.token_network_identifier
+        and state_change.channel_identifier == transaction.channel_identifier
     )
+
     if is_valid_close:
         return True
 
@@ -1158,6 +1164,7 @@ def is_transaction_effect_satisfied(
         and state_change.token_network_identifier == transaction.token_network_identifier
         and state_change.channel_identifier == transaction.channel_identifier
     )
+
     if is_valid_settle:
         return True
 
@@ -1309,7 +1316,8 @@ def update_queues(iteration: TransitionResult[ChainState], state_change: StateCh
             queue.append(event)
 
         if isinstance(event, ContractSendEvent):
-            chain_state.pending_transactions.append(event)
+            if is_transaction_pending(chain_state, event, state_change):
+                chain_state.pending_transactions.append(event)
 
 
 def state_transition(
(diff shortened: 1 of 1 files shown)


raiden/transfer/node.py
```diff
@@ -1148,7 +1148,13 @@ def is_transaction_effect_satisfied(
         and isinstance(transaction, ContractSendChannelClose)
         and state_change.token_network_identifier == transaction.token_network_identifier
         and state_change.channel_identifier == transaction.channel_identifier
+    ) or (
```
mortelli: why not place an implicit `or` on line `1147`?

```python
isinstance(state_change, (ContractReceiveChannelClosed, ContractReceiveChannelClosedLight))
and isinstance(transaction, ContractSendChannelClose)
and state_change.token_network_identifier == transaction.token_network_identifier
and state_change.channel_identifier == transaction.channel_identifier
```

raiden/transfer/node.py
```diff
@@ -1309,7 +1316,8 @@ def update_queues(iteration: TransitionResult[ChainState], state_change: StateCh
             queue.append(event)
 
         if isinstance(event, ContractSendEvent):
-            chain_state.pending_transactions.append(event)
+            if is_transaction_pending(chain_state, event, state_change):
+                chain_state.pending_transactions.append(event)
```
mortelli, line 1316: this doesn't seem specific to "send close".

could you please explain why this wasn't affecting other types of transactions?

marcosmartinez7 (commented):
im not completely sure about the check of if `is_transaction_pending` 

the possible reasons of why is this needed:

- raiden bug
- all the tx from a snapshot (open, close for full nodes, etc) are being processed incorrectly too (we could test that if neccessary)

i think we can go with this anyway, but, maybe the unit tests detect something, did you run them?

raiden/transfer/node.py
JONAF2103: You are right

raiden/transfer/node.py
JONAF2103: Other transactions don't throw exceptions that pull down the node so that's why you don't see any error but wrong anyway

JONAF2103:
@marcosmartinez7 i agree maybe in the future we can improve this or maybe remove it if we see that is not necesary

marcosmartinez7 (approved):
lgtm

raiden/transfer/node.py
mortelli: feel free to push the change or leave as-is, this isn't a blocker for me

raiden/transfer/node.py
mortelli: thanks

mortelli (approved):
master bug catcher :star2:

marcosmartinez7 (approved):
lgtm

PR #174, opened 2020-12-08
Feature/rlp 859/rif comms onboarding impl

marcosmartinez7:
## Description 

This PR implements the onborading for rif comms. It must be merged with https://github.com/rsksmart/lumino-light-client-sdk/pull/105

## Changelog

- `save_light_client` on `python.py` was split into two different methods according to the transport mode
- client table has now nullable fields (password ,display_name, seed_retry) since they are not used by rif comms
- `light_client_onboarding_data` now returns a `transport_mode` field in order to give information to the SDK of which transport implementation the hub is using
- implemented `register_light_client` and `light_client_onboarding_data` for `RIFCommsLayer`
- both RIFCommsLayer and MatrixLayer now have a transport_type class attribute, used in `light_client_onboarding_data`

## Manually tested

- onboarding for new light client for matrix works
- same for rif comms
- onboarding for an already registered client works for both transports


raiden/api/python.py
```diff
@@ -1764,6 +1765,38 @@ def save_light_client(self,
 
         return result
 
+    def store_rif_comms_light_client(self, address):
```
JONAF2103: I think this one and the `store_matrix_light_client` method are the same code but you just put none on some of the parameters to save, maybe you could put all this code in one method only and then put the parameters like optional? so you don't repeat the code just for a few params.

raiden/storage/utils.py
```diff
@@ -88,9 +88,9 @@ def __getattr__(self, item):
 DB_CREATE_CLIENT = """
 CREATE TABLE IF NOT EXISTS client (
```
JONAF2103, line 89: We could put the type of client here maybe, it could be useful if you want to query data from database and know which type of transport is using that client. Just a thought

JONAF2103 (approved):
LGTM in general, i put just some comments there of improvements that you could do but it's ok if you don't

raiden/api/python.py
```diff
@@ -1711,12 +1711,13 @@ def get_all_light_clients(self):
...
-                          signed_password,
-                          server_name,
-                          signed_display_name,
-                          signed_seed_retry):
+    def store_matrix_light_client(self,
+                                  address,
+                                  signed_password=None,
+                                  server_name=None,
+                                  signed_display_name=None,
+                                  signed_seed_retry=None,
```
mortelli: why have these default values been added? 

can the method actually work if these are `None`? will they not cause a crash when calling `decode` on them?

raiden/api/python.py
```diff
@@ -1711,12 +1711,13 @@ def get_all_light_clients(self):
         light_clients = self.raiden.wal.storage.get_all_light_clients()
         return light_clients
 
-    def save_light_client(self,
-                          address,
-                          signed_password,
-                          server_name,
-                          signed_display_name,
-                          signed_seed_retry):
+    def store_matrix_light_client(self,
```
mortelli: could this not be a method for the `MatrixLayer` class?

raiden/api/python.py
mortelli: could this not be a method for the `RIFCommsLayer` class?

raiden/storage/utils.py
marcosmartinez7: I have thought about that but i could find any situation in which you want to know the transport, since the transport mode is the transport mode that the hub is running

raiden/api/python.py
marcosmartinez7: The matrix one uses a lot of matrix server related variables, i.e

```
 pubhex = self.raiden.config["pubkey"].hex()
 encrypt_signed_password = encrypt(pubhex, signed_password.encode())
 encrypt_signed_display_name = encrypt(pubhex, signed_display_name.encode())
 encrypt_signed_seed_retry = encrypt(pubhex, signed_seed_retry.encode())
```

but.. i dont like the idea of passing None, i think i can improve that

transport/matrix/layer.py
```diff
@@ -183,3 +184,5 @@ def register_light_client(self, raiden_api: RaidenAPI, registration_data: dict)
         return light_client
 
 
+
+
```
mortelli:
```suggestion
```

transport/matrix/layer.py
```diff
@@ -115,14 +116,14 @@ def light_client_onboarding_data(self, address: typing.Address) -> dict:
         server_url = client.api.base_url
         server_name = urlparse(server_url).netloc
         return {
+            "transport_mode": self.transport_type,
```
mortelli: i wish we'd call this `transport_type`. we already have transport _layer_, and the transport classes have `type` defined as well—and this matches the config code.

i think adding _mode_ to this list of terms might add noise.

transport/rif_comms/layer.py
```diff
@@ -44,7 +46,27 @@ def construct_light_clients_nodes(self, config):
             sys.exit(1)
 
     def light_client_onboarding_data(self, address: Address) -> dict:
-        pass
+        return {
+            "transport_mode": self.transport_type,
```
mortelli: same here

transport/rif_comms/layer.py
```diff
@@ -44,7 +46,27 @@ def construct_light_clients_nodes(self, config):
...
+        }
 
     def register_light_client(self, raiden_api: 'RaidenAPI', registration_data: dict) -> TransportNode:
-        pass
+        config = raiden_api.raiden.config["transport"]["rif_comms"]
+        address = bytearray.fromhex(remove_0x_prefix(registration_data['address']))
```
mortelli: i think it's better to explicitly call this `lc_address` to avoid confusion

transport/rif_comms/layer.py
```diff
@@ -44,7 +46,27 @@ def construct_light_clients_nodes(self, config):
...
+
+        light_client = raiden_api.store_rif_comms_light_client(
+            address
+        )
```
mortelli:
```suggestion
        light_client = raiden_api.store_rif_comms_light_client(address)
```

transport/rif_comms/layer.py
```diff
@@ -44,7 +46,27 @@ def construct_light_clients_nodes(self, config):
             sys.exit(1)
 
     def light_client_onboarding_data(self, address: Address) -> dict:
-        pass
+        return {
+            "transport_mode": self.transport_type,
+        }
 
     def register_light_client(self, raiden_api: 'RaidenAPI', registration_data: dict) -> TransportNode:
```
mortelli: something isn't right with the types hints here. the abstract transport layer has a matching signature for this method, but a `TransportNode` is not what's being returned.

one of these 2 is wrong

mortelli: this applies to the corresponding matrix method as well

mortelli (commented):
i am happy to hear that this is working :+1: 

my 2 main issues here are:
1. language: i would avoid the term `mode` since we already have `layer` and `type`
2. responsibilities/segregation: each of the storing methods for light clients naturally belong in their respective concrete transport layer classes. i do not see much of a reason to keep them in the `RaidenAPI` class when they are specific to each implementation. it also does not take advantage of the current design we're working with in terms of abstractions.

regarding n°2: i suspect i might be missing something here, so i'd appreciate your comments on this.

in any case, i don't expect these to cause _too_ many problems, so afaic we're almost ready to merge

raiden/api/python.py
marcosmartinez7: it could, but i dont think the layer is related with storing data on the db. 

I think that we could move this to light_client_service that is in charge of those kind of operations

but i prefeer to do that on a future refactor

raiden/api/python.py
marcosmartinez7: i will test that

raiden/api/python.py
marcosmartinez7: same as the matrix layer comment

transport/matrix/layer.py
marcosmartinez7: agree, i will do it

raiden/api/python.py
marcosmartinez7: it crashes, i will keep it like this since i dont want to modify the sqlite function right now

transport/matrix/layer.py
marcosmartinez7: done in next commit

transport/rif_comms/layer.py
marcosmartinez7: done in next commit

transport/rif_comms/layer.py
marcosmartinez7: done in next commit

raiden/api/python.py
mortelli: but if it crashes with `None`, then why do you want to keep it like this?

transport/rif_comms/layer.py
marcosmartinez7: removed

raiden/api/python.py
marcosmartinez7: sorry, i was confused, now its fixed

raiden/api/python.py
marcosmartinez7: moved to lc service

raiden/api/python.py
marcosmartinez7: moved to lc service

transport/rif_comms/layer.py
marcosmartinez7: removed

raiden/storage/utils.py
JONAF2103: Make sense

raiden/api/python.py
JONAF2103: Yeah what i suggest was that on those ocurrences just do a check and if it's not None then you do that otherwise you just pass the value from the parameter directly to the save method, i just don't like the duplicated code but anyway is just a detail.

JONAF2103 (approved):
LGTM

mortelli (approved):
while i _do_ think it's better that the LC storing method is on the LC service now (rather than in the raiden API class), i am not a big fan of having 2 separate methods for storing the same data on the same table. 

reasons for this are:
1. it goes against the abstractions and generalizations in place. this is purely on a programmatic and design level; we know that the number of transport layers will not change. but if it _did_, we'd have to change the amount of methods for storing LCs on the service
2. it duplicates code
3. it does not follow SRP so much (2 functions for the same responsibility)

for this reason, it seems to me a single storing method would be better suited for this. 

i still think this is good enough, so i am approving :+1: 

--- 

for discussion's sake however, here's how i would go about it:

```python
@staticmethod
def store_light_client(
    address,
    storage,
    pubhex,
    signed_password=None,
    server_name=None,
    signed_display_name=None,
    signed_seed_retry=None,
):
    def encrypt_auth_param(param):
        return encrypt(pubhex, param.encode()).hex() if param else None

    # merges two dictionaries skipping keys with None values
    def merge_response(mandatory_response: dict, optional_response: dict) -> dict:
        return {
            **mandatory_response,
            **{k: v for k, v in optional_response.items() if v is not None},
        }

    checksum_addr = to_checksum_address(address)
    api_key = hexlify(os.urandom(20)).decode("utf-8")

    light_client = storage.get_light_client(checksum_addr)

    auth_params = {
        "encrypt_signed_password": encrypt_auth_param(signed_password),
        "encrypt_signed_display_name": encrypt_auth_param(signed_display_name),
        "encrypt_signed_seed_retry": encrypt_auth_param(signed_seed_retry),
    }

    if light_client is None:
        result = storage.save_light_client(
            api_key=api_key,
            address=checksum_addr,
            encrypt_signed_password=auth_params["encrypt_signed_password"],
            encrypt_signed_display_name=auth_params["encrypt_signed_display_name"],
            encrypt_signed_seed_retry=auth_params["encrypt_signed_seed_retry"],
            current_server_name=server_name,
            pending_for_deletion=0,
        )
        if result > 0:
            return merge_response(
                {
                    "address": checksum_addr,
                    "api_key": api_key,
                    "message": "successfully registered",
                    "result_code": 200,
                },
                auth_params,
            )
        else:
            return {"message": "An unexpected error has occurred.", "result_code": 500}
    else:
        return merge_response(
            {
                "address": checksum_addr,
                "api_key": light_client["api_key"],
                "message": "Already registered",
                "result_code": 409,
            },
            auth_params,
        )
```
Pros:
- the call is now a `@staticmethod` since it does not use `cls`
- does not redefine variables (`var = something(var)`) which is not great practice imo
- it's 1 function only for storing any LC. avoids duplicated code and is more aligned with OCP

Cons:
- slightly more complicated code (2 embedded functions); these could be defined elsewhere, though

Other:
- `None` default params are not so cool, but this is already happening in the `save_light_client` method anyway

---

note: would also means changing the layer calls like so:

```python
light_client = LightClientService.store_light_client(
    lc_address, raiden_api.raiden.wal.storage, raiden_api.raiden.config["pubkey"].hex()
)
```

```python
light_client = LightClientService.store_light_client(
    lc_address,
    raiden_api.raiden.wal.storage,
    raiden_api.raiden.config["pubkey"].hex(),
    signed_password,
    password,
    signed_display_name,
    signed_seed_retry,
)
```

PR #178, opened 2020-12-11
RLP-949 Mediated payments fix

JONAF2103:
### Problem
Can't make a mediated payment after reseting the database of the lumino node

### Escenario
Any escenario of mediated payments, but for this bug we test it with
FN1 -> FN2 -> FN3

### Steps to Reproduce

1. Setup the topology
2. Pull down the lumino nodes
3. Clear the lumino databases
4. Start the lumino nodes again
5. Send a mediated payment from FN1 to FN3 through the FN2

### Result
The payment fails because it doesn't have any route available

### Solution
The solution is to remove the filtering function for reachablity on the routes when we are mediating a transfer. That function causes the problem since it uses a map that it doesn't have the updated values due to the reachability issues that we have. So for now we are removing this function and the test for that too.

### Changes

* Removing filter_reachable_routes and the test for that since it was causing the problem of reachability on the mediated payment.

### Scope

This affects the mediated payment flow

### DIFF
--- a/raiden/tests/unit/transfer/mediated_transfer/test_mediatorstate.py
+++ b/raiden/tests/unit/transfer/mediated_transfer/test_mediatorstate.py
@@ -837,7 +837,6 @@ def test_mediate_transfer():
         channels.get_routes_by_index([1]),
         channels.get_sub_channel(0),
         channels.sub_channel_map,
-        channels.nodeaddresses_to_networkstates,
         pseudo_random_generator,
         payer_transfer,
         block_number,
@@ -1266,7 +1265,6 @@ def test_payee_timeout_must_be_equal_to_payer_timeout():
         channel_set.get_routes_by_index([1]),
         channel_set.get_sub_channel(0),
         channel_set.sub_channel_map,
-        channel_set.nodeaddresses_to_networkstates,
         pseudo_random_generator,
         payer_transfer,
         block_number,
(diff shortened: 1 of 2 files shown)


marcosmartinez7 (approved):
lgtm

mortelli (commented):
>The solution is to remove the filtering function for reachablity on the routes when we are mediating a transfer.

i am very wary of removing entire functions as a solution to a problem, so i have a couple of questions.

> That function causes the problem since it uses a map that it doesn't have the updated values due to the reachability issues that we have. 

1. what are "the reachability issues that we have"?
2. we've successfully done mediated payments many times before, and this function was still around. how come this is a problem only now?

> So for now we are removing this function and the test for that too.

3. do we have a plan to add this back at some point?

JONAF2103:
> > The solution is to remove the filtering function for reachablity on the routes when we are mediating a transfer.
> 
> i am very wary of removing entire functions as a solution to a problem, so i have a couple of questions.
> 
> > That function causes the problem since it uses a map that it doesn't have the updated values due to the reachability issues that we have.
> 
>     1. what are "the reachability issues that we have"?
> 
>     2. we've successfully done mediated payments many times before, and this function was still around. how come this is a problem only now?
> 
> 
> > So for now we are removing this function and the test for that too.
> 
>     1. do we have a plan to add this back at some point?

I answer by point:

1. The issues that we have is that the reachability state is not being updated properly so we can't trust on those statuses
2. Yes that's true, but before we had also another code that was causing some other performance issue and we removed that, the problem was solved but that cause this problem and that's why we are deleting this code.

1. I think we are going to use rif-comms in the future so this is not going to be necessary, but i think we want to keep matrix as legacy so probably we will need to at least make sure that these reachability problems are not causing other bugs. Also i think we are going to add reachability to rif-comms in some point but i'm not sure.

I hope this clear your doubts.

mortelli (approved):
thanks @JONAF2103 for clearing this up

PR #180, opened 2020-12-14
RLP-951 Concurrent deposit for lc

JONAF2103:
### Introduction
In light mode we can't make multiple deposit operations from different LC's to one partner concurrently, all those operations are being secuentially executed instead.

### Change
We are changing the way we lock those operations to use the channel identifier instead of the partner address to avoid having to wait for each client. To do this we have to modify token network proxy to put the channel identifier as the key for the lock map.

This affects concurrent deposit, close and settlement operations.

### Pending

1. Manual testing :heavy_check_mark:
2. Run baseline :heavy_check_mark: 

[Here](https://github.com/rsksmart/lumino/files/5704447/test.results.log) the test results, everything is ok compared to [this](https://github.com/anarancio/lumino-docs/blob/master/testing/pytest/results/develop/2020-11-06/summary.txt) baseline.

### DIFF
--- a/raiden/exceptions.py
+++ b/raiden/exceptions.py
@@ -310,11 +310,9 @@ class ProxyTransactionError(RaidenError):
     """ Raised when an operation is sent out to the blockchain and it returns an error and we need to handle it """
 
     def __init__(self,
-                 tx_error_prefix: Optional[str],
-                 tx_error: Optional[Any],
-                 tx_gas_limit: Optional[int]):
-        super().__init__()
+                 tx_error_prefix: str,
+                 tx_error: Optional[Any]):
+        super(RaidenError, self).__init__(f"A proxy transaction error has occurred: {tx_error_prefix}")
         self.tx_error_prefix = tx_error_prefix
         self.tx_error = tx_error
-        self.tx_gas_limit = tx_gas_limit
 
--- a/raiden/exceptions.py
+++ b/raiden/exceptions.py
@@ -312,7 +312,7 @@ class ProxyTransactionError(RaidenError):
     def __init__(self,
                  tx_error_prefix: str,
                  tx_error: Optional[Any]):
-        super(RaidenError, self).__init__(f"A proxy transaction error has occurred: {tx_error_prefix}")
+        super(ProxyTransactionError, self).__init__(f"A proxy transaction error has occurred: {tx_error_prefix}")
         self.tx_error_prefix = tx_error_prefix
         self.tx_error = tx_error
 
--- a/raiden/exceptions.py
+++ b/raiden/exceptions.py
@@ -305,3 +305,16 @@ class RawTransactionFailed(RaidenError):
 class UnhandledLightClient(RaidenRecoverableError):
     """Raised if someone tries to create a channel using this node as a hub and the light clients are not registered."""
 
+
+class ProxyTransactionError(RaidenError):
+    """ Raised when an operation is sent out to the blockchain and it returns an error and we need to handle it """
+
+    def __init__(self,
+                 tx_error_prefix: Optional[str],
+                 tx_error: Optional[Any],
+                 tx_gas_limit: Optional[int]):
+        super().__init__()
+        self.tx_error_prefix = tx_error_prefix
+        self.tx_error = tx_error
+        self.tx_gas_limit = tx_gas_limit
+
--- a/raiden/network/proxies/token_network.py
+++ b/raiden/network/proxies/token_network.py
@@ -142,9 +142,8 @@ class TokenNetwork:
         self.node_address = self.client.address
         self.open_channel_transactions: Dict[OpenChannelTrKey, AsyncResult] = dict()
 
-        # Forbids concurrent operations on the same channel
-
-        self.channel_operations_lock: Dict[Address, RLock] = defaultdict(RLock)
+        # Forbids concurrent operations on the same channel on the same participant
+        self.channel_operations_lock: Dict[ChannelID, RLock] = defaultdict(RLock)
 
         # Serializes concurent deposits on this token network. This must be an
         # exclusive lock, since we need to coordinate the approve and
@@ -161,6 +160,17 @@ class TokenNetwork:
             raise RuntimeError(f"Call to '{function_name}' returned nothing")
 
         return call_result
+
+    def lock_and_execute_channel_operation(self,
+                                           channel_identifier: ChannelID,
+                                           semaphore: Optional[Semaphore],
+                                           code_to_block) -> Optional[Any]:
+        if semaphore:
+            with self.channel_operations_lock[channel_identifier], semaphore:
+                return code_to_block()
+        else:
+            with self.channel_operations_lock[channel_identifier]:
+                return code_to_block()
 
     def token_address(self) -> Address:
         """ Return the token of this manager. """
@@ -723,7 +733,8 @@ class TokenNetwork:
             contract_manager=self.contract_manager,
         )
         checking_block = self.client.get_checking_block()
-        with self.channel_operations_lock[partner], self.deposit_lock:
+
+        def make_deposit():
             previous_total_deposit = self._detail_participant(
                 channel_identifier=channel_identifier,
                 participant=creator,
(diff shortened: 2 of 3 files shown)


raiden/network/proxies/token_network.py
```diff
@@ -1095,24 +1116,22 @@ def close_light(
...
 
             if channel_onchain_detail.state != ChannelState.OPENED:
-                msg = (
+                raise RaidenUnrecoverableError((
                     f"The channel was not open at the provided block "
                     f"({given_block_identifier}). This call should never have "
                     f"been attempted."
-                )
-                raise RaidenUnrecoverableError(msg)
+                ))
```
mortelli: good change, but it looks like there are redundant parentheses

raiden/network/proxies/token_network.py
```diff
@@ -162,6 +163,17 @@ def _call_and_check_result(
...
+    def lock_and_execute_channel_operation(self,
+                                           channel_identifier: ChannelID,
+                                           semaphore: Optional[Semaphore],
+                                           code_to_block) -> Optional[Any]:
+        if semaphore:
+            with self.channel_operations_lock[channel_identifier], semaphore:
+                return code_to_block()
+        else:
+            with self.channel_operations_lock[channel_identifier]:
+                return code_to_block()
```
mortelli: i don't like this `if` so much, but i don't see a very elegant alternative; i can only think of making `semaphore` non optional, and then passing a new `RLock` every time the function is called from outside in the cases where originally `semaphore is None` (this new `RLock` would always be free to use).

i say leave as is

raiden/network/proxies/token_network.py
```diff
@@ -1206,21 +1211,23 @@ def close_light(
...
+                    ))
 
                 if detail.state >= ChannelState.CLOSED:
-                    msg = (
+                    raise RaidenRecoverableError((
                         f"cannot call close on a channel that has been closed already. "
                         f"current_state={detail.state}"
-                    )
-                    raise RaidenRecoverableError(msg)
+                    ))
```
mortelli: `((...))`

raiden/network/proxies/token_network.py
```diff
@@ -1160,13 +1167,12 @@ def close_light(
...
                     if receipt_or_none["cumulativeGasUsed"] == gas_limit:
-                        msg = (
+                        raise RaidenUnrecoverableError((
                             "update transfer failed and all gas was used. Estimate gas "
                             "may have underestimated update transfer, or succeeded even "
                             "though an assert is triggered, or the smart contract code "
                             "has an conditional assert."
-                        )
-                        raise RaidenUnrecoverableError(msg)
+                        ))
```
mortelli: and here

raiden/network/proxies/token_network.py
```diff
@@ -1126,18 +1145,6 @@ def close_light(
...
-                #     safe_gas_limit(gas_limit, GAS_REQUIRED_FOR_CLOSE_CHANNEL),
-                #     channel_identifier=channel_identifier,
-                #     partner=partner,
-                #     balance_hash=balance_hash,
-                #     nonce=nonce,
-                #     additional_hash=additional_hash,
-                #     signature=signature,
-                # )
-
-                print("Sending signed_close_tx")
```
mortelli, line 1140: _GOOD_

raiden/network/proxies/token_network.py
mortelli: ok scratch that, i have an idea.

what if we remove `semaphore` from the params, and then do this:

```python
def lock_and_execute_channel_operation(self, channel_identifier: ChannelID, code_to_block) -> Optional[Any]:
        with self.channel_operations_lock[channel_identifier]:
            return code_to_block()
```

and then:

```python
def make_deposit():
    with self.deposit_lock:
        # rest of code
```

mortelli (approved):
solid work, looking forward to the test results

raiden/network/proxies/token_network.py
JONAF2103: Yup i will remove them, thanks

raiden/network/proxies/token_network.py
JONAF2103: Got it i will try it

mortelli (commented):
a couple more comments:

### docstrings
please add a docstring to the new `lock_and_execute_channel_operation` function.

### function structure
i don't think it's great style to use `return` inside `lock_and_execute_channel_operation` when the code to be executed is arguably _anything_, and in fact for most cases there's nothing to return. 

this can lead to `TypeError` exceptions when (for example) trying to unpack a `None` result into multiple (or a less-than-needed amount of) variables, and in general is just untidy as well as confusing.

(this is also the reason for the LGTM complaint)

i would either:
1. make all functions return something meaningful (an unconditional `return None` would not fit this criteria), or 
2. make all functions just execute without returning anything.

the first one implies too many changes to the current code, so i would go with the second one.

this raises the question of how to extract values from the result of the callables to be executed.

as far as i can see, these so-called results are always "errors or nothing", so i would go about doing this through the use of exceptions.

so let me just say first: **i recommend we leave this as it is,** since this code is quite critical, and we could go on changing the entire architecture of this module if we're being picky.

---

_having said that,_ here's how i would do it:

```python
def lock_and_execute_channel_operation(self, channel_id: ChannelID, blocking_operation: Callable):
    with self.channel_operations_lock[channel_id]:
        blocking_operation()
```

```python
def settle():  # omitting params
    
    # ...func code...
    
    def make_settlement():
        tx_gas_limit = self.proxy.estimate_gas(
            checking_block, "settleChannel", channel_identifier=channel_identifier, **kwargs
        )

        if tx_gas_limit:
            tx_gas_limit = safe_gas_limit(tx_gas_limit, GAS_REQUIRED_FOR_SETTLE_CHANNEL)

            transaction_hash = self.proxy.transact(
                "settleChannel", gas_limit, channel_identifier=channel_identifier, **kwargs
            )
            self.client.poll(transaction_hash)
            tx_error = check_transaction_threw(self.client, transaction_hash)

            if tx_error:
                raise TransactionFailure(error_prefix="call to settle has failed", tx_error=tx_error)
        else:
            raise TransactionFailure(error_prefix="Call to settle will fail", tx_error=None)
        
    try:
        self.lock_and_execute_channel_operation(
            channel_id=channel_identifier,
            blocking_operation=make_settlement
        )
    except TransactionFailure as e:
        self.handle_transaction_error_for_settlement(
            channel_identifier=channel_identifier,
            checking_block=checking_block,
            error_prefix=str(e),
            log_details=log_details,
            partner=partner,
            transaction_error=e.tx_error,
        )

    log.info("settle successful", **log_details)
```
with a custom exception defined as such:
```python
class TransactionFailure(Exception):
    def __init__(self, error_prefix: str, tx_error):
        super(Exception, self).__init__(error_prefix)
        self.error_prefix = error_prefix
        self.tx_error = tx_error

    def __str__(self):
        return self.error_prefix
```

and something analogous to the light client version of this functionality.

this would allow you to avoid needless `None` returns, plus having a cleaner code for the `make_settlement` function as an added bonus.

disclaimer: in this example i've left `check_transaction_threw` as it is, but that one is a function with _very_ problematic syntax on which i would also like to iterate (someday).

marcosmartinez7 (approved):
lgtm

raiden/exceptions.py
```diff
@@ -305,3 +305,16 @@ class RawTransactionFailed(RaidenError):
...
+    """ Raised when an operation is sent out to the blockchain and it returns an error and we need to handle it """
+
+    def __init__(self,
+                 tx_error_prefix: Optional[str],
+                 tx_error: Optional[Any],
+                 tx_gas_limit: Optional[int]):
+        super().__init__()
+        self.tx_error_prefix = tx_error_prefix
+        self.tx_error = tx_error
+        self.tx_gas_limit = tx_gas_limit
```
mortelli: - `super()` should be `super(RaidenError, self)` (it's _always_ better to specify which concrete parent class you are calling due to possible multiple inheritance)
- there is no need for the `tx_gas_limit` field, since this field is not read when the exception is raised.
- since the `RaidenError` parent class does not override the `__init__` method and it inherits from `Exception`, you should pass a `str` as an argument to the `super` call, as python exceptions should be initialized with an exception message.
  - `tx_error_prefix` is an acceptable argument here. it can also be something fancier than that, if you wish
- `tx_error_prefix` should not be optional, an exception without a message is unhelpful.

the rest looks good :+1:

mortelli: @JONAF2103 whoops, made a mistake in my code. check LGTM results (my bad!)

JONAF2103: Done

PR #182, opened 2020-12-15
RLP-918: update rif-comms bindings

mortelli:
This PR tackles [RLP-918](https://jirainfuy.atlassian.net/browse/RLP-918), which is about updating the RIF Comms transport layer implementation with the latest pub-sub release (and its proto) and testing it.

## Overview
Since the pub-sub node is being actively worked on, the [`grpc-api` branch](https://github.com/rsksmart/rif-communications-pubsub-bootnode/tree/grpc-api) was used to test the client. [Subsequent updates](https://jirainfuy.atlassian.net/browse/RLP-956) are expected at a later time.

The main changes are related to the new signatures for sending messages, checking subscriptions and unsubscribing from topics.

Besides updating the `proto` package, the RIF Comms client integration tests were iterated upon as a means of verifying the correctness of its behaviour.

On top of this, manual tests with 4 full nodes were verified to work, including a case of refund.

## Details
- the RIF Comms client `_get_peer_id` function was removed as it was no longer used.
  - as a consequence, its integration tests were also removed.
- one of the new tests is currently skipped, as it hangs. the test attempts to send a message to an unregistered address.
- the entire suite was executed dozens of times. [flakiness](https://jirainfuy.atlassian.net/browse/RLP-936) is occasionally observed _only_ in the tear-down process. it appears [connecting](https://jirainfuy.atlassian.net/browse/RLP-897) and [disconnecting](https://jirainfuy.atlassian.net/browse/RLP-925) are the remaining problematic functions in the client.
  - due to this, tests for these functions are pending (see [here](https://jirainfuy.atlassian.net/browse/RLP-957) and [here](https://jirainfuy.atlassian.net/browse/RLP-926)). the test suite can be [iterated further](https://jirainfuy.atlassian.net/browse/RLP-937) in addition to these. 
  - the result of `EndCommunications` was also assigned since it seems to help with this.
- [tests for multiple connections](https://jirainfuy.atlassian.net/browse/RLP-955) are also pending. this is expected to be worked on through #185 and subsequent PRs.

## New dependencies
- [`pytest-repeat`](https://pypi.org/project/pytest-repeat/) was added to the testing environment dependencies in order to easily execute a single or multiple tests `n` times in a row. this is useful for checking flakiness.


raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -22,23 +21,6 @@ def comms_nodes(amount_of_nodes) -> {int, CommsNode}:
         node.stop()
 
 
-@pytest.mark.parametrize("amount_of_nodes", [1])
-def test_locate_own_peer_id(comms_nodes):
```
marcosmartinez7, line 26: why it was removed?

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -22,23 +21,6 @@ def comms_nodes(amount_of_nodes) -> {int, CommsNode}:
...
-    comms_node = comms_nodes[1]
-
-    # attempt to locate self
-    assert comms_node.client._get_peer_id(comms_node.address) is not ""
-
-
-# FIXME: causes ERR_NO_PEERS_IN_ROUTING_TABLE in comms node although it passes
-@pytest.mark.parametrize("amount_of_nodes", [1])
-def test_locate_unregistered_peer_id(comms_nodes):
-    comms_node = comms_nodes[1]
```
marcosmartinez7, line 36: same here, i know it was failing, but why removed?

requirements-dev.txt
```diff
@@ -21,3 +21,4 @@ bump2version==0.5.8
 # Test support
 matrix-synapse==1.5.1
 json5
+pytest-repeat
```
marcosmartinez7, line 24: was this added for flakes ones? i could not found any usage

transport/rif_comms/client.py
```diff
@@ -62,13 +61,7 @@ def is_subscribed_to(self, rsk_address: Address) -> bool:
...
         :return: boolean value indicating whether the client is subscribed or not
         """
-        topic_id = self._get_peer_id(rsk_address)
-        return self.stub.HasSubscriber(
-            Subscriber(
-                peerId=topic_id,
-                channel=Channel(channelId=topic_id)
-            )
-        ).value
+        return self.stub.IsSubscribedToRskAddress(RskAddress(address=to_checksum_address(rsk_address)))
```
marcosmartinez7: is_subscribed_to returns a bool, therefore this must be 

`         return self.stub.IsSubscribedToRskAddress(RskAddress(address=to_checksum_address(rsk_address))).value`

transport/rif_comms/client.py
```diff
@@ -77,11 +70,10 @@ def send_message(self, payload: str, rsk_address: Address):
         :param payload: the message data to be sent
         :param rsk_address: the destination for the message to be sent to
         """
-        topic_id = self._get_peer_id(to_checksum_address(rsk_address))
         # TODO: message encoding
```
marcosmartinez7: i know its not related with this pr but i think we can remove this TODO

transport/rif_comms/client.py
```diff
@@ -92,8 +84,7 @@ def unsubscribe_from(self, rsk_address: Address):
         Invokes the CloseTopic GRPC API endpoint.
         :param rsk_address: RSK address which corresponds to the topic which the client is unsubscribing from.
         """
-        topic_id = self._get_peer_id(to_checksum_address(rsk_address))
-        self.stub.CloseTopic(Channel(channelId=topic_id))
+        self.stub.CloseTopicWithRskAddress(RskAddress(address=to_checksum_address(rsk_address)))
```
marcosmartinez7: i found confusing that here we use to_checksum_address and on SendMessageToRskAddress we not

i think we must use always the checksum

transport/rif_comms/client.py
```diff
@@ -102,16 +93,3 @@ def disconnect(self):
         """
         self.stub.EndCommunication(Void())
         self.grpc_channel.unsubscribe(lambda: self.grpc_channel.close())
-
-    def _get_peer_id(self, rsk_address: Address) -> str:
```
marcosmartinez7, line 98: im ok with deleting this since we are not using anymore, but, i also think we can use this call (using our address as rsk_address) as a kind of healhtcheck or ping (to know when rif comms is ready)

probably a ping or healtcheck special endpoint would be nicer, but we can consider this option if needed

marcosmartinez7 (changes requested):

raiden/tests/integration/network/transport/rif_comms/test_client.py
mortelli: because the client's `_get_peer_id` method was removed

raiden/tests/integration/network/transport/rif_comms/test_client.py
mortelli: same as previous reply

requirements-dev.txt
mortelli: i haven't specified this in the PR description yet. it's a useful tool to repeat a test through the `pytest` command in the CLI

transport/rif_comms/client.py
mortelli: correct

transport/rif_comms/client.py
mortelli: agreed

transport/rif_comms/client.py
mortelli: agreed

transport/rif_comms/client.py
mortelli: sure. let's add it back when we need it, otherwise unused functions just bloat up the codebase even more

transport/rif_comms/client.py
mortelli: fixed

transport/rif_comms/client.py
mortelli: done

transport/rif_comms/client.py
mortelli: fixed

transport/rif_comms/proto/api_pb2_grpc.py
```diff
@@ -2,7 +2,7 @@
 """Client and server classes corresponding to protobuf-defined services."""
 import grpc
 
-from transport.rif_comms.proto import api_pb2 as api__pb2
+import api_pb2 as api__pb2
```
marcosmartinez7: this should be rollbacked

marcosmartinez7:
Manually tested: 

FN - FN: ok
FN - FN - FN: ok

transport/rif_comms/proto/api_pb2_grpc.py
mortelli: > this should be rollbacked

why?

marcosmartinez7: in my case, lumino didnt started because it canot find that package.

mortelli: you're right, fixed.

thanks

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -23,92 +24,173 @@ def comms_nodes(amount_of_nodes) -> {int, CommsNode}:
...
+    assert client_1.is_subscribed_to(address_2) is False
+    assert client_2.is_subscribed_to(address_1) is False
+    assert client_2.is_subscribed_to(address_2) is False
 
     # subscribe from node 1 to 2
     client_1.subscribe_to(address_2)
 
     # check subscriptions
-    assert client_1.is_subscribed_to(address_1) is True
+    assert client_1.is_subscribed_to(address_1) is False
```
mortelli, line 113: a node no longer needs to subscribe to itself to accept other subscriptions

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -23,92 +24,173 @@ def comms_nodes(amount_of_nodes) -> {int, CommsNode}:
...
+    client_2.subscribe_to(address_1)
+
+    assert client_1.is_subscribed_to(address_1) is False
+    assert client_1.is_subscribed_to(address_2) is True
+    assert client_2.is_subscribed_to(address_1) is True
+    assert client_2.is_subscribed_to(address_2) is False
+
+
+@pytest.mark.parametrize("amount_of_nodes", [1])
+@pytest.mark.skip(reason="hangs, incomplete")
```
mortelli, line 128: added to the pending issues doc.

it's not a huge deal for us. but it's worrying because it seems to cause problems in the caller as well as the comms node

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -121,3 +203,49 @@ def test_two_clients_cross_messaging_same_topic(comms_nodes):
         received_message = notification_to_payload(resp)
         assert received_message == payload_2
         break  # only 1 message is expected
+
+
+@pytest.mark.parametrize("amount_of_nodes", [1])
+@pytest.mark.xfail(reason="incorrect exception from comms")
```
mortelli, line 209: line `221` is outputting "no peers in routing table" or something of the sort, rather than the "not subscribed" error that we'd expect

transport/rif_comms/client.py
```diff
@@ -47,8 +46,11 @@ def subscribe_to(self, rsk_address: Address) -> (str, Notification):
         :return: peer id and notification stream for receiving messages
         """
         topic_id = None
-        # TODO: catch already subscribed and any error
```
mortelli, line 50: this is no longer outputting an error from the pub-sub node

transport/rif_comms/client.py
```diff
@@ -92,28 +89,15 @@ def unsubscribe_from(self, rsk_address: Address):
...
+            RskAddress(address=to_checksum_address(rsk_address))
+        )
 
     def disconnect(self):
         """
          Invokes the EndCommunication GRPC API endpoint.
          Disconnects from RIF Communications Node. Closes grpc connection
         """
-        self.stub.EndCommunication(Void())
+        # FIXME: ending communications should not need variable assignment
```
mortelli, line 101: this _appears_ to help a lot in reducing the flakiness of the tear-down process.

i have no idea why, but we have issues to look into this (see PR description)

mortelli:
> Manually tested:
> 
> FN - FN: ok
> FN - FN - FN: ok

i've double checked this successfully :+1:

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -23,92 +24,173 @@ def comms_nodes(amount_of_nodes) -> {int, CommsNode}:
...
-def test_locate_own_peer_id(comms_nodes):
-    comms_node = comms_nodes[1]
+def test_subscribe_to_invalid(comms_nodes):
+    client = comms_nodes[1].client
+    # subscribe to unregistered address
+    topic_id, _ = client.subscribe_to(generate_address())
 
-    # attempt to locate self
-    assert comms_node.client._get_peer_id(comms_node.address) is not ""
+    # FIXME: this should probably throw an exception rather than an empty result
```
marcosmartinez7, line 32: This is really important, since we are going to threat each incorrect behaviour as an exception. 

The subscribe to flow into the payment of lumino, should check for SubscriberError

Lets report this

raiden/tests/integration/network/transport/rif_comms/test_client.py
marcosmartinez7: I think this is a big deal for us. 

If a partner of a node goes offline (and the rif comms node related to it too), the behaviour will be that the sender hangs..

raiden/tests/integration/network/transport/rif_comms/test_client.py
marcosmartinez7: this should be a UnsubscribeError

can you add this to pending issues?  this is important for exception handling

marcosmartinez7 (approved):
Lgtm

raiden/tests/integration/network/transport/rif_comms/process.py
```diff
@@ -32,7 +32,7 @@ def start(env_name: str) -> Popen:
             preexec_fn=os.setsid,  # set to later kill process group
         )
 
-        # FIXME: we need some sort of ping call, this will sometimes not be enough
```
JONAF2103, line 35: I can suggest to use some sort of port process detection while you don't have the ping functionallity, that can be done by using a command call to lsof or something programatically.

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -23,92 +24,173 @@ def comms_nodes(amount_of_nodes) -> {int, CommsNode}:
...
 
 
-# FIXME: causes ERR_NO_PEERS_IN_ROUTING_TABLE in comms node although it passes
 @pytest.mark.parametrize("amount_of_nodes", [1])
-def test_locate_unregistered_peer_id(comms_nodes):
-    comms_node = comms_nodes[1]
+def test_subscribe_to_self(comms_nodes):
+    client, address = comms_nodes[1].client, comms_nodes[1].address
+
+    topic_id, _ = client.subscribe_to(address)
```
JONAF2103, line 40: Why are we ignoring a value here, since this is a test we should check all values right?

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -23,92 +24,173 @@ def comms_nodes(amount_of_nodes) -> {int, CommsNode}:
 
 
 @pytest.mark.parametrize("amount_of_nodes", [1])
-def test_locate_own_peer_id(comms_nodes):
-    comms_node = comms_nodes[1]
+def test_subscribe_to_invalid(comms_nodes):
+    client = comms_nodes[1].client
```
JONAF2103, line 28: we are sure that comms_nodes are always greater than 0?

JONAF2103, line 28: I don't like the index access that we need to do for testing in general, anyway i don't know how to avoid that neither

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -23,92 +24,173 @@ def comms_nodes(amount_of_nodes) -> {int, CommsNode}:
...
+    assert topic_id
+
+
+@pytest.mark.parametrize("amount_of_nodes", [2])
+def test_subscribe_to_peers(comms_nodes):
+    comms_node_1, comms_node_2 = comms_nodes[1], comms_nodes[2]
+    client_1, address_1 = comms_node_1.client, comms_node_1.address
+    client_2, address_2 = comms_node_2.client, comms_node_2.address
+
+    topic_id_1, _ = client_1.subscribe_to(address_2)
```
JONAF2103, line 51: Same ignored value for the rest of the ocurrences

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -23,92 +24,173 @@ def comms_nodes(amount_of_nodes) -> {int, CommsNode}:
...
+
+
+@pytest.mark.parametrize("amount_of_nodes", [1])
+@pytest.mark.skip(reason="hangs, incomplete")
+def test_send_message_invalid(comms_nodes):
+    client, address = comms_nodes[1].client, comms_nodes[1].address
+
+    # send message to unregistered peer
+    # FIXME: hangs
+    client.send_message("echo", generate_address())
```
JONAF2103, line 134: That's it? i mean no checks can be made on the client response? or maybe a timeout?

requirements-dev.txt
JONAF2103: I had some errors using this command before, i don't know what was the problem (if test syntax or just something that i was doing wrong) anyway it's a good way to discard flake tests yes.

JONAF2103 (approved):
LGTM in general, i can't be more specific since i don't fully understand the comms ecosystem yet (basically i don't know about the proto files)

raiden/tests/integration/network/transport/rif_comms/test_client.py
mortelli: > I think this is a big deal for us.
> 
> If a partner of a node goes offline (and the rif comms node related to it too), the behaviour will be that the sender hangs..

not sure if this is the case, though. an unregistered address is not the same as an offline node.

i've added a note to [RLP-956](https://jirainfuy.atlassian.net/browse/RLP-956) so that we can have an integration test for this. we'll see what happens then

raiden/tests/integration/network/transport/rif_comms/test_client.py
mortelli: > this should be a UnsubscribeError
> 
> can you add this to pending issues? this is important for exception handling

yes, it's there. i just have to describe it better, but i'm on it

raiden/tests/integration/network/transport/rif_comms/test_client.py
mortelli: > This is really important, since we are going to threat each incorrect behaviour as an exception.
> 
> The subscribe to flow into the payment of lumino, should check for SubscriberError
> 
> Lets report this

it's in the pending issues doc, i'm in the process of describing it in full

raiden/tests/integration/network/transport/rif_comms/process.py
mortelli: > I can suggest to use some sort of port process detection while you don't have the ping functionallity, that can be done by using a command call to lsof or something programatically.

i don't think this is good enough. the comms program will boot on a specific port, but it takes a few seconds before it registers the given RSK address, and in the meantime the port query will return `ok` (as i understand it).

it has to be programmed in the pub sub node in one way or another (synchronous start up, or separate ping call)

raiden/tests/integration/network/transport/rif_comms/test_client.py
mortelli: > Why are we ignoring a value here, since this is a test we should check all values right?

the value ignored here is the stream for subscriptions, which is only used for receiving or sending messages through it.

since this test only checks that a subscription is successfully made, it only cares that there is a non-empty topic id for the subscription.

moreover, since no messages are expected to be received or sent through the aforementioned stream, there would be no point in checking it; we'd just wait for a message which would never come.

there are tests which _do_ check this variable, namely: `test_send_message_subscription`, `test_send_message_self`, and `test_send_message_peers`. in contrast to this test, these do not care about the topic id, precisely for the reason of having each test test a different thing.

raiden/tests/integration/network/transport/rif_comms/test_client.py
mortelli: > I don't like the index access that we need to do for testing in general, anyway i don't know how to avoid that neither

what do you not like about it? we are receiving comms nodes through the fixture, in the form of a collection.

how would you prefer to access this collection? by name? something like `comms_nodes["one"]`?

raiden/tests/integration/network/transport/rif_comms/test_client.py
mortelli: > Same ignored value for the rest of the ocurrences

see [previous reply](https://github.com/rsksmart/lumino/pull/182#discussion_r546843587)

raiden/tests/integration/network/transport/rif_comms/test_client.py
mortelli: > That's it? i mean no checks can be made on the client response? or maybe a timeout?

this test currently hangs, therefore it was left incomplete (and marked as such)

requirements-dev.txt
mortelli: > I had some errors using this command before, i don't know what was the problem (if test syntax or just something that i was doing wrong) anyway it's a good way to discard flake tests yes.

syntax is `pytest <your_test_module_here.py> --count=n` where `n` is an integer

raiden/tests/integration/network/transport/rif_comms/test_client.py
marcosmartinez7: an offline node will not be part of the dht after 6 hrs

a unregistered address will not be part too

raiden/tests/integration/network/transport/rif_comms/test_client.py
JONAF2103: i mean i know this is how it works but it looks weird like saying something like:
```python
@pytest.mark.parametrize("amount_of_nodes", [1])
def test_locate_own_peer_id(comms_nodes):
    comms_nodes[0]
...
```
Why? because if you are saying that comms_nodes is length 1 it's better to have something like:
```python
def test_locate_own_peer_id(comms_node):
comms_node.something
```
Anyway i know this is part of the language or the framework that we are using, but i just don't like to access lists by index in any language

raiden/tests/integration/network/transport/rif_comms/test_client.py
JONAF2103: Sounds like subscribe_to does more than one thing, get the stream and the topic id.

mortelli: > Sounds like subscribe_to does more than one thing, get the stream and the topic id.

correct. this is the way the function is programmed in the pubsub node

raiden/tests/integration/network/transport/rif_comms/test_client.py
mortelli: > i mean i know this is how it works but it looks weird like saying something like:
> 
> ```python
> @pytest.mark.parametrize("amount_of_nodes", [1])
> def test_locate_own_peer_id(comms_nodes):
>     comms_nodes[0]
> ...
> ```
> 
> Why? because if you are saying that comms_nodes is length 1 it's better to have something like:
> 
> ```python
> def test_locate_own_peer_id(comms_node):
> comms_node.something
> ```
> 
> Anyway i know this is part of the language or the framework that we are using, but i just don't like to access lists by index in any language

i disagree only due to the fact that i don't like the readability of this:

```python
client_1 = comms_node[0]
```

and prefer the way it currently is:
```python
client_1 = comms_node[1]
```

but i appreciate the sentiment regarding list access by index. as you say, it is unavoidable for now.

PR #183, opened 2020-12-15
Feature/rlp 927/shutdown lumino if comms stops

marcosmartinez7:
## Description 

This PR aims to tackle two issues: 
1. lumino must stop if transport layer is down
2. when rif comms transport node stops, it was calling the endComms stub method, that implies that the topics will be destroyed, and i dont found any use case for that

## Changelog 

- greenlet.kill doesnt make any sense, and also was causing incorrect shutdown behaviour: removed
- rif comms client now has two endpoints, a) disconnect: closes the grpc channel, b) end_communications: [todo pending behaviour]

## Manually tested

- rif comms start
- lumino start
- stop rif comms api server after lumino subscribes to rif comms node
- lumino should output: 
`2020-12-15 22:06:09.188221 [warning  ] RIF Comms transport node stopped, shutting down Lumino [transport.rif_comms.node] greenlet_name=RaidenService._run node:60a12f09`

## Pending

Im not happy with the exception handling and transport.stop invocations between raiden_service and the trasnsport layer. 

there is  exception linking between raiden_service and transport and exception handling into the rif comms transport node class

### DIFF
--- a/transport/rif_comms/client.py
+++ b/transport/rif_comms/client.py
@@ -97,9 +97,7 @@ class Client:
 
     def terminate(self):
         """
-        Disconnects from RIF Communications Node. Closes grpc connection
         """
-        self.grpc_channel.unsubscribe(lambda: self.grpc_channel.close())
 
     def _get_peer_id(self, rsk_address: Address) -> str:
         """
@@ -117,6 +115,9 @@ class Client:
     def disconnect(self):
         """
          Invokes the EndCommunication GRPC API endpoint.
+         Disconnects from RIF Communications Node. Closes grpc connection
         """
         self.stub.EndCommunication(Void())
+        self.grpc_channel.unsubscribe(lambda: self.grpc_channel.close())
+
 
--- a/transport/rif_comms/client.py
+++ b/transport/rif_comms/client.py
@@ -95,10 +95,6 @@ class Client:
         topic_id = self._get_peer_id(to_checksum_address(rsk_address))
         self.stub.CloseTopic(Channel(channelId=topic_id))
 
-    def terminate(self):
-        """
-        """
-
     def _get_peer_id(self, rsk_address: Address) -> str:
         """
         Gets the peer ID associated with a node RSK address.
--- a/transport/rif_comms/client.py
+++ b/transport/rif_comms/client.py
@@ -98,9 +98,7 @@ class Client:
     def disconnect(self):
         """
         Disconnects from RIF Communications Node.
-        Invokes the EndCommunication GRPC API endpoint.
         """
-        self.stub.EndCommunication(Void())
         self.grpc_channel.unsubscribe(lambda: self.grpc_channel.close())
 
     def _get_peer_id(self, rsk_address: Address) -> str:
@@ -115,3 +113,9 @@ class Client:
                 details = "Failed to lookup key! No peers from routing table!"
         """
         return self.stub.LocatePeerId(RskAddress(address=to_checksum_address(rsk_address))).address
+
+    def end_communications(self):
+        """
+        Invokes the EndCommunication GRPC API endpoint.
+        """
+        self.stub.EndCommunication(Void())
--- a/transport/rif_comms/client.py
+++ b/transport/rif_comms/client.py
@@ -95,9 +95,9 @@ class Client:
         topic_id = self._get_peer_id(to_checksum_address(rsk_address))
         self.stub.CloseTopic(Channel(channelId=topic_id))
 
-    def disconnect(self):
+    def terminate(self):
         """
-        Disconnects from RIF Communications Node.
+        Disconnects from RIF Communications Node. Closes grpc connection
         """
         self.grpc_channel.unsubscribe(lambda: self.grpc_channel.close())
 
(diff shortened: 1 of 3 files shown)


transport/rif_comms/client.py
```diff
@@ -115,3 +113,9 @@ def _get_peer_id(self, rsk_address: Address) -> str:
                 details = "Failed to lookup key! No peers from routing table!"
         """
         return self.stub.LocatePeerId(RskAddress(address=to_checksum_address(rsk_address))).address
+
+    def end_communications(self):
+        """
+        Invokes the EndCommunication GRPC API endpoint.
+        """
+        self.stub.EndCommunication(Void())
```
mortelli: i wouldn't add this method if we're not going to use it, so i think we should either call it somewhere or remove it

mortelli: also, i know we've talked about this before, but it should be clearer what `end_communications` does different than `disconnect`.

for me, we should have connections and disconnections on 2 levels:
- `grpc.insecure_channel` opens up a GRPC channel, then we call `stub.ConnectToCommunicationsNode` over this channel.
- `client.disconnect` should be the counterpart to the `client.connect` call, meaning the GRPC channel should still be open before and after both of these.
- `grpc_channel.unsubscribe` should be the counterpart to `grpc.insecure_channel`, closing the GRPC channel.

so if the `client.__init__` method calls `grpc.insecure_channel`, then some sort of client `stop` method should call the `grpc_channel.unsubscribe`. this could be called when the transport node is stopped.

with this in mind, `stub.EndCommunication` should be called under `client.disconnect`; assuming it is the counterpart to `stub.ConnectToCommunicationsNode` under `client.connect`.

while the closing of the GRPC channel seems clear, i'm not exactly sure what effect `EndCommunication` is having on the comms node though. is it any different from calling `grpc_channel.unsubscribe` from the perspective of the comms node?

finally, in my opinion all of this should be orthogonal to **topic** subscription/unsubscription.

transport/rif_comms/node.py
```diff
@@ -176,23 +177,21 @@ def stop(self):
             if message_queue.greenlet:
                 message_queue.notify()  # if we need to send something, this is the time
 
-        if self.greenlet:
-            self.greenlet.kill()
-            self.greenlet.get()
```
mortelli, line 181: i am not sure it is correct to delete this.

what if `_run` exits through `except Exception`?

in that case `self.greenlet` will not have been stopped, therefore these original statements kind of make sense to me.

what do you think? could you please explain a bit more about this "incorrect shutdown behavior"?

transport/rif_comms/node.py
```diff
@@ -176,23 +177,21 @@ def stop(self):
...
         try:
             del self._log
         except AttributeError:
             # During shutdown the log attribute may have already been collected
             pass
-        # parent may want to call get() after stop(), to ensure _run errors are re-raised
-        # we don't call it here to avoid deadlock when self crashes and calls stop() on finally
+
+        # end grpc communication
+        self._comms_client.disconnect()
```
mortelli, line 183: it's possible we'll have to change this if we go by [my previous suggestion regarding the client's code](https://github.com/rsksmart/lumino/pull/183#discussion_r544266764)

mortelli (commented):
i've left some general comments, but i have an important question besides these:

what if a LC stops its execution? its transport should be stopped, but lumino as a whole should not be. there could be other LCs running.

other than this, i don't think this statement is true atm:
> endComms stub method implies that the topics will be destroyed

mortelli:
also, could you please expand on the pending issue in the PR description? i'm not sure i'm following

transport/rif_comms/client.py
marcosmartinez7: Agree, next commit:

init: creates grpc channl
connect: invokes ConnectToCommunicationsNode
disconnect: invokes EndCommunications
destroy: removes grpc channel subscription (better name for this? the opposites for init are like end, disconnect, deactivate, etc, and is kinda confussing with disconnect method)

mortelli: > Agree, next commit:
> 
> init: creates grpc channl
> connect: invokes ConnectToCommunicationsNode
> disconnect: invokes EndCommunications
> destroy: removes grpc channel subscription (better name for this? the opposites for init are like end, disconnect, deactivate, etc, and is kinda confussing with disconnect method)

`stop`, `end`, `cleanup`, `terminate`

transport/rif_comms/node.py
marcosmartinez7: of run exits through except Exception it will call the stop method and lumino will be shutted down

i dont found any issues with self.greenlet being unstopped. Why? because that greenlet is just for the _run execution. The run execution was stopped by the exception, then the greenlet isnt running anyway, and also, the lumino process will be shutted down.
 
About the incorrect shutdown behavior, code hangs on greenlet.kill, dont know why, but i dont found coherent to kill a greenlet when a exception on the run method occurs (because the run method is the greenlet execution itself). Or at least, i dont fund ok to kill it from inside the method that is executing.

that code was "copied" from the matrix client implemention.

the matrix transport node stop method calls self._client.stop_listener_thread() and stop the client thread. I think thats because the greenlet of the matrix node transport is used for different things. It was used for the global msg worker, and the client was responsible of the sync with the matrix homeserver

since we handle the streams for comms message reception on the transport node (not in the rif comms client), the behaviour is different

transport/rif_comms/node.py
marcosmartinez7: yes indeed, on it

transport/rif_comms/client.py
marcosmartinez7: done, terminate.

marcosmartinez7:
> also, could you please expand on the pending issue in the PR description? i'm not sure i'm following

basically im not happy whit the exception handling between raiden_service and the transport layer. Thats an inherited architectural issue.

mortelli (approved):

PR #184, opened 2020-12-16
RLP-933 Insufficient funds controls

JONAF2103:
### Introduction

In this PR we attempt to solve issues with LC transactions executed using accounts without funds. The idea is to avoid having exceptions that can pull the hub down. With this change the hub will respond with an error instead.

#### Changes

* Added a decorator to check the account balance for the LC
* Added a decorator to control the LC operations to respond with errors instead of pulling down the hub node.
* Refactoring `check_for_insufficient_eth` function and all it's occurrences to remove the `block_identifier` parameter since it's not neccesary and also it causes problems on the `web3.eth.getBalance` call, instead we are using a better value provided by the client itself in it's inner function called `get_checking_block`. With this refactor we avoid format problems and we make sure that the getBalance call will never fail.

#### Scope

This affects the LC operations and the error handling for all the on chain operations.


raiden/api/validations/light_client_balance.py
```diff
@@ -0,0 +1,33 @@
...
+log = structlog.get_logger(__name__)
+
+
+def requires_lc_balance(func):
+    """
+        Check if the endpoint handled by func was called with a Lc that has balance greater than 0.
+    """
+    def inner(*args, **kwargs):
+        raiden_api = args[0].raiden_api
+        key_error = get_key_error(raiden_api.raiden.wal)
```
marcosmartinez7: im not sure about checking the api key here.

i thought that those rest endpoints were checking the api key already, but seems that only the payment messages ones were controled. 

anyway, i found more clear to split this into two decorators

- requires_api_key
- requires_lc_balance

then on the open_light you can just put the two decorators on that order

what do u think?

raiden/api/validations/light_client_balance.py
```diff
@@ -0,0 +1,33 @@
...
+        if key_error:
+            return key_error
+        api_key = request.headers.get("x-api-key")
+        light_client = LightClientService.get_by_api_key(api_key=api_key, wal=raiden_api.raiden.wal)
+        web3 = raiden_api.raiden.default_registry.client.web3
```
marcosmartinez7: lets add an issue and a todo for moving this

raiden/api/validations/light_client_balance.py
JONAF2103: I tried that but you need to depend on requires_api_key and maybe the user doesn't use that in the call and you can have a problem there.

raiden/api/validations/light_client_balance.py
JONAF2103: Sure i don't like it

raiden/api/validations/light_client_authorization.py
```diff
@@ -22,7 +22,7 @@ def inner(*args, **kwargs):
             return ApiErrorBuilder.build_and_log_error(
                 errors="Missing api_key auth header", status_code=HTTPStatus.BAD_REQUEST, log=log
             )
-        light_client = LightClientService.get_by_api_key(api_key, args[0].raiden_api.raiden.wal)
+        light_client = LightClientService.get_by_api_key(api_key=api_key, wal=args[0].raiden_api.raiden.wal)
```
mortelli: it seems like this entire function depends on having a `RaidenAPI` object in `args[0]`. with this in mind, what are your thoughts on this?

```python
def requires_api_key(func):
    """
        Check if the endpoint handled by func was called with a valid 'x-api-key' header.
        This decorator requires the class where func is defined to have a property raiden_api,
        of type raiden.api.python.RaidenAPI.
    """

    def inner(rest_api: RestAPI, *args, **kwargs):
        api_key = request.headers.get("x-api-key")
        if not api_key:
            return ApiErrorBuilder.build_and_log_error(
                errors="Missing api_key auth header", status_code=HTTPStatus.BAD_REQUEST, log=log
            )
        light_client = LightClientService.get_by_api_key(api_key=api_key, wal=rest_api.raiden_api.raiden.wal)
        if not light_client:
            return ApiErrorBuilder.build_and_log_error(
                errors="There is no light client associated with the api key provided",
                status_code=HTTPStatus.FORBIDDEN,
                log=log
            )
        return func(rest_api, *args, **kwargs)

    return inner
```

raiden/api/validations/light_client_balance.py
```diff
@@ -0,0 +1,31 @@
...
+from flask import request
+
+from raiden.api.validations.api_error_builder import ApiErrorBuilder
+from raiden.api.validations.light_client_authorization import requires_api_key
+from raiden.lightclient.handlers.light_client_service import LightClientService
+
+log = structlog.get_logger(__name__)
+
+
+def requires_lc_balance(func):
```
mortelli: as with the other one, maybe we can do this:

```python
def requires_lc_balance(func):
    """
        Check if the endpoint handled by func was called with a Lc that has balance greater than 0.
    """
    @requires_api_key
    def inner(rest_api: RestAPI, *args, **kwargs):
        raiden_api = rest_api.raiden_api
        api_key = request.headers.get("x-api-key")
        light_client = LightClientService.get_by_api_key(api_key=api_key, wal=raiden_api.raiden.wal)
        web3 = raiden_api.raiden.default_registry.client.web3
        current_lc_balance = web3.eth.getBalance(light_client.address)
        if current_lc_balance > 0:
            return func(rest_api, *args, **kwargs)
        return ApiErrorBuilder.build_and_log_error(
            errors="Insufficient Funds",
            status_code=HTTPStatus.PAYMENT_REQUIRED,
            log=log
        )
    return inner
```

raiden/api/validations/light_client_safe_operation.py
```diff
@@ -0,0 +1,34 @@
...
+from raiden.api.validations.api_error_builder import ApiErrorBuilder
+from raiden.api.validations.light_client_authorization import requires_api_key
+from raiden.exceptions import InsufficientFunds
+
+log = structlog.get_logger(__name__)
+
+
+def lc_safe_operation(func):
+    """
+        Executes an lc operation and never pull down the node, instead it always respond with an error code.
```
mortelli:
```suggestion
        Executes an operation and never shuts down the node, instead it always responds with an error code.
```

raiden/api/rest.py
```diff
@@ -1845,6 +1853,8 @@ def _deposit(
         result = self.channel_schema.dump(updated_channel_state)
         return api_response(result=result.data)
 
+    @requires_lc_balance
```
marcosmartinez7: correct me if im wrong, this function will: 

1. fail if there is no api key
2. fail if balance is 0
3. fail if balance is greater than 0 but insufficient to make the onchain operation

if 3) happens, it will be catched by is_safe_operation. The possible errors are InsufficientFunds (raised by check_why_[operation]_fail and an unexpected one.

is that corect?

mortelli (commented):
thanks for this PR :+1:

i have not tested it, but i have a few comments besides the ones further down below.

### `@requires_lc_balance`

i see that this decorator is missing from these functions:
- `register_secret_light` 
- `initiate_payment_light`
- `post_unlock_payment_light`

it seems these already catch `InsufficientFunds` so i wanted to ask for your thoughts on this.

### `@lc_safe_operation`

this doesn't seem to be light client specific, no? based on this:
- why don't we call it something like `api_safe_operation`?
- is there a reason why other API methods do not have this decorator applied to them?

### modularization 

do we need a reason to put each decorator in a separate python module inside `api/validations`?

we can _at least_ put all the lc-related decorators in a single module, and leave the rest as is.

### `check_for_insufficient_eth`

i see that many of these calls were added; are you aware that all of these can raise exceptions up to the caller?

### LGTM
it appears you have 2 new alerts.

raiden/api/validations/light_client_safe_operation.py
```diff
@@ -0,0 +1,34 @@
...
+import traceback
+
+from raiden.api.validations.api_error_builder import ApiErrorBuilder
+from raiden.api.validations.light_client_authorization import requires_api_key
+from raiden.exceptions import InsufficientFunds
+
+log = structlog.get_logger(__name__)
+
+
+def lc_safe_operation(func):
```
marcosmartinez7: is_safe_operation sounds like "this function will ceck for something and executes if that condition met" 

but in fact, this is a wrapper decorator that executes a function always and catch the exception

i think that `execute_lc_onchain_operation` is better

not a blocker anyway

raiden/network/proxies/payment_channel.py
```diff
@@ -223,6 +223,7 @@ def close_light(
         """ Closes the channel using the provided balance proof. """
         self.token_network.close_light(
             channel_identifier=self.channel_identifier,
+            closing_participant=self.participant1,
```
marcosmartinez7, line 226: this isnt related with the PR issue

why was this added?

raiden/network/proxies/token_network.py
```diff
@@ -205,6 +205,13 @@ def _new_channel_preconditions(
         if channel_exists:
             raise DuplicatedChannelError("Channel with given partner address already exists")
 
+        self.proxy.jsonrpc_client.check_for_insufficient_eth(
```
marcosmartinez7, line 208: this is great

raiden/network/proxies/token_network.py
```diff
@@ -234,6 +241,12 @@ def new_netting_channel_light(self, creator: Address, partner: Address, signed_t
                     log.critical("new_netting_channel_light failed", **log_details)
                     raise RaidenRecoverableError("creating new channel failed")
             except HTTPError as e:
+                self.proxy.jsonrpc_client.check_for_insufficient_eth(
```
marcosmartinez7, line 245: first i thought that this must be an invocation to check_why_open_failed, but that function did not exists

we can assume that an HTTPError is probably a balance stuff and check for insufficient eth

this will throw an InsuficientFunds, and if not, it will throw RawTransactionFailed and catched by the is_safe_operation decorator

therefore im ok with this, just toughts

raiden/network/proxies/token_network.py
```diff
@@ -234,6 +241,12 @@ def new_netting_channel_light(self, creator: Address, partner: Address, signed_t
...
+                    transaction_name="openChannel",
+                    address=creator,
+                    transaction_executed=False,
+                    required_gas=GAS_REQUIRED_FOR_OPEN_CHANNEL
+                )
                 log.warning("new_netting_channel failed: transaction malformed", ex=e, **log_details)
                 new_open_channel_transaction.set_exception(e)
```
marcosmartinez7: it line relevant? 

note that if the check_for_insufficient_eth raises insufficient funds, this will not be set

im not familiar with this, the set exception basically does a notifier/subscribe machanism,  not sure if we are subscribing on any part of the system to that events. I dont think so:

```    
def set_exception(self, exception, exc_info=None):
        """Store the exception and wake up any waiters.

        All greenlets blocking on :meth:`get` or :meth:`wait` are awakened.
        Subsequent calls to :meth:`wait` and :meth:`get` will not block at all.

        :keyword tuple exc_info: If given, a standard three-tuple of type, value, :class:`traceback`
            as returned by :func:`sys.exc_info`. This will be used when the exception
            is re-raised to propagate the correct traceback.
        """

marcosmartinez7 (commented):
lgtm in general, just some toughts to discuss that im not familiar enough

raiden/api/rest.py
JONAF2103: yes

raiden/api/validations/light_client_authorization.py
JONAF2103: Actually this function shouldn't be changed here, i will reset this since i didn't added any change at all here

raiden/api/validations/light_client_balance.py
JONAF2103: I will try

raiden/api/validations/light_client_safe_operation.py
JONAF2103: Actually we should call it something like `lc_request` or `handled_lc_request` because actually we don't want to pull down the hub on any lc operation, and you can use this decorator for other operations in the future that are not related to on chain operations.

raiden/api/validations/light_client_safe_operation.py
JONAF2103: Shuts seems like a controlled scenario and this is for unhandled scenarios, maybe `throw the node down` or something similar sounds better.

raiden/network/proxies/payment_channel.py
JONAF2103: Don't mix, be careful i put closing_participant but is not related to any other issue is just to pass the initiator address to the child function to check for insufficient eth that's all

raiden/network/proxies/token_network.py
JONAF2103: Probably you are right i will move the call after this because we don't know

raiden/api/validations/light_client_authorization.py
JONAF2103: Or better i will try to add your suggestion

raiden/api/validations/light_client_safe_operation.py
mortelli: i'd say `crashes` if you want uncontrolled shutdown semantics. `throw down` means a lot of other different things :stuck_out_tongue:

raiden/api/validations/light_client_authorization.py
mortelli: not a blocker then, up to you :+1:

JONAF2103: Applied

raiden/api/validations/light_client_safe_operation.py
JONAF2103: Sounds better will fix it

JONAF2103:
> thanks for this PR +1
> 
> i have not tested it, but i have a few comments besides the ones further down below.
> ### `@requires_lc_balance`
> 
> i see that this decorator is missing from these functions:
> 
>     * `register_secret_light`
> 
>     * `initiate_payment_light`
> 
>     * `post_unlock_payment_light`
> 
> 
> it seems these already catch `InsufficientFunds` so i wanted to ask for your thoughts on this.
> ### `@lc_safe_operation`
> 
> this doesn't seem to be light client specific, no? based on this:
> 
>     * why don't we call it something like `api_safe_operation`?
> 
>     * is there a reason why other API methods do not have this decorator applied to them?
> 
> 
> ### modularization
> 
> do we need a reason to put each decorator in a separate python module inside `api/validations`?
> 
> we can _at least_ put all the lc-related decorators in a single module, and leave the rest as is.
> ### `check_for_insufficient_eth`
> 
> i see that many of these calls were added; are you aware that all of these can raise exceptions up to the caller?
> ### LGTM
> 
> it appears you have 2 new alerts.

### `@lc_safe_operation`
initiate_payment_light is a protocol logic related, basically is an offchain operation, in the first approach we never talk about adding the control for all the light operations, probably it's a good idea but i didn't add that here because could cause some problem. Basically i only added that to the on chain related operations, this post_unlock_payment_light and this register_secret_light,  are missing ones, i will add them, but maybe we can discuss about if we should add that control for all the light operations (on chain and offchain).

### modularization
I didn't give a tought about it actually, i just keep the old structure but maybe you are right, i don't know if this should be done here or in another PR maybe

### `check_for_insufficient_eth`
Yes, it's the idea, actually we raise that because it's an specific case and we want to handle it properly.

### LGTM
I will check that out

marcosmartinez7 (approved):
Superb 👍🏻

JONAF2103:
All suggested changes are applied and all tests are ok, i will merge this now

PR #186, opened 2020-12-21
RLP-960 Fixing matrix server error

JONAF2103:
In this PR we are adding our own matrix homeserver to avoid having availability issues, also removing the yml source for development since we are not going to change this anymore.

### DIFF
--- a/raiden/settings.py
+++ b/raiden/settings.py
@@ -21,9 +21,7 @@ DEFAULT_MATRIX_KNOWN_SERVERS = {
         "https://raw.githubusercontent.com/raiden-network/raiden-transport"
         "/master/known_servers.main.yaml"
     ),
-    Environment.DEVELOPMENT: (
-        "https://raw.githubusercontent.com/marcosmartinez7/matrix-known-servers/master/known_servers.test.yaml"
-    ),
+    Environment.DEVELOPMENT: ["http://64.225.11.151"],
 }
 
 DEFAULT_REVEAL_TIMEOUT = 50
--- a/raiden/ui/app.py
+++ b/raiden/ui/app.py
@@ -66,7 +66,8 @@ def _setup_matrix(config):
     if config["transport"]["matrix"].get("available_servers") is None:
         # fetch list of known servers from raiden-network/raiden-tranport repo
         available_servers_url = DEFAULT_MATRIX_KNOWN_SERVERS[config["environment_type"]]
-        available_servers = get_matrix_servers(available_servers_url)
+        available_servers = available_servers_url if \
+            config["environment_type"] == Environment.DEVELOPMENT else get_matrix_servers(available_servers_url)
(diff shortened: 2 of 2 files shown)


raiden/ui/app.py
```diff
@@ -66,7 +66,8 @@ def _setup_matrix(config):
     if config["transport"]["matrix"].get("available_servers") is None:
         # fetch list of known servers from raiden-network/raiden-tranport repo
         available_servers_url = DEFAULT_MATRIX_KNOWN_SERVERS[config["environment_type"]]
-        available_servers = get_matrix_servers(available_servers_url)
+        available_servers = available_servers_url if \
```
marcosmartinez7, line 69: why we need to check the environment type here? 
if we use:

`         available_servers_url = DEFAULT_MATRIX_KNOWN_SERVERS[config["environment_type"]] `

the `DEFAULT_MATRIX_KNOWN_SERVERS` variable will point to the development or production ones already, right?

JONAF2103: It's because of the typing, if config["environment_type"] it's PRODUCTION then the format for available_servers_url it's the list of yml files to get the urls from, but for DEVELOPMENT we are using the server url directly so we don't need to parse it.

JONAF2103: We could improve this by removing the get_matrix_servers function call and modifying the format on PRODUCTION config value too and always using the final url values instead of yml files.

marcosmartinez7: Ok got it, its ok for me to leave it this way and iterate when the Production environment is used correctly.

marcosmartinez7 (approved):

raiden/ui/app.py
JONAF2103: Sounds good to me

raiden/settings.py
```diff
@@ -21,9 +21,7 @@
         "https://raw.githubusercontent.com/raiden-network/raiden-transport"
         "/master/known_servers.main.yaml"
     ),
-    Environment.DEVELOPMENT: (
-        "https://raw.githubusercontent.com/marcosmartinez7/matrix-known-servers/master/known_servers.test.yaml"
-    ),
+    Environment.DEVELOPMENT: ["http://64.225.11.151"],
```
mortelli, line 24: why was this value type changed to `list`?

mortelli (approved):

raiden/settings.py
JONAF2103: Before this change we had a list of yml files to use to get the server urls, now we decided to put the server url directly here instead, the type change is because we don't need to get the yml file and parse it to get the server urls, now we just put the server url's in the list directly and we use them from there.

marcosmartinez7: because of the method that parses asumes a yaml, i have created an issue for this comment and the previous one https://jirainfuy.atlassian.net/browse/RLP-962

PR #189, opened 2020-12-21
RLP-928/sender validation

marcosmartinez7:
## Description

On the latest release of rif comms pubsub bootnode, the team introduced a `sender` field that identifies the sender of a message

On this PR, we validate that the `recovered address` from a signature of a received raiden message is equals to the `sender`

This way, we validate that the messages are sent by the sender unequivocally (*)

(*) A handshake between lumino and rif comms is needed in order to take that sender attribute as a secure data field.

## Changelog 

- created `get_sender_from_notification` method that returns the sender from a notification
-  added the peer_address param on the reception of rif comms messages 
- re-added the sender validation on `validate_and_parse_messages`

## Acceptance criteria

- [x] unit tests should pass
- [x] payments using matrix should work 
- [x] payments using rif comms shuld work

### DIFF
--- a/transport/rif_comms/utils.py
+++ b/transport/rif_comms/utils.py
@@ -1,5 +1,8 @@
 import json
 
+from eth_utils import to_canonical_address
+
+from raiden.utils import Address
 from transport.rif_comms.proto.api_pb2 import ChannelNewData
 
 
@@ -11,7 +14,12 @@ def notification_to_payload(notification_data: ChannelNewData) -> str:
     content_data = notification_data.channelNewData.data
     """
     ChannelNewData has the following structure:
-        from: "16Uiu2HAm8wq7GpkmTDqBxb4eKGfa2Yos79DabTgSXXF4PcHaDhWJ"
+        peer:  {
+          address: "16Uiu2HAmV4KttHooePhKsDsHQTJUzHH6FLv1hRA86hT5Js994hFz"
+        }
+        sender {
+          address: "0xB2D3c4055C4d6B3832a1ee0eF8a5EB25F4D56292"
+        }
         data: "{\"type\":\"Buffer\",\"data\":[104,101,121]}"
         nonce: "\216f\225\232d\023e{"
         channel {
@@ -22,3 +30,8 @@ def notification_to_payload(notification_data: ChannelNewData) -> str:
         content = json.loads(content_data.decode())  # deserialize `data` dict
         return bytes(content["data"]).decode()  # deserialize `data.data` field
     return ""
+
+
+def get_sender_from_notification(notification_data: ChannelNewData) -> Address:
+    return to_canonical_address(notification_data.sender.address)
+
--- a/transport/rif_comms/utils.py
+++ b/transport/rif_comms/utils.py
@@ -33,5 +33,5 @@ def notification_to_payload(notification_data: ChannelNewData) -> str:
 
 
 def get_sender_from_notification(notification_data: ChannelNewData) -> Address:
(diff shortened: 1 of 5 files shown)


transport/rif_comms/utils.py
```diff
@@ -22,3 +30,8 @@ def notification_to_payload(notification_data: ChannelNewData) -> str:
         content = json.loads(content_data.decode())  # deserialize `data` dict
         return bytes(content["data"]).decode()  # deserialize `data.data` field
     return ""
+
+
+def get_sender_from_notification(notification_data: ChannelNewData) -> Address:
+    return to_canonical_address(notification_data.channelNewData.sender.address)
```
mortelli: docstring here would be nice :sparkles:

transport/rif_comms/utils.py
```diff
@@ -11,7 +14,12 @@ def notification_to_payload(notification_data: ChannelNewData) -> str:
     content_data = notification_data.channelNewData.data
     """
     ChannelNewData has the following structure:
```
mortelli, line 16: just to have this docstring a little bit more consistent, i'd suggest:
```
ChannelNewData has the following structure:
    peer: {
      address: "16Uiu2HAmV4KttHooePhKsDsHQTJUzHH6FLv1hRA86hT5Js994hFz"
    }
    sender: {
      address: "0xB2D3c4055C4d6B3832a1ee0eF8a5EB25F4D56292"
    }
    data: "{\"type\":\"Buffer\",\"data\":[104,101,121]}"
    nonce: "\216f\225\232d\023e"
    channel: {
      channelId: "16Uiu2HAm9otWzXBcFm7WC2Qufp2h1mpRxK1oox289omHTcKgrpRA"
    }
```

transport/rif_comms/utils.py
mortelli:
```suggestion
def get_sender_from_notification(notification_data: ChannelNewData) -> Address:
    try:
        return to_canonical_address(notification_data.channelNewData.sender.address)
    except AttributeError:
        return Address(b'')
```
also, the type hint of the return should come from `eth_typing`, not `raiden.utils`

mortelli (approved):
so this looks good, but i'd add a `try`-`except` as detailed below to prevent unnecessary crashes in case the pub-sub node messes up the sender.

besides this, i'd add an integration test for the client like so:
```python
@pytest.mark.parametrize("amount_of_nodes", [2])
def test_send_message_sender(comms_nodes):
    comms_node_1, comms_node_2 = comms_nodes[1], comms_nodes[2]
    client_1, address_1 = comms_node_1.client, comms_node_1.address
    client_2, address_2 = comms_node_2.client, comms_node_2.address

    # subscribe both nodes to each other
    _, sub_1_to_2 = client_1.subscribe_to(address_2)
    _, sub_2_to_1 = client_2.subscribe_to(address_1)

    # send messages and listen
    client_1.send_message("ping", address_2)
    client_2.send_message("pong", address_1)

    for resp in sub_1_to_2:
        sender = get_sender_from_notification(resp)
        assert sender == address_1
        break  # only 1 message is expected

    for resp in sub_2_to_1:
        sender = get_sender_from_notification(resp)
        assert sender == address_2
        break  # only 1 message is expected
```

transport/rif_comms/utils.py
marcosmartinez7: i dont understand, which is the inconsistence?

mortelli: just a couple of `:` and whitespaces, it's very minor

PR #190, opened 2020-12-22
RLP-959/use private rooms

marcosmartinez7:
## Description 

On this PR the `config["transport"]["matrix"]["private_rooms"] ` is set to True on any scenario.

Previously, it was only for Production, but since we always use development environment, we must change this to protect the messages on mainnet.

We use development env even in mainnet, because Production env is too coupled with raiden contracts, matrix servers, etc.

Thats something we need to improve.

## Acceptance criteria

- [x] unit test should pass
- [x] matrix payments using private rooms works


JONAF2103 (approved):
LGTM, we should make this configurable from the script parameters but it's ok for now

PR #191, opened 2020-12-23
RLP-956: integrate new rif-comms version

mortelli:
This PR solves issue [RLP-956](https://jirainfuy.atlassian.net/browse/RLP-956), which is about integrating the new version of the RIF Comms Pub-sub node into Lumino.

## Overview
Besides adding the new proto files, the RIF Comms client code had its signatures updated.

With the aim of validating these new functionalities, the integration test suite for RIF Comms transport was extended.

### RIF Comms client
- the `connect` method does not have a `return` statement anymore; it was unneeded and unused.
- the `_get_peer_id` method was added back to the client. it is currently only used to verify correctness in the integration tests.
- `subscribe_to` and `unsubscribe_from` have had their signatures updated to match the latest pubsub implementation.
- `is_subscribed_to` is now an internal function. it is currently only used to verify correctness in the integration tests; we no longer check for subscription in order to send a raiden message.
- `disconnect` no longer calls `EndCommunication`, as the latest version for this call does nothing. see pending issue [here](https://jirainfuy.atlassian.net/browse/RLP-925)

### RIF Comms integration tests
- 3 separate tests for the client `connect` function were added.
  - since we don't want automatic connection for the comms nodes in these tests, the existing fixture was not used here.
  - an `auto_connect` field was added to the `Config` and `Node` test classes in order to re-use them without automatic connection. its default value is `True` so that this is all compatible with the existing `comms_nodes` fixture.
  - the aforementioned tests are:
      1. boot up a comms node and create a client with an RSK address. first, check that the address cannot be found in the comms node. then, connect and check that it can. this test currently **fails** because of an incorrect exception.
      2. boot up a comms node and create a client with an RSK address. connect to the node and get the correspondent peer id. then, repeat the connection call, and check that the peer id remains the same.
      3. boot up 2 comms nodes and connect separate clients with different RSK addresses for each of them. then, try to locate the peer ID from each client based on the other client's address.
- since the client method `_is_subscribed_to` is (now) internal, its separate tests were removed. however, the logic in these tests was merged into the related `subscribe_to` tests, which now use this internal function as a means of verification.
    - `test_is_subscribed_to_invalid` logic is now in `test_subscribe_to_invalid`
    - `test_is_subscribed_to_self` logic is now in `test_subscribe_to_self`
    - `test_is_subscribed_to_peers` logic is now in `test_subscribe_to_peers`
[...]

PR #192, opened 2020-12-24
RLP-969/ Multi addr test cases

marcosmartinez7:
## Description

This PR aims to solve two issues

1. Create test cases for comms nodes with more than 1 client
2. Extend the testing framework to use different scenarios (1-N clients per node)

## Changelog

- modified Config and CommsNode to allow more than 1 client per node.
- modified `comms_nodes` fixture, now is called `comms_clients`. The fixture returns all the clients without taking into account the topology of the scenario, therefore the test cases can be invoked for each client transparently. 
- new test cases: `test_multiple_subscriptions_same_node` and `test_subscribe_to_peers_multi_address` are now xfailing since rif comms isue

## Multi address tests

The following are the test cases that uses multi addr topology (2 clients into the same comms node) 

1. test_subscribe_to_invalid: two clients subscribing to invalid address into the same node
2. test_subscribe_to_self: two clients subscribing to self into the same node
3. test_subscribe_to_peers: two clients suscribing to each other into the same node
4. test_subscribe_to_repeated: same as 3 but repeated
5. test_is_subscribed_to_self_mutli_address: same as 2 but to test side effects of suscriptions (can be merged i think) 
6. test_send_message_subscription:send msgs from two clients and the same node
7. test_send_message_peers: same as 6
8. test_unsubscribe_from_self: same as 2
9. test_unsubscribe_from_peers: same as 3
10. test_send_message_sender: validate sender with multiple clients

## Acceptance criteria

- [x] Test cases should pass

## Current state of tests

`25 passed, 1 skipped, 1 warnings in 286.58s (0:04:46) `

## TODOs

- Evaluate extend coverage for multi addr scenarios.


raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -1,252 +1,57 @@
...
 
-@pytest.mark.parametrize("amount_of_nodes", [1])
-def test_subscribe_to_invalid(comms_nodes):
-    client = comms_nodes[1].client
-    unregistered_address = generate_address()
-
-    # no subscriptions should be present
-    assert client._is_subscribed_to(unregistered_address) is False
-
+@pytest.mark.parametrize("cluster", [{1: 1}, {1: 2}, {1: 2, 2: 2}])
```
JONAF2103: Awesome change

raiden/tests/integration/network/transport/rif_comms/node.py
```diff
@@ -39,26 +39,29 @@ class Node:
...
+            client.connect()
         return process
 
     def stop(self):
         """
         Disconnect from RIF Comms node program, and stop its process.
         """
         try:
-            self.client.disconnect()
+            for client in self.clients:
```
JONAF2103, line 67: Maybe an except could be usefull here and maybe adding a log with the exception, or we are trowing that to the top level?

JONAF2103 (approved):
LGTM good job with the refactor of the tests

raiden/tests/integration/network/transport/rif_comms/node.py
```diff
@@ -39,26 +39,29 @@ class Node:
     """
 
     def __init__(self, config: Config):
-        self.address = config.address
+        self.clients = list()
```
mortelli:
```suggestion
        self.clients = []
```
(imo more pythonic)

raiden/tests/integration/network/transport/rif_comms/node.py
```diff
@@ -39,26 +39,29 @@ class Node:
...
         self._api_endpoint = config.api_endpoint
         self._env_name = config.env_name
-
-        self.client = RIFCommsClient(rsk_address=self.address, grpc_api_endpoint=self._api_endpoint)
+        for i in range(config.amount_of_clients):
```
mortelli:
```suggestion
        for _ in range(config.amount_of_clients):
```

raiden/tests/integration/network/transport/rif_comms/node.py
mortelli: we absolutely need to keep the `finally` statement in order to stop the processes at the os level—this should **never** be skipped.

we can expand the `try-finally` statements into a `try-except-finally` block if you guys think it's worth it, though. although i wonder what we'd do with these exceptions, should they be raised.

come to think of it, i'd prefer these exceptions make the tests fail.

i'm open to hearing arguments against this

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -1,252 +1,152 @@
...
+
+    {
+        1: 3 // a node with 3 clients connected
+        2: 1 // a node with a client
+        3: 2 // another node with 2 clients connected
+    }
+
+    Keys must be in increasing order from 1 to N
+    Values are the amount of clients connected to that node.
+    """
```
mortelli: so this is a good start, but i've been thinking about these cluster keys.

currently we are requiring that "keys must be in increasing order from 1 to N". i think it is a tad confusing to have these cluster keys represent "comms node #", while the cluster values represent "amount of clients".

therefore, i would suggest to change the keys type to `str`, like so:

```python
{
    "comms_node_1": 3,  # a comms node with 3 clients connected to it
    "comms_node_2": 1,  # a comms node with a single client connected to it
    "comms_node_3": 2,  # another comms node with 2 clients connected to it
}
```

maybe we don't even need the `comms_node` part in the name, but i would insist on changing the keys to the `str` type to avoid confusion with the values

mortelli: also, the cluster concept is definitely strong enough for it to be a class, even if it is just internal to these integration tests.

these can also be passed in `parametrize` call afaik

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -1,252 +1,152 @@
...
+        @returns a dict of clients. Each key is the client number to access
+        on each test. The value is a RIFCommsClient instance.
+        """
+        cluster_clients = {}
+        for comms_node in comms_nodes:
+            for client in comms_node.clients:
+                cluster_clients[len(cluster_clients.keys()) + 1] = client
+        return cluster_clients
+
+    nodes = list()
```
mortelli: i think `nodes` should be the result of a specific function, probably one defined for the `Cluster` class.

might i suggest:
```python
nodes = []
for cluster_number, amount_of_clients in sorted(cluster.items()):
    nodes.append(CommsNode(CommsConfig(cluster_number, amount_of_clients)))
```

the dict should probably be iterated in order, to be safe

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -1,252 +1,152 @@
...
-@pytest.mark.parametrize("amount_of_nodes", [1])
-def test_subscribe_to_self(comms_nodes):
-    client, address = comms_nodes[1].client, comms_nodes[1].address
+@pytest.mark.parametrize("cluster", [{1: 1}, {1: 2}, {1: 2, 2: 2}])
+def test_subscribe_to_invalid(comms_clients):
+    for client in comms_clients.values():
+        # subscribe to unregistered address
+        topic_id, _ = client.subscribe_to(generate_address())
+        # FIXME: this should probably throw an exception rather than an empty result
+        assert topic_id is ""
```
mortelli: this is one of the tests for which we don't really need a lot of values in the `parametrize` deco. 

1 client on 1 node should be enough, since the created clients on whatever topology you supply the test with won't even interact with one another.

we have to think about whether it makes sense to—for example—attempt to subscribe to a random address on 4 separate rif comms clients, connected to 2 rif comms nodes that the test has booted up (this takes time!).

i guess we could leave it just to be safe (we shouldn't really assume behavior about the thing that we are testing) but, as an example, 3 different topologies seems like an excessive waste of resources when running this test in particular

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -1,252 +1,152 @@
...
+        1: 3 // a node with 3 clients connected
+        2: 1 // a node with a client
+        3: 2 // another node with 2 clients connected
+    }
+
+    Keys must be in increasing order from 1 to N
+    Values are the amount of clients connected to that node.
+    """
+
+    def get_all_clients_from_cluster_nodes(comms_nodes: list) -> dict:
```
mortelli: this should also be a `Cluster` method.

i'd also rename it to something simpler like `clients_from_cluster` or `cluster_clients`

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -1,252 +1,152 @@
...
-        node = CommsNode(CommsConfig(i))
-        nodes[i] = node
-
-    yield nodes
+    for cluster_key in cluster:
+        amount_of_clients = cluster[cluster_key]
+        node = CommsNode(CommsConfig(cluster_key, amount_of_clients))
+        nodes.append(node)
+    all_clients = get_all_clients_from_cluster_nodes(nodes)
+    yield all_clients
```
mortelli:
```suggestion
    yield get_all_clients_from_cluster_nodes(nodes)
```

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -1,252 +1,152 @@
...
+        assert topic_id is ""
 
-    # no subscription should be present
-    assert client._is_subscribed_to(address) is False
 
-    # subscribe to self and check subscription
-    topic_id, _ = client.subscribe_to(address)
-    assert topic_id
-    assert client._is_subscribed_to(address) is True
+@pytest.mark.parametrize("cluster", [{1: 1}, {1: 2}, {1: 2, 2: 2}])
```
mortelli: this also seems like a "1 client is enough"-kind of test.

i'd go with `[{1:1}]`

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -1,252 +1,152 @@
...
+        topic_id, _ = client.subscribe_to(client.rsk_address.address)
+        assert topic_id
 
 
-@pytest.mark.parametrize("amount_of_nodes", [2])
-def test_subscribe_to_peers(comms_nodes):
-    comms_node_1, comms_node_2 = comms_nodes[1], comms_nodes[2]
-    client_1, address_1 = comms_node_1.client, comms_node_1.address
-    client_2, address_2 = comms_node_2.client, comms_node_2.address
+@pytest.mark.parametrize("cluster", [{1: 1, 2: 1}, {1: 2}])
```
mortelli: this one, on the other hand, is an excellent use of the `parametrize` mark :+1:

raiden/tests/integration/network/transport/rif_comms/node.py
```diff
@@ -15,7 +15,7 @@ class Config:
     api_endpoint_prefix = "localhost"
     env_file_prefix = "testing_"
 
-    def __init__(self, node_number: int):
+    def __init__(self, node_number: int, amount_of_clients: int = 1):
```
mortelli: do we really need this default value? if so, why?

mortelli (commented):
i see that this PR is _already_ paying off, since the tests here have raised a failure for a specific LC topology. 

i'm very happy about this :nerd_face: 

i've left some comments as a sort of preliminary review, but my main recommendations here are:
1. avoid excessive parametrization in tests which are really about just 1 client/node. we don't need to iterate over `n` clients in different topologies and repeat the same action for each of them in _all_ tests
2. pull from the `update-rif-comms` branch. this is a tricky one. although the code in #191 is done, i think this branch is some commits behind in terms of tests. i'm partly responsible for this, so i'm happy to pair up with you to merge the changes

raiden/tests/integration/network/transport/rif_comms/test_client.py
marcosmartinez7: not sure about this, subscribing to self assertions must evaluate to the right boolean value even if there are N clients connected

raiden/tests/integration/network/transport/rif_comms/test_client.py
marcosmartinez7: i moved the unregistered address to calculate it only once, therefore each test for each client should work

is not really necessary but this tests that client1 subscribes to a random address and the assertions pass, the next client2 will get the same results (idempotent)

raiden/tests/integration/network/transport/rif_comms/node.py
marcosmartinez7: i agree that if the disconnect fails on the client therefore test must fail, so i would keep it like that

raiden/tests/integration/network/transport/rif_comms/test_client.py
marcosmartinez7: added cluster class, not sure about keys.

raiden/tests/integration/network/transport/rif_comms/test_client.py
marcosmartinez7: `nodes` are now Cluster `self.nodes`

didnt understand the suggested function, cluster_number doesnt make any sense to me, there is only one cluster.

raiden/tests/integration/network/transport/rif_comms/test_client.py
marcosmartinez7: done

raiden/tests/integration/network/transport/rif_comms/test_client.py
marcosmartinez7: done

raiden/tests/integration/network/transport/rif_comms/node.py
marcosmartinez7: i think that the default case is 1 client per node

a better option would be to validate amount_of_clients > 0

transport/rif_comms/cluster.py
```diff
@@ -0,0 +1,27 @@
+from raiden.tests.integration.network.transport.rif_comms.node import Node, Config
```
mortelli: this module does not belong here, it should reside in `raiden/tests/integration/network/transport/rif_comms`.

it can also be merged with the `node.py` module in that directory

transport/rif_comms/cluster.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class Cluster:
```
mortelli: we need a short docstring here

transport/rif_comms/cluster.py
```diff
@@ -0,0 +1,27 @@
...
+    def __init__(self, nodes_to_clients: dict):
+        self.nodes_to_clients = nodes_to_clients
+        self.nodes = list()
```
mortelli:
```suggestion
        self.nodes = []
```

transport/rif_comms/cluster.py
```diff
@@ -0,0 +1,27 @@
+from raiden.tests.integration.network.transport.rif_comms.node import Node, Config
+
+
+class Cluster:
+    def __init__(self, nodes_to_clients: dict):
+        self.nodes_to_clients = nodes_to_clients
```
mortelli: for me—just a personal opinion—the phrase "nodes to clients" suggest a function or method.

i would name this `node_clients` or `clients_topology` instead and explain this attribute through the code documentation.

not a blocker

transport/rif_comms/cluster.py
```diff
@@ -0,0 +1,27 @@
...
+        self.nodes = list()
+        for cluster_key in nodes_to_clients:
```
mortelli: the concept of "cluster key" is not clear by reading the code, please rename this to `node_number` or something clearer

raiden/tests/integration/network/transport/rif_comms/test_client.py
mortelli: > added cluster class, not sure about keys.

could you please give a more in-depth reply to the keys issue?

raiden/tests/integration/network/transport/rif_comms/test_client.py
mortelli: > `nodes` are now Cluster `self.nodes`
> 
> didnt understand the suggested function, cluster_number doesnt make any sense to me, there is only one cluster.

ok, then:

```python
nodes = []
for node_number, amount_of_clients in sorted(cluster.items()):
    self.nodes.append(CommsNode(CommsConfig(node_number, amount_of_clients)))
```

transport/rif_comms/cluster.py
```diff
@@ -0,0 +1,27 @@
...
+        @returns a dict of clients. Each key is the client number to access
+        on each test. The value is a RIFCommsClient instance.
+        """
+        cluster_clients = {}
+        for comms_node in self.nodes:
+            for client in comms_node.clients:
+                cluster_clients[len(cluster_clients.keys()) + 1] = client
+        return cluster_clients
+
+    def shutdown(self):
```
mortelli: is there any particular reason why this is called `shutdown` instead of `stop`?

i don't mind either choice, i just don't see reason to name the `Node` method something different than the `Cluster` method, it's unnecessary indirection

raiden/tests/integration/network/transport/rif_comms/test_client.py
mortelli: > i moved the unregistered address to calculate it only once, therefore each test for each client should work
> 
> is not really necessary but this tests that client1 subscribes to a random address and the assertions pass, the next client2 will get the same results (idempotent)

that's good, i agree on that change. i would test this directly with `{1: 2}` then, but it's also fine as it is.

raiden/tests/integration/network/transport/rif_comms/test_client.py
mortelli: > not sure about this, subscribing to self assertions must evaluate to the right boolean value even if there are N clients connected

that's true. however, `{1: 2, 2: 2}` contains `{1: 2}`, so (for starters) we could shorten the `parametrize` to  `[{1: 1}, {1: 2, 2: 2}]`.

however, i don't see why we need 2 different comms nodes here if they don't interact with each other.

i would change this to just `[{1: 1}, {1: 2}]`

raiden/tests/integration/network/transport/rif_comms/node.py
mortelli: > i think that the default case is 1 client per node
> 
> a better option would be to validate amount_of_clients > 0

i'm sorry, but i don't understand; who is going to call `Config.__init__` without this parameter, and why would we take a defensive approach in such a case?

raiden/tests/integration/network/transport/rif_comms/test_client.py
marcosmartinez7: "1" doesnt make sense to me, 1 is an integer. 

maybe "A", "B", "C" is better

raiden/tests/integration/network/transport/rif_comms/node.py
marcosmartinez7: i tought that maybe some tests that dindt use the fixture can create a CommsNode providing a in-place Config object

transport/rif_comms/cluster.py
marcosmartinez7: wow 🤦🏻

raiden/tests/integration/network/transport/rif_comms/node.py
mortelli: > i tought that maybe some tests that dindt use the fixture can create a CommsNode providing a in-place Config object

correct, but none of these omit the `amount_of_clients` parameter. 

we should also not worry ourselves with arming our own code with tools against its incorrect use; we don't need to do that as long as this is not a public library or something of that sort

raiden/tests/integration/network/transport/rif_comms/test_client.py
mortelli: > "1" doesnt make sense to me, 1 is an integer.
> 
> maybe "A", "B", "C" is better

exactly. why are we using an integer in the place of a name instead of a quantity?

as values, integers make sense: we are going to create as many clients as the integer value dictates.

as for the cluster keys: they are merely names. `2` is not going to be used as an integer value under any circumstance in the current implementation. therefore, `"2"` is a more apt value, as it represents an (ordered) name.

as long as it is a string, i don't mind the value. if `A`, `B`, `C`, etc. is acceptable, then let's go with that.

fixtures would look something like `[{"A": 2, "B": 1}]`

raiden/tests/integration/network/transport/rif_comms/node.py
```diff
@@ -53,15 +53,19 @@ def start(self) -> Popen:
         Start a RIF Comms node process and connect to it.
         """
         process = CommsProcess.start(env_name=self._env_name)
-        if self._auto_connect:
-            self.client.connect()
+        for client in self.clients:
+            if self._auto_connect:
+                client.connect()
```
mortelli:
```suggestion
        if self._auto_connect:
            for client in self.clients:
                client.connect()
```

raiden/tests/integration/network/transport/rif_comms/node.py
```diff
@@ -15,7 +15,7 @@ class Config:
     api_endpoint_prefix = "localhost"
     env_file_prefix = "testing_"
 
-    def __init__(self, node_number: int, auto_connect=True):
+    def __init__(self, node_number: int, amount_of_clients: int = 1, auto_connect = True):
```
mortelli:
```suggestion
    def __init__(self, node_number: int, amount_of_clients: int, auto_connect=True):
```

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -1,36 +1,47 @@
...
 
 
 @pytest.fixture()
-@pytest.mark.parametrize("amount_of_nodes")
-def comms_nodes(amount_of_nodes) -> {int, CommsNode}:
-    nodes = {}
+@pytest.mark.parametrize("nodes_to_clients")
+def comms_clients(nodes_to_clients: dict) -> Dict[int, Client]:
+    """
+    The cluster dict has the following structure:
```
mortelli:
```suggestion
    The `nodes_to_clients` dict has the following structure:
```

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -1,36 +1,47 @@
...
 @pytest.mark.xfail(reason="wrong exception message from comms node")
 def test_connect():
     # the comms_nodes fixture is not used to prevent automatic connect
     nodes = []
     try:
-        node = CommsNode(CommsConfig(node_number=1, auto_connect=False))
+        node = CommsNode(CommsConfig(node_number=1, amount_of_clients=1, auto_connect=False))
         nodes.append(node)
-        client, address = node.client, node.address
+        client, address = node.clients[0], node.clients[0].rsk_address.address
```
mortelli, line 30: since we went through all the trouble of starting indexes at `1` for every other collection in these tests, it would be nice for `clients` to also be like this.

it's a NTH, not a blocker though

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -90,223 +101,207 @@ def test_connect_peers():
...
-    assert client._is_subscribed_to(unregistered_address) is False
-
-    # subscribe to unregistered address
-    topic_id, _ = client.subscribe_to(unregistered_address)
-
-    # check subscription again
-    # FIXME: this should probably throw an exception rather than an empty result
-    assert topic_id is ""
-    assert client._is_subscribed_to(unregistered_address) is False
+def test_send_message_shutdown():
```
mortelli: please keep the original order for the tests (e.g.: message-sending tests should be grouped together instead of spread out over the module)

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -90,223 +101,207 @@ def test_connect_peers():
...
+    topic_id_1, _ = client_1.subscribe_to(client_2.rsk_address.address)
     assert topic_id_1
-    assert client_1._is_subscribed_to(address_1) is False
-    assert client_1._is_subscribed_to(address_2) is True
-    assert client_2._is_subscribed_to(address_1) is False
-    assert client_2._is_subscribed_to(address_2) is False
+    assert client_1._is_subscribed_to(client_1.rsk_address.address) is False
+    assert client_1._is_subscribed_to(client_2.rsk_address.address) is True
+    assert client_2._is_subscribed_to(client_1.rsk_address.address) is False
+    assert client_2._is_subscribed_to(client_2.rsk_address.address) is False # FIXME this fails
```
mortelli: [RLP-977](https://jirainfuy.atlassian.net/browse/RLP-977)

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -90,223 +101,207 @@ def test_connect_peers():
...
-@pytest.mark.skip(reason="hangs, incomplete")
-def test_send_message_invalid(comms_nodes):
-    client, address = comms_nodes[1].client, comms_nodes[1].address
-
-    # send message to unregistered peer
-    # FIXME: hangs
-    client.send_message("echo", generate_address())
+@pytest.mark.xfail(reason="fails, last assertion fails, multi addr issue")
+@pytest.mark.parametrize("nodes_to_clients", [{1: 1, 2: 1}, {1: 2}])
+def test_is_subscribed_to_self_mutli_address(comms_clients):
```
mortelli:
```suggestion
def test_is_subscribed_to_self_multi_address(comms_clients):
```

mortelli: this one and `test_subscribe_to_self` should be the same test

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -317,31 +312,66 @@ def test_unsubscribe_from_invalid(comms_nodes):
...
     # subscribe to self, then unsubscribe from self
-    client.subscribe_to(address)
-    client.unsubscribe_from(address)
+    client_1.subscribe_to(client_1.rsk_address.address)
+    assert client_1._is_subscribed_to(client_1.rsk_address.address) is True
+    assert client_1._is_subscribed_to(client_2.rsk_address.address) is False
+
+    client_2.subscribe_to(client_2.rsk_address.address)
+    assert client_2._is_subscribed_to(client_2.rsk_address.address) is True
+    assert client_2._is_subscribed_to(client_1.rsk_address.address) is False  # FIXME this fails scenario cluster 2
```
mortelli: [RLP-977](https://jirainfuy.atlassian.net/browse/RLP-977)

mortelli: since this is a `self`-kind of test, clients should be iterated through using a `for` sentence

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -90,223 +101,207 @@ def test_connect_peers():
...
+        assert received_message == payload_1
+        break  # only 1 message is expected
 
 
-@pytest.mark.parametrize("amount_of_nodes", [1])
+@pytest.mark.parametrize("nodes_to_clients", [{1: 1}])
 @pytest.mark.xfail(reason="exceptions are not raised from comms node")
-def test_unsubscribe_from_invalid(comms_nodes):
-    client, address = comms_nodes[1].client, comms_nodes[1].address
+def test_unsubscribe_from_invalid(comms_clients):
```
mortelli, line 296: i think we should be consistent with the strategy adopted in `test_subscribe_to_invalid` here

raiden/tests/integration/network/transport/rif_comms/test_client.py
marcosmartinez7: done

raiden/tests/integration/network/transport/rif_comms/node.py
marcosmartinez7: done

transport/rif_comms/cluster.py
marcosmartinez7: done

transport/rif_comms/cluster.py
marcosmartinez7: done

transport/rif_comms/cluster.py
marcosmartinez7: done

transport/rif_comms/cluster.py
marcosmartinez7: nodes_to_clients represents the mapping for me, but node_clients too

changed.

transport/rif_comms/cluster.py
marcosmartinez7: changed to `node_id`

transport/rif_comms/cluster.py
marcosmartinez7: agree, changed

raiden/tests/integration/network/transport/rif_comms/node.py
marcosmartinez7: nice catch, done

raiden/tests/integration/network/transport/rif_comms/node.py
marcosmartinez7: done

raiden/tests/integration/network/transport/rif_comms/test_client.py
marcosmartinez7: done

raiden/tests/integration/network/transport/rif_comms/test_client.py
marcosmartinez7: we dont need to access trought index now since we iterate over the clients and the cluster just gives us a flattened list

this test case doesnt use the fixture, so i dont find useful to change the way is access to the first element of a list

raiden/tests/integration/network/transport/rif_comms/test_client.py
marcosmartinez7: done

raiden/tests/integration/network/transport/rif_comms/test_client.py
marcosmartinez7: done

marcosmartinez7: I have moved the body of `test_is_subscribed_to_self_mutli_address` to `test_subscribe_to_self` and removed this test

raiden/tests/integration/network/transport/rif_comms/test_client.py
marcosmartinez7: I dont agree

each client suscription and unsuscription must not affect the other client 

we could transform this into a for loop, start from one client and check if the rest of the clients subscriptions were unaffected, but there are only 2 clients, so i dont found a for loop useful for now.

raiden/tests/integration/network/transport/rif_comms/test_client.py
marcosmartinez7: can you give me more info?

raiden/tests/integration/network/transport/rif_comms/test_client.py
marcosmartinez7: will be tested with the PR of update https://github.com/rsksmart/lumino/pull/193

raiden/tests/integration/network/transport/rif_comms/test_client.py
mortelli: sure, makes sense

raiden/tests/integration/network/transport/rif_comms/test_client.py
mortelli: apologies, my original comment was meant for an older version of the test.

thanks for your comment, you are correct :+1:

raiden/tests/integration/network/transport/rif_comms/test_client.py
mortelli: > can you give me more info?

sorry, of course.

we are testing subscriptions to invalid addresses on a cluster with 2 nodes. we should analogously test for unsubscriptions from invalid addresses on the same topology too.

transport/rif_comms/cluster.py
marcosmartinez7: node_to_clients seems like a dict key_to_value, the docs are on the cluster class  so i think we can keep it like this

raiden/tests/integration/network/transport/rif_comms/test_client.py
marcosmartinez7: agree, added

transport/rif_comms/cluster.py
mortelli: yeah, we certainly can :+1:

raiden/tests/integration/network/transport/rif_comms/cluster.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class Cluster:
+    """
+        The Cluster class represents a set of topologies in which RIF Comms bootnode can be tested.
+        The topologies includes different scenarios, on each of them, it can be one or more rif comms nodes and clients present.
+        For example:
+        - Topology A) a rif comms node with two rif comms clients connected
+        - Topology B) two rif comms nodes, each of them with one client connected
+    """
```
mortelli:
```suggestion
    """
    The Cluster class represents a topology in which RIF Comms bootnodes can be tested.
    These topologies includes different scenarios; there can be one or more RIF comms nodes and clients present for each of these.
    For example:
    - Topology A: a single RIF comms node with two RIF comms clients connected to it.
    - Topology B: 2 RIF comms nodes, each of them with one client connected to it.
    """
```

raiden/tests/integration/network/transport/rif_comms/cluster.py
```diff
@@ -0,0 +1,46 @@
...
+        The nodes_to_clients dict must respect the following structure:
+        {
+            "A": 3 // a node with 3 clients connected
+            "B": 1 // a node with a client
+            "C": 2 // another node with 2 clients connected
+        }
+
+        Keys must be letters of the alphabet.
+        Values are the amount of clients connected to that node.
+        """
```
mortelli:
```suggestion
        """
        The nodes_to_clients dict must respect the following structure:
        {
            "A": 3, // a node with 3 clients connected
            "B": 1, // a node with a client
            "C": 2, // another node with 2 clients connected
        }

        Keys must be letters of the alphabet, which represent node names.
        Values correspond the amount of clients connected to that node.
        """
```

raiden/tests/integration/network/transport/rif_comms/cluster.py
```diff
@@ -0,0 +1,46 @@
...
+    def get_clients(self) -> dict:
+        """
+        Aux function to flatten comms clients.
+        @returns a dict of clients. Each key is the client number to access
+        on each test. The value is a RIFCommsClient instance.
+        """
+        cluster_clients = {}
+        for comms_node in self.nodes:
+            for client in comms_node.clients:
+                cluster_clients[len(cluster_clients.keys()) + 1] = client
```
mortelli:
```suggestion
                cluster_clients[len(cluster_clients.keys()) + 1] = client  # start from key "1"
```

raiden/tests/integration/network/transport/rif_comms/node.py
```diff
@@ -15,12 +15,12 @@ class Config:
     api_endpoint_prefix = "localhost"
     env_file_prefix = "testing_"
 
-    def __init__(self, node_number: int, auto_connect=True):
+    def __init__(self, node_id: str, amount_of_clients: int, auto_connect=True):
         """
         Load and set a configuration attributes for a RIF Comms node.
         A valid configuration file must exist at RIF_COMMS_PATH/config/testing_<node_number>.json5
```
mortelli:
```suggestion
        A valid configuration file must exist at RIF_COMMS_PATH/config/testing_<node_id>.json5
```

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -90,123 +87,122 @@ def test_connect_peers():
...
-    assert client_1._is_subscribed_to(address_1) is False
-    assert client_1._is_subscribed_to(address_2) is True
-    assert client_2._is_subscribed_to(address_1) is False
-    assert client_2._is_subscribed_to(address_2) is False
+    assert client_1._is_subscribed_to(client_1.rsk_address.address) is False
+    assert client_1._is_subscribed_to(client_2.rsk_address.address) is True
+    assert client_2._is_subscribed_to(client_1.rsk_address.address) is False
+    assert client_2._is_subscribed_to(client_2.rsk_address.address) is False
 
-    # subscribe from node 2 to 1 and check subscriptions
```
mortelli, line 141: this comment should not be removed

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -90,123 +87,122 @@ def test_connect_peers():
...
-    # check subscription again
-    # FIXME: this should probably throw an exception rather than an empty result
-    assert topic_id is ""
-    assert client._is_subscribed_to(unregistered_address) is False
-
-
-@pytest.mark.parametrize("amount_of_nodes", [1])
-def test_subscribe_to_self(comms_nodes):
-    client, address = comms_nodes[1].client, comms_nodes[1].address
+@pytest.mark.parametrize("nodes_to_clients", [{"A": 1, "B": 2}])
```
mortelli: this is not correct, it should be something like:

`[{"A": 1, "B": 1}, {"A": 2}])`

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -90,123 +87,122 @@ def test_connect_peers():
...
-    # FIXME: hangs
+    # FIXME: raises no peers on routing table
     client.send_message("echo", generate_address())
 
 
-@pytest.mark.parametrize("amount_of_nodes", [1])
-def test_send_message_self(comms_nodes):
-    client, address = comms_nodes[1].client, comms_nodes[1].address
+@pytest.mark.parametrize("nodes_to_clients", [{"A": 1}])
+def test_send_message_self(comms_clients):
```
mortelli, line 176: so, when subscribing to/unsubscribing from _self_, we are usually testing 2 clients on the same node to verify this works OK with the multi address implementation.

something similar could be done to `send_message_self`; e.g. clients 1 and 2 send different messages to themselves, but in the test you check that the payload is as expected (and they are not receiving the messages from the other client instead)

mortelli (approved):
superb work :clap: 

it's very nice to finally be able to test LC transport functionality through `pytest` :smiley: 

i've left some corrections and suggestions, but nothing major.

other than these, a couple of general comments:
1. it's kind of verbose to always have `rsk_address.address`. this is something that can be improved in the future.
2. you can also reduce a couple of lines by going from:
```python
client_1 = comms_clients[1]
client_2 = comms_clients[2]
```
to
```python
client_1, client_2 = comms_clients[1], comms_clients[2]
```
without losing readability (in my opinion).

mortelli:
also, these are my `pytest` results:

```shell
(test_env) rafa@rafa-laptop:~/repos/github/rsksmart/lumino.git/raiden/tests/integration/network/transport/rif_comms$ pytest test_client.py --count=3
=========================================================================================== test session starts ============================================================================================
platform linux -- Python 3.7.9, pytest-5.2.0, py-1.10.0, pluggy-0.13.1
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/home/rafa/repos/github/rsksmart/lumino.git/raiden/tests/integration/network/transport/rif_comms/.hypothesis/examples')
rootdir: /home/rafa/repos/github/rsksmart/lumino.git, inifile: setup.cfg
plugins: timeout-1.3.3, select-0.1.2, hypothesis-3.88.3, random-0.2, cov-2.5.1, repeat-0.9.1
timeout: 540.0s
timeout method: signal
timeout func_only: False
collected 75 items                                                                                                                                                                                         

test_client.py ..............................sss..........................................                                                                                                           [100%]

============================================================================================= warnings summary =============================================================================================
/home/rafa/repos/github/rsksmart/lumino.git/raiden/tests/conftest.py:30
  /home/rafa/repos/github/rsksmart/lumino.git/raiden/tests/conftest.py:30: PytestAssertRewriteWarning: Module already imported so cannot be rewritten: raiden.tests.utils.factories
    pytest.register_assert_rewrite("raiden.tests.utils.factories")

-- Docs: https://docs.pytest.org/en/latest/warnings.html
========================================================================== 72 passed, 3 skipped, 1 warnings in 815.63s (0:13:35) ===========================================================================
```

raiden/tests/integration/network/transport/rif_comms/test_client.py
marcosmartinez7: agree, changed

PR #193, opened 2020-12-29
Feature/rlp 969/multiaddr test cases update client

mauricioirace:
Changes:
- Updated RIF Comms' client according to new proto buffer
- Updated RIF Comms' test cases according to new RIF Comms' client


marcosmartinez7 (approved):
lgtm, thanks

PR #196, opened 2020-12-30
RLP-970 Fixing register secret loop error

JONAF2103:
#### Introduction
When we have a low balance on our account and that's insufficient to send a register secret transaction to the chain it fails, when it fails it keeps repeating until the operation can be executed, this loop is caused by a missing condition when we create the event for register secret light. This creates a loop that generates messages for the light client and increases the database size indefinitely.

#### Fix
Adding the condition solves the problem, now we have a correct condition when we create the event for this.

### DIFF
--- a/raiden/transfer/mediated_transfer/target.py
+++ b/raiden/transfer/mediated_transfer/target.py
@@ -612,7 +612,9 @@ def handle_block(
         lock_expiration_threshold=channel.get_receiver_expiration_threshold(lock),
     )
 
-    if lock_has_expired and target_state.state != "expired":
+    target_state_expired = target_state.state == TargetTransferState.EXPIRED
+
+    if lock_has_expired and not target_state_expired:
         failed = EventUnlockClaimFailed(
             identifier=transfer.payment_identifier,
             secrethash=transfer.lock.secrethash,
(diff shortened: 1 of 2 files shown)


marcosmartinez7 (approved):
lgtm

i think this pr must include to run the baseline

raiden/transfer/mediated_transfer/target.py
```diff
@@ -612,15 +612,17 @@ def handle_block(
...
+    if lock_has_expired and not target_state_expired:
         failed = EventUnlockClaimFailed(
             identifier=transfer.payment_identifier,
             secrethash=transfer.lock.secrethash,
             reason=f"lock expired",
         )
         target_state.state = TargetTransferState.EXPIRED
         events = [failed]
-    elif secret_known:
+    elif secret_known and not target_state_expired:
```
mortelli, line 625:
```
if A && !B:
    do_something()
elif C && !B:
    do_something_else()
```
is equivalent to—and in my opinion clearer than—:
```
if !B:
    if A:
        do_something()
    elif C:
        do_something_else()
```

mortelli (approved):

PR #199, opened 2021-01-06
feature/RLP-900/rif-comms-exception-handling

marcosmartinez7:
## Description

This PR aims to resolve the RIF Comms Client exceptions handling, that is: 

- Timeouts on the grpc stub
- Controlled errors raised from the rif comms node

## Changelog 

- Defined exceptions.py module under rif_comms directory (@JONAF2103  @mortelli  maybe we need to rename this module to exception_handler, see the comments of the review)
- Defined a mapping between the expected grpc errors and custom exceptions that the client will raise and must be handled by a transport node
- rif comms transport node now handle the exceptions for send_message
- rif comms node transport node ignore the exceptions for the rest of the methods, therefore the raised exception is bubbled up and the lumino node will exit (connect and subscribe are unrecoverable errors right now for the lumino business logic)
- defined a decorator to wrap the stub invocations and handle those exceptions
- defined a default timeout that is part of the lumino config

## Acceptance criteria 

- [x] unit test should pass
- [x] rif comms client unit tests must be extended in order to cover the new exceptions


raiden/app.py
```diff
@@ -72,9 +73,9 @@ class App:  # pylint: disable=too-few-public-methods
             },
             "rif_comms": {
                 "grpc_endpoint": DEFAULT_RIF_COMMS_GRPC_ENDPOINT,
+                "grpc_client_timeout": DEFAULT_RIF_COMMS_GRPC_CLIENT_TIMEOUT_SECONDS,
```
marcosmartinez7: neat!

raiden/settings.py
```diff
@@ -25,6 +25,7 @@
 }
 
 DEFAULT_RIF_COMMS_GRPC_ENDPOINT = "localhost:5013"
+DEFAULT_RIF_COMMS_GRPC_CLIENT_TIMEOUT_SECONDS = 30
```
marcosmartinez7: I dont know which timeout is ok, for example, for me 10 seconds is enough, for connect, subscribe and try to send a message. If the node is taking more that 10 seconds, i think that there is an issue with the transport layer or the rif comms node itself.

So, i think that we can put the less number possible here

I dont know what do you guys think, we can go for

- 5-10 seconds (optimistic approach) 
- 30 seconds (pesimistic approach)
- exponential backoff (we can reuse timeout_exponential_backoff from udp/utils i think)

tbh im ok with 30 seconds for this release, but i want to know your opinions and just give you guys my thoughts

raiden/ui/cli.py
```diff
@@ -353,6 +354,20 @@ def options(func):
                 show_default=True,
             ),
         ),
+        option_group(
+            "RIF Comms Transport Options",
+            option(
+                "--grpc-client-timeout",
+                help=(
+                    "RIF Comms client timeout for the grpc client.\n"
```
marcosmartinez7:
```suggestion
                    "A floating point number specifying a timeout for RIF Comms Client requests, expressed in seconds.\n"
```

transport/rif_comms/client.py
```diff
@@ -13,38 +13,49 @@
...
 class Client:
+
     """
     Class to connect and operate against a RIF Communications pub-sub node.
     """
 
-    def __init__(self, rsk_address: Address, grpc_api_endpoint: str):
+    def __init__(self, rsk_address: Address,
+                 grpc_api_endpoint: str,
+                 grpc_client_timeout: float = 30):
```
marcosmartinez7: i think this should be `grpc_client_timeout: float = DEFAULT_RIF_COMMS_GRPC_CLIENT_TIMEOUT_SECONDS`

transport/rif_comms/exceptions.py
```diff
@@ -0,0 +1,116 @@
...
+    @property
+    @abstractmethod
+    def get_message(self):
+        """
+        return the code property
+        """
+
+    def __str__(self):
+        message = super(ClientException, self).__str__()
+        return f"Code: {self.get_code}, Message: {message}"
```
marcosmartinez7: not sure if this is ok, super constructor only receives one argument? so message can be retrieved from super.__str__?

transport/rif_comms/exceptions.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+class InvalidArgumentException(ClientException):
+    def __init__(self, code, message):
+        self.code = code
+        self.message = message
+        super(InvalidArgumentException, self).__init__(message)
```
marcosmartinez7: same here, not sure about if allows to pass more than 1 value to the init

transport/rif_comms/exceptions.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def get_message(self):
+        return self.message
+
+
+class InternalException(ClientException):
+    def __init__(self, code, message):
+        self.code = code
+        self.message = message
+        super(InternalException, self).__init__(message)
```
marcosmartinez7: same

transport/rif_comms/exceptions.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def get_message(self):
+        return self.message
+
+
+class NotFoundException(ClientException):
+    def __init__(self, code, message):
+        self.code = code
+        self.message = message
+        super(NotFoundException, self).__init__(message)
```
marcosmartinez7: same

transport/rif_comms/exceptions.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def get_message(self):
+        return self.message
+
+
+class FailedPreconditionException(ClientException):
+    def __init__(self, code, message):
+        self.code = code
+        self.message = message
+        super(FailedPreconditionException, self).__init__(message)
```
marcosmartinez7: same

transport/rif_comms/exceptions.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def get_message(self):
+        return self.message
+
+
+class TimeoutException(ClientException):
+    def __init__(self, code, message):
+        self.code = code
+        self.message = message
+        super(TimeoutException, self).__init__(message)
```
marcosmartinez7: same

transport/rif_comms/exceptions.py
```diff
@@ -0,0 +1,116 @@
...
+EXCEPTION_MAPPING = {
+    StatusCode.DEADLINE_EXCEEDED: TimeoutException,
+    StatusCode.INTERNAL: InternalException,
+    StatusCode.INVALID_ARGUMENT: InvalidArgumentException,
+    StatusCode.NOT_FOUND: NotFoundException,
+    StatusCode.FAILED_PRECONDITION: FailedPreconditionException
+}
+
+
+def client_handled_operation(func):
```
marcosmartinez7: love it

transport/rif_comms/exceptions.py
```diff
@@ -0,0 +1,116 @@
...
+    def inner(client: "Client", *args, **kwargs):
+        try:
+            return func(client, *args, **kwargs)
+        except RpcError as error:
+            raise get_exception(error)
+    return inner
+
+
+def get_exception(rpc_error: RpcError) -> Union[RpcError, ClientException]:
```
marcosmartinez7: if this module (exceptions.py) will 

- define exceptions
- define the mapping grpc status to exception
- define the way to convert them

i think it should be a class named ExceptionHandler or something

or.. the decorator, the exception mapping and mapping function can be moved back to the client

basically i think this file has enough responsability to be consider as a exception handler or converter class

raiden/settings.py
mortelli:
```suggestion
DEFAULT_RIF_COMMS_GRPC_CLIENT_TIMEOUT = 30  # in seconds
```

we have other `TIMEOUT` module variables, but they do not include the unit in their name—a minor squabble for consistency's sake

mortelli: > I dont know which timeout is ok, for example, for me 10 seconds is enough, for connect, subscribe and try to send a message. If the node is taking more that 10 seconds, i think that there is an issue with the transport layer or the rif comms node itself.
> 
> So, i think that we can put the less number possible here
> 
> I dont know what do you guys think, we can go for
> 
>     * 5-10 seconds (optimistic approach)
> 
>     * 30 seconds (pesimistic approach)
> 
>     * exponential backoff (we can reuse timeout_exponential_backoff from udp/utils i think)
> 
> 
> tbh im ok with 30 seconds for this release, but i want to know your opinions and just give you guys my thoughts

i would opt for going by usability here. i don't think waiting 10 seconds for a single message is very usable (unless it is a very particular hiccup) so i would go for 10.

raiden/ui/cli.py
```diff
@@ -22,7 +22,8 @@
     DEFAULT_PATHFINDING_MAX_FEE,
     DEFAULT_PATHFINDING_MAX_PATHS,
     INITIAL_PORT,
-    DEFAULT_RIF_COMMS_GRPC_ENDPOINT)
+    DEFAULT_RIF_COMMS_GRPC_ENDPOINT,
+    DEFAULT_RIF_COMMS_GRPC_CLIENT_TIMEOUT_SECONDS)
```
mortelli:
```suggestion
    DEFAULT_RIF_COMMS_GRPC_CLIENT_TIMEOUT_SECONDS
)
```

raiden/ui/app.py
```diff
@@ -138,6 +138,7 @@ def run_app(
     routing_mode: RoutingMode,
     config: Dict[str, Any],
     grpc_endpoint: str,
+    grpc_client_timeout: float,
```
mortelli, line 141: do we really need this timeout to be a floating number?

1. would someone really care about timing out after, say, `10.8` seconds instead of `11` or `10`?
2. our default value is already an int

simple is better than complex :snake:

raiden/ui/cli.py
mortelli: i would go for `calls` instead of `operations`, but i agree

transport/rif_comms/client.py
mortelli: > i think this should be `grpc_client_timeout: float = DEFAULT_RIF_COMMS_GRPC_CLIENT_TIMEOUT_SECONDS`

yes

transport/rif_comms/client.py
```diff
@@ -13,38 +13,49 @@
...
         """
         Constructs the RIF Communications Client.
         :param rsk_address: RSK address of the node that wants to use the RIF Communications server
         :param grpc_api_endpoint: GRPC URI of the RIF Communications pub-sub node
+        :param grpc_client_timeout: GRPC Client request timeout
         """
         self.rsk_address = RskAddress(address=to_checksum_address(rsk_address))
         self.grpc_channel = insecure_channel(grpc_api_endpoint)  # TODO: how to make this secure?
         self.stub = CommunicationsApiStub(self.grpc_channel)
+        self.grpc_client_timeout = grpc_client_timeout
```
mortelli: i'd move the `grpc_client_timeout` assignment up 1 line, so that both `grpc`-named attributes are together

transport/rif_comms/node.py
```diff
@@ -233,7 +235,11 @@ def send_message(self, payload: str, recipient: Address):
             recipient=pex(recipient)
         )
         # send the message
-        self._comms_client.send_message(payload, recipient)  # TODO: exception handling for RIF Comms client
+        try:
+            self._comms_client.send_message(payload, recipient)
+        except (NotFoundException, InvalidArgumentException) as e:
+            self.log.warn("RIF Comms send message failed", exception=e)
```
mortelli: this has to be consistent with the message on line `244`. we should probably add `success` to that one and keep `failure` on this one

mortelli: also, `payload` and `recipient` should be added here too

transport/rif_comms/client.py
```diff
@@ -53,20 +64,19 @@ def subscribe_to(self, rsk_address: Address) -> (str, Notification):
...
                 subscriber=self.rsk_address
-            )
+            ),
+            timeout=self.grpc_client_timeout
         )
         for response in topic:
             topic_id = response.channelPeerJoined.peerId
-            break
-        return topic_id, topic
+            return topic_id, topic
```
mortelli:
```suggestion
            return topic_id, topic  # only 1 response is expected
```

transport/rif_comms/exceptions.py
```diff
@@ -0,0 +1,116 @@
+from abc import ABC, abstractmethod
+from typing import Union
+
+from grpc import RpcError, StatusCode
+
+
+class ClientException(Exception, ABC):
+
```
mortelli: a docstring is needed here

transport/rif_comms/exceptions.py
mortelli: > not sure if this is ok, super constructor only receives one argument? so message can be retrieved from super.**str**?

yeah, this is the python 2 syntax, which we might not need after all (my bad).

i think:
1. `ClientException.__str__(self)` 
2. `super().__str__()`

for line `24` might also work (not _so_ sure about option 1)

transport/rif_comms/exceptions.py
```diff
@@ -0,0 +1,116 @@
...
+        """
+        return the code property
+        """
+
+    @property
+    @abstractmethod
+    def get_message(self):
+        """
+        return the code property
+        """
```
mortelli: if these are properties, they should not include `get_` in their names

transport/rif_comms/exceptions.py
mortelli: > same here, not sure about if allows to pass more than 1 value to the init

what do you mean by this? which `__init__` are we talking about? (`29` or `32`?)

transport/rif_comms/exceptions.py
```diff
@@ -0,0 +1,116 @@
...
+        return self.message
+
+
+EXCEPTION_MAPPING = {
+    StatusCode.DEADLINE_EXCEEDED: TimeoutException,
+    StatusCode.INTERNAL: InternalException,
+    StatusCode.INVALID_ARGUMENT: InvalidArgumentException,
+    StatusCode.NOT_FOUND: NotFoundException,
+    StatusCode.FAILED_PRECONDITION: FailedPreconditionException
+}
```
mortelli: i thought we had made this a class field, am i wrong? if not, why was this changed?

raiden/settings.py
JONAF2103: I put 30 just because i didn't knew but yeah maybe 10 it's ok or something in the middle like 15

transport/rif_comms/exceptions.py
JONAF2103: About moving back i don't know because all these are exception related, but maybe we can add a class called exception handler yeah or something like that

transport/rif_comms/exceptions.py
mortelli: > love it

me too. please add a comment or docstring to this deco to make it even clearer :smiley:

transport/rif_comms/exceptions.py
mortelli: > if this module (exceptions.py) will
> 
>     * define exceptions
> 
>     * define the mapping grpc status to exception
> 
>     * define the way to convert them
> 
> 
> i think it should be a class named ExceptionHandler or something
> 
> or.. the decorator, the exception mapping and mapping function can be moved back to the client
> 
> basically i think this file has enough responsability to be consider as a exception handler or converter class

i don't think this should be a stand-alone utils function. these are _client_ exceptions, they belong in the client class. just make this a static function and we should be ok.

same goes for the deco

mortelli (changes requested):
great approach to solve this issue :+1: i have one major concern, though.

i don't understand the point of the `ClientException` abstraction as it currently is. all of the classes which inherit this one have _the same_ code when it comes to methods.

if we want to differentiate exceptions based on their class, that's cool. in fact, i think we should. but we don't need to repeat the `__init__`, `get_code` and `get_message` definitions for each type of exception. 

just define each custom exception class, make them inherit from `ClientException`, and we should be done, no? in fact, `ClientException` might not need to be abstract at all.

raiden/ui/app.py
JONAF2103: For testing you will need to pass a float number to the client, for example 0.1 seconds to make it fail by timeout to check if the timeout is ok, i agree on the default values but the client should accept floats as the parameter.

transport/rif_comms/exceptions.py
mortelli: > About moving back i don't know because all these are exception related, but maybe we can add a class called exception handler yeah or something like that

i am also OK with an exception handler class. just have this in _a_ class

transport/rif_comms/exceptions.py
JONAF2103: Because is not needed in the client class, the mapper is used by the decorator now and that is in this file so we don't need the mapping on the other file.

raiden/ui/app.py
marcosmartinez7: 1. i will keep it like it is defined on grpc, a float. I know that nobody will use 10.1.. but our test cases probably will use 0.1 or something to force failures
2. it should be a float then

raiden/settings.py
marcosmartinez7: lets go for 15

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -90,14 +121,61 @@ def test_connect_peers():
...
+                RskSubscription(
+                    topic=invalid_address,
+                    subscriber=invalid_address
+                ),
+                timeout=30
+            )
+            for response in topic:
+                # should fail before this assert
+                assert not response
+        assert "0x123 is not a valid RSK address" == e.value.details()
```
marcosmartinez7:
```suggestion
        assert f"{invalid_address} is not a valid RSK address" == e.value.details()
```

transport/rif_comms/exceptions.py
```diff
@@ -0,0 +1,44 @@
...
+class FailedPreconditionException(ClientException):
+    """
+        A RIF Comms pubsub bootnode precondition failed
+    """
+
+
+class TimeoutException(ClientException):
+    """
+        Request to RIF Comms pubsub bootnode exceeded
+    """
```
mortelli: no need for the indentation in each of the docstrings here

raiden/ui/cli.py
```diff
@@ -22,7 +22,8 @@
     DEFAULT_PATHFINDING_MAX_FEE,
     DEFAULT_PATHFINDING_MAX_PATHS,
     INITIAL_PORT,
-    DEFAULT_RIF_COMMS_GRPC_ENDPOINT)
+    DEFAULT_RIF_COMMS_GRPC_ENDPOINT,
+    DEFAULT_RIF_COMMS_GRPC_CLIENT_TIMEOUT)
```
mortelli:
```suggestion
    DEFAULT_RIF_COMMS_GRPC_CLIENT_TIMEOUT
)
```

raiden/ui/cli.py
```diff
@@ -353,6 +354,20 @@ def options(func):
...
             ),
         ),
+        option_group(
+            "RIF Comms Transport Options",
+            option(
+                "--grpc-client-timeout",
+                help=(
+                    "A floating point number specifying a timeout for RIF Comms Client requests, expressed in seconds.\n"
+                    "Valid values:\n"
+                    "A number of seconds, it can be float values"
```
mortelli:
```suggestion
                    "An amount of seconds; integer or decimals are allowed."
```

transport/rif_comms/client_exception_handler.py
```diff
@@ -0,0 +1,21 @@
+from typing import Union
+
+from grpc import RpcError, StatusCode
+
+from transport.rif_comms.exceptions import ClientException, TimeoutException, InternalException, \
+    InvalidArgumentException, NotFoundException, FailedPreconditionException
+
+
+class ClientExceptionHandler:
```
mortelli, line 9: add docstring:
```python
"""
A class to map RPC exceptions to RIF Comms client exceptions.
"""
```

transport/rif_comms/client_exception_handler.py
```diff
@@ -0,0 +1,21 @@
...
+    EXCEPTION_MAPPING = {
+        StatusCode.DEADLINE_EXCEEDED: TimeoutException,
+        StatusCode.INTERNAL: InternalException,
+        StatusCode.INVALID_ARGUMENT: InvalidArgumentException,
+        StatusCode.NOT_FOUND: NotFoundException,
+        StatusCode.FAILED_PRECONDITION: FailedPreconditionException
+    }
+
+    @classmethod
+    def get_exception(cls, rpc_error: RpcError) -> Union[RpcError, ClientException]:
```
mortelli: i would also rename this to `map_exception`, `convert_exception`, `translate_exception`, etc.

transport/rif_comms/exceptions.py
```diff
@@ -0,0 +1,44 @@
...
+
+class FailedPreconditionException(ClientException):
+    """
+        A RIF Comms pubsub bootnode precondition failed
+    """
+
+
+class TimeoutException(ClientException):
+    """
+        Request to RIF Comms pubsub bootnode exceeded
```
mortelli:
```suggestion
        Request to RIF Comms pubsub bootnode took too long to finish
```

mortelli (commented):
some comments before doing a final review for `test_client.py`

transport/rif_comms/exceptions.py
mortelli: @marcosmartinez7 this was not solved

raiden/ui/app.py
mortelli: we agreed to keep this as a float because it's already going to be used as such in the client, plus it's not such a hindrance on the user to have this option besides the integer

transport/rif_comms/exceptions.py
marcosmartinez7: fixed

transport/rif_comms/client_exception_handler.py
marcosmartinez7: Basically this is get an Exception from an error, so translate doenst look good

map_exception, error_to_exception, convert_error looks better

get_exception(rpc error) seems good to me anyway, the purpose is get an exception instance from an error

transport/rif_comms/exceptions.py
JONAF2103: I think is a matter of taste, i like the docstrings with identation actually, it make it cleaner to the eye for me

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -347,3 +520,66 @@ def test_unsubscribe_from_peers(comms_clients):
...
+        nodes.append(node_1)
+        client_1, address_1 = node_1.clients[0], node_1.clients[0].rsk_address.address
+
+        node_2 = CommsNode(CommsConfig(node_id="B", amount_of_clients=1, auto_connect=False))
+        nodes.append(node_2)
+        client_2, address_2 = node_2.clients[0], node_2.clients[0].rsk_address.address
+
+        tiny_timeout = 1e-100
+
+        client_1.grpc_client_timeout = tiny_timeout
```
marcosmartinez7: i would just put   `client_1.grpc_client_timeout = 1e-100` and remove the `tiny_timeout` var

JONAF2103: True will fix that

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -347,3 +520,66 @@ def test_unsubscribe_from_peers(comms_clients):
...
+
+        # connect timeout
+        with pytest.raises(TimeoutException) as e:
```
marcosmartinez7: generally speaking this function is ok for me, we could improved it creating something like

```
def expect_timeout(client, params, function_to_call):
      with pytest.raises(TimeoutException) as e:
            client.function_to_call(params)

        assert "Deadline Exceeded" == e.value.message
        assert StatusCode.DEADLINE_EXCEEDED == e.value.code
```

in order to not duplicate code but i think is ok since this isnt going to grow a lot

JONAF2103: Like your idea, it shouldn't take to long, let me try it

transport/rif_comms/exceptions.py
marcosmartinez7: done

raiden/tests/integration/network/transport/rif_comms/test_client.py
JONAF2103: Done

transport/rif_comms/exceptions.py
mortelli: > I think is a matter of taste, i like the docstrings with identation actually, it make it cleaner to the eye for me

that's fine. let's just be consistent, the `ClientException` class in this very same module did not follow the same style

transport/rif_comms/client_exception_handler.py
mortelli: > Basically this is get an Exception from an error, so translate doenst look good
> 
> map_exception, error_to_exception, convert_error looks better
> 
> get_exception(rpc error) seems good to me anyway, the purpose is get an exception instance from an error

"get something" is really _so_ undescriptive to me. get it from where? from what, and how? i'm not a fan of the word `get` in almost any method name, to be honest.

it's not a blocker anyways so it's fine :+1:

raiden/tests/conftest.py
```diff
@@ -16,7 +16,6 @@
 from pathlib import Path
```
mortelli, line 16: why are there changes to this file now?

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -1,13 +1,19 @@
...
 from raiden.tests.integration.network.transport.rif_comms.cluster import Cluster
 from raiden.tests.integration.network.transport.rif_comms.node import Node as CommsNode, Config as CommsConfig
 from raiden.tests.integration.network.transport.utils import generate_address
+from raiden.utils import Address
 from transport.rif_comms.client import Client
+from transport.rif_comms.client_exception_handler import ClientExceptionHandler
+from transport.rif_comms.exceptions import NotFoundException, FailedPreconditionException, InvalidArgumentException, \
+    TimeoutException, ClientException
+from transport.rif_comms.proto.api_pb2 import RskAddressPublish, Msg
+from transport.rif_comms.proto.api_pb2 import RskSubscription, RskAddress
```
mortelli:
```suggestion
from transport.rif_comms.proto.api_pb2 import RskAddressPublish, Msg, RskSubscription, RskAddress
```

raiden/tests/integration/network/transport/rif_comms/test_client.py
```diff
@@ -347,3 +520,79 @@ def test_unsubscribe_from_peers(comms_clients):
...
+def test_client_timeouts():
+    nodes = []
+    try:
+        node_1 = CommsNode(CommsConfig(node_id="A", amount_of_clients=1, auto_connect=False))
+        nodes.append(node_1)
+        client_1, address_1 = node_1.clients[0], node_1.clients[0].rsk_address.address
+
+        node_2 = CommsNode(CommsConfig(node_id="B", amount_of_clients=1, auto_connect=False))
+        nodes.append(node_2)
+        client_2, address_2 = node_2.clients[0], node_2.clients[0].rsk_address.address
```
mortelli: do we really need 2 clients here?

if the operations are supposed to be timing out, it doesn't really matter if—for example—the second client's address is in the DHT anyway, no?

(maybe i am missing something... :thinking:)

mortelli (approved):
LGTM, some minor remarks.

please do not merge _yet_:
- i am currently awaiting `pytest` results
- i want to see if i can re-use the `expect_error` function for the rest of the tests that were modified

thanks

raiden/tests/conftest.py
JONAF2103: Just to remove test warnings

raiden/tests/integration/network/transport/rif_comms/test_client.py
JONAF2103: Actually we could remove it yes because that it's an old logic, now it's just to get the receiver address for the send message operation, i will remove that yes.

JONAF2103 (approved):
LGTM

PR #201, opened 2021-01-12
Fixing eth_getLogs timeout

JONAF2103:
### Introduction

When we run lumino in a chain with to many blocks from genesis block we have an issue because lumino try to fetch all the events from genesis to latest block. That causes a timeout and the lumino node fails to start. This PR aims to fix that fetching events from 1 million block to latest instead. This applies only to chains with more than 1 million blocks.


JONAF2103:
We discard this and close it because it doesn't apply

PR #204, opened 2021-01-20
Bug fix/rlp 991/fixing transaction poll bug

JONAF2103:
This PR fixes the bug reported on this ticket https://jirainfuy.atlassian.net/browse/RLP-991

### DIFF
--- a/raiden/network/rpc/client.py
+++ b/raiden/network/rpc/client.py
@@ -816,7 +816,7 @@ class JSONRPCClient:
 
             # if the transaction was added to the pool and then removed
             if transaction is None and last_result is not None:
-                raise Exception("invalid transaction, check gas price")
+                log.error("invalid transaction, check gas price")
 
             # the transaction was added to the pool and mined
             if transaction and transaction["blockNumber"] is not None:
@@ -831,7 +831,7 @@ class JSONRPCClient:
                 if block_number >= confirmation_block:
                     return transaction
 
-            gevent.sleep(1.0)
+            gevent.sleep(10)
 
     def new_filter(
         self,
--- a/raiden/network/rpc/client.py
+++ b/raiden/network/rpc/client.py
@@ -28,7 +28,6 @@ from web3.utils.empty import empty
 from web3.utils.toolz import assoc
 
 from raiden import constants
-from raiden.constants import GENESIS_BLOCK_NUMBER
 from raiden.exceptions import (
     AddressWithoutCode,
     EthNodeCommunicationError,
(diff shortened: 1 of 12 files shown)


mortelli (approved):
1. the sleep is OK.
2. i understand that `transaction is None and last_result is not None` does not _necessarily_ mean we should raise an exception, and that is also OK.
3. with this change applied, how will we deal with the cases in which the previous condition means something _has_ indeed gone wrong?

if the last point is not tackled anywhere, please create a follow-up issue for this.

thanks

raiden/network/rpc/client.py
```diff
@@ -830,7 +830,7 @@ def poll(self, transaction_hash: bytes):
                 if block_number >= confirmation_block:
                     return transaction
 
-            gevent.sleep(1.0)
+            gevent.sleep(10)
```
marcosmartinez7, line 833: This can slow down a bit regtest but we can modify it later

raiden/utils/rsk_gas_price_strategy/rsk_gas_price_strategy.py
```diff
@@ -172,7 +172,7 @@ def time_based_gas_price_strategy(web3, _transaction_params):
         else:
             latest_block_gas_price = int(latest_block_gas_price)
         gas_price = _compute_gas_price(probabilities, probability / 100)
-        max_gas = max(gas_price, latest_block_gas_price)
+        max_gas = max(gas_price, latest_block_gas_price * 1.10)
```
marcosmartinez7, line 180: lets add:

- docstring that explains why 1.10
- a TODO related with this issue https://jirainfuy.atlassian.net/browse/RLP-992

marcosmartinez7 (approved):
Lgmt, i think we need to document the gas price change and add a TODO related with the issue on the conversation

marcosmartinez7:
>     1. the sleep is OK.
> 
>     2. i understand that `transaction is None and last_result is not None` does not _necessarily_ mean we should raise an exception, and that is also OK.
> 
>     3. with this change applied, how will we deal with the cases in which the previous condition means something _has_ indeed gone wrong?
> 
> 
> if the last point is not tackled anywhere, please create a follow-up issue for this.
> 
> thanks

Created an issue for 3 https://jirainfuy.atlassian.net/browse/RLP-993

raiden/network/rpc/client.py
JONAF2103: We could add that to settings or a function that calculates that from the current network or something like that.

PR #205, opened 2021-01-21
[Release 0.1.0] Merge develop into master

marcosmartinez7:
This PR merges develop into master for the release 0.1.0

After this PR is merged a pre-release and tag 0.1.0 will be created


mortelli:
2 things i'd suggest doing here:
1. taking a look at the **new** LGTM alerts. it seems there are a couple of `is` which should be `==` instead, plus 1 suspicious SQL alert.
2. doing a diff between `master` _after_ this merge and `develop` (the squashed merge commit can be reverted if we find something went wrong)

mortelli (dismissed):

PR #207, opened 2021-02-01
RLP-984 Fixing light payments without rbtc

JONAF2103:
This PR fixes an error that we introduce when we created the api safe operation decorator for light operations. For payments there is not need to have rbtc balance since those operations are off chain operations. The fix basically is to change a flag to skip the rbtc balance check.


mortelli (approved):
LGTM, we properly missed this one it seems

marcosmartinez7 (approved):
lgtm

PR #209, opened 2021-02-15
RLP-1013 Adding checksum address fix 

JONAF2103:
When we open channel or we send tokens to a partner if we don't use a checksum address on the partner address or token address fields the web ui crashes with a blank screen. This PR fixes that by adding a forced checksum on those values before executing the operations.


mortelli (approved):
i think the resulting code is possibly redundant (some `if` statements after the checksum conversions probably won't ever evaluate to `True`) but since we are basically hot-fixing this, it's OK by me not to remove code.

JONAF2103:
Closing this PR to re-upload this change again to avoid having the LGTM failing

PR #210, opened 2021-02-17
RLP-1013 Address checksum fix

JONAF2103:
### Introduction

When we open channel or we send tokens to a partner if we don't use a checksum address on the partner address or token address fields the web ui crashes with a blank screen. This PR fixes that by adding a forced checksum on those values before executing the operations.

#### Changes

* Adding checksum address fix for open channel and send tokens on partner and token address.
* Adding some comments
* Removing unrelated files
* Updating the webui with the fixes from lumino-web
* Fixing explorer and notifier scripts and exposing explorer web.


mortelli (approved):
for the record, i would have done the UI changes in a separate PR

PR #212, opened 2021-02-19
RLP-995: update installation guides

mortelli:
This PR solves [RLP-995](https://jirainfuy.atlassian.net/browse/RLP-995), which is about updating the installation guide for MacOS.

Previews for the installation guides can be seen here: [for MacOS](https://github.com/rsksmart/lumino/blob/feature/RLP-995/update-installation-guides/docs/0.1.0/install_macos.md), [for Ubuntu](https://github.com/rsksmart/lumino/blob/feature/RLP-995/update-installation-guides/docs/0.1.0/install_ubuntu.md).

## Important
**The `gnureadline` dependency has been removed**, as it was causing problems on MacOS. This was a constraint [added 2 years ago](https://github.com/rsksmart/lumino/blame/4213519ff65d1bc7f1c15d821a7fad498c18f2b4/constraints.txt#L30) and does not appear to be used _anywhere_ in the code.

Moreover, [its official documentation says](https://pypi.org/project/gnureadline/):

> Stand-alone GNU readline module
>
> First… STOP
> 
> Consider this: do you really need this package in 2019? You typically don’t if
> 
> - you use the Python provided by a standard Linux distribution like Ubuntu Debian, CentOS, etc. (It already uses the proper readline.)
> -  you run Windows (It won’t work! Try pyreadline instead.)
> - you use the Python provided by Homebrew or Fink on macOS (It has real readline already!)
> - you want it for IPython (It switched to prompt_toolkit in version 5.0.)
> - you use a Python distribution like Anaconda or Enthought / Canopy (Again, real readline.)
>
> You might need it if
> 
> - you use Python provided by MacPorts or the system on macOS (Python compiled against libedit.)
> - you use a Python distribution like ActivePython on Linux or macOS (This used to ship without readline.)
> - you want to get the latest bug fixes and features in either the readline library or its Python module (Typically when stuck on older systems.)

## Other details
1. A lot of the changes to the MacOS guide were made in the spirit of having the guides for both OS be as similar as possible.
2. The MacOS guide was tested by @mauricioirace, @JONAF2103 and myself. 
3. Some corrections were also applied to the Ubuntu installation guide, such as the instruction to update `pip`.

### DIFF
--- a/docs/0.1.0/install_macos.md
+++ b/docs/0.1.0/install_macos.md
@@ -8,7 +8,7 @@
 
 ## Recommended
 
-1. An RSK account with an RBTC balance of **at least** 0.001 RBTC, in order to be able to use the system.
+1. An RSK account with an RBTC balance of **at least** 0.001 RBTC, in order to be able to use all of the system features.
 
 ## Install required libraries/software
 
--- a/docs/0.1.0/install_macos.md
+++ b/docs/0.1.0/install_macos.md
@@ -2,128 +2,184 @@
 
 ## Prerequisites
 
-1. Access to a synched RSK node. You can do this in a variety of ways:
-	1. Run your own node on TestNet or MainNet, see [https://github.com/rsksmart/rskj/wiki/Install-RskJ-and-join-the-RSK-Orchid-Mainnet-Beta](https://github.com/rsksmart/rskj/wiki/Install-RskJ-and-join-the-RSK-Orchid-Mainnet-Beta)
-	2. Compile and run a RSK node locally, see [https://github.com/rsksmart/rskj/wiki/Compile-and-run-a-RSK-node-locally](https://github.com/rsksmart/rskj/wiki/Compile-and-run-a-RSK-node-locally)
-2. An RSK account with an RBTC balance NOT lower than 0.001 RBTC
-3. XCode
+1. Access to a synced RSK node. You can do this in a variety of ways:
+	1. Run your own node on Testnet or Mainnet, see [Node (RSKj): Install](https://developers.rsk.co/rsk/node/install/).
+	2. Compile and run a RSK node locally, see [Node (RSKj): Contribute](https://developers.rsk.co/rsk/node/contribute/).
+
+## Recommended
+
+1. An RSK account with an RBTC balance of **at least** 0.001 RBTC, in order to be able to use the system.
 
 ## Install required libraries/software
 
-### Install command line tools
+### 1. Install command line tools
 
-```
+Do this by opening a terminal and executing:
+
+```shell script
 xcode-select --install
 ```
 
-### Install Homebrew
+### 2. Install Homebrew
 
-From a Terminal window run:
+You can install Homebrew by executing:
 
+```shell script
+/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
 ```
-/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
-```
 
-### Install OpenSSL
+### 3. Install OpenSSL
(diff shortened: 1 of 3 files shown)


docs/0.1.0/install_macos.md
```diff
@@ -2,128 +2,184 @@
...
 ## Prerequisites
 
-1. Access to a synched RSK node. You can do this in a variety of ways:
-	1. Run your own node on TestNet or MainNet, see [https://github.com/rsksmart/rskj/wiki/Install-RskJ-and-join-the-RSK-Orchid-Mainnet-Beta](https://github.com/rsksmart/rskj/wiki/Install-RskJ-and-join-the-RSK-Orchid-Mainnet-Beta)
-	2. Compile and run a RSK node locally, see [https://github.com/rsksmart/rskj/wiki/Compile-and-run-a-RSK-node-locally](https://github.com/rsksmart/rskj/wiki/Compile-and-run-a-RSK-node-locally)
-2. An RSK account with an RBTC balance NOT lower than 0.001 RBTC
-3. XCode
+1. Access to a synced RSK node. You can do this in a variety of ways:
+	1. Run your own node on Testnet or Mainnet, see [Node (RSKj): Install](https://developers.rsk.co/rsk/node/install/).
+	2. Compile and run a RSK node locally, see [Node (RSKj): Contribute](https://developers.rsk.co/rsk/node/contribute/).
```
mortelli, line 7: i'd like to add here—as we have in the Ubuntu guide—which is the oldest version of MacOS that will support Lumino and this installation method. unfortunately, i do not know which version this is.

this can be tackled through [RLP-1018](https://jirainfuy.atlassian.net/browse/RLP-1018).

it's a low priority issue.

docs/0.1.0/install_macos.md
```diff
@@ -2,128 +2,184 @@
...
+
+## Recommended
+
+1. An RSK account with an RBTC balance of **at least** 0.001 RBTC, in order to be able to use all of the system features.
```
mortelli, line 11: i've marked this as _Recommended_ since it is not strictly a requirement

docs/0.1.0/install_macos.md
```diff
@@ -2,128 +2,184 @@
...
+```shell script
 brew install libpq
 ```
 
-### Install postgresql
-
-```
+```shell script
 brew install postgresql
 ```
```
mortelli, line 75: relevant for these 2: [RLP-1018](https://jirainfuy.atlassian.net/browse/RLP-1018)

docs/0.1.0/install_macos.md
```diff
@@ -2,128 +2,184 @@
...
-env LDFLAGS="-I/usr/local/opt/openssl/include -L/usr/local/opt/openssl/lib" pip install psycopg2
+
+```shell script
+brew install libtool
 ```
 
+```shell script
+env LDFLAGS="-I/usr/local/opt/openssl/include -L/usr/local/opt/openssl/lib" pip3 install psycopg2
+```
```
mortelli, line 93: relevant for these 3: [RLP-1018](https://jirainfuy.atlassian.net/browse/RLP-1018)

docs/0.1.0/install_macos.md
JONAF2103: i agree, but don't worry to much about the oldest version of MacOS since Apple doesn't support to old MacOS versions, they always have all the macs using the latest so i think we are good with that.

docs/0.1.0/install_macos.md
JONAF2103: I agree, keep in mind that if they want to make payments and open channels and all that they need that, is not a requirement to make the system work but it's a requirement to operate with it

JONAF2103 (approved):
LGTM

PR #217, opened 2021-03-02
RLP-1028: invoice payment w/decimal amount

mortelli:
This PR solves issue [RLP-1028](https://jirainfuy.atlassian.net/browse/RLP-1028), which is about generating invoices with decimal amounts through the Lumino REST API.

I've manually verified generating and paying invoices between 2 local nodes, using both integer and decimal amounts.

### DIFF
--- a/raiden/utils/typing.py
+++ b/raiden/utils/typing.py
@@ -105,7 +105,7 @@ PaymentHashInvoice = NewType("PaymentHashInvoice", T_PaymentHashInvoice)
 T_PaymentAmount = int
 PaymentAmount = NewType("PaymentAmount", T_PaymentAmount)
 
-T_InvoiceAmount = int
+T_InvoiceAmount = float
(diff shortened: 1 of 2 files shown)


raiden/utils/typing.py
```diff
@@ -105,7 +105,7 @@
 T_PaymentAmount = int
 PaymentAmount = NewType("PaymentAmount", T_PaymentAmount)
 
-T_InvoiceAmount = int
+T_InvoiceAmount = float
```
mortelli, line 108: afaik this is _only_ used as a type hint, but it's still good to change it

(thanks @mauricioirace)

JONAF2103 (approved):
LGTM

raiden/utils/typing.py
mauricioirace: (you're welcome!)

mauricioirace (approved):
LGTM!

PR #219, opened 2021-03-04
RLP-1035 Adding missing dependencies on requirements-dev.txt

JONAF2103:
### introduction

In this PR we are fixing an error on the testing dependencies, this is related to:

[RLP-1035](https://jirainfuy.atlassian.net/browse/RLP-1035)
[RLP-714](https://jirainfuy.atlassian.net/browse/RLP-714)

### Changes

Adding these dependencies to the `requirement-dev.txt` file:

```
Pygments==2.7.4
PyHamcrest==2.0.2
Twisted==20.3.0
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -40,3 +40,6 @@ git+git://github.com/infuy/web3.py@rsk-4.8.2#egg=web3
 # RIF Comms protobuffer
 grpcio-tools==1.33.2
 grpcio==1.33.2
+Pygments==2.7.4
+PyHamcrest==2.0.2
+Twisted==20.3.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -40,6 +40,3 @@ git+git://github.com/infuy/web3.py@rsk-4.8.2#egg=web3
 # RIF Comms protobuffer
 grpcio-tools==1.33.2
 grpcio==1.33.2
-Pygments==2.7.4
-PyHamcrest==2.0.2
-Twisted==20.3.0
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -22,3 +22,6 @@ bump2version==0.5.8
 matrix-synapse==1.5.1
 json5
 pytest-repeat
+Pygments==2.7.4
+PyHamcrest==2.0.2
+Twisted==20.3.0


requirements.txt
```diff
@@ -39,4 +39,7 @@ eciespy==0.3.0
 git+git://github.com/infuy/web3.py@rsk-4.8.2#egg=web3
 # RIF Comms protobuffer
 grpcio-tools==1.33.2
-grpcio==1.33.2
\ No newline at end of file
+grpcio==1.33.2
+Pygments==2.7.4
+PyHamcrest==2.0.2
+Twisted==20.3.0
```
mauricioirace: I'd move all the tests related dependencies to requirements-dev (like hamcrest).

mortelli: i agree. we should not unnecessarily bloat the envs for regular clients

JONAF2103: So if someone download the code and try to run the tests it will fail, also if we have automated checks in the future for testing on production releases it will fail too? i mean i don't know we should specify then somewhere that for testing you should use requirements-dev.txt right? because we don't have that

JONAF2103: I'm not ok with that but i will change it then.

mortelli: > So if someone download the code and try to run the tests it will fail, also if we have automated checks in the future for testing on production releases it will fail too? i mean i don't know we should specify then somewhere that for testing you should use requirements-dev.txt right? because we don't have that

actually, this is specified in [the pytest guide](https://github.com/anarancio/lumino-docs/blob/master/testing/pytest/guide.md).

in any case: even if we _did_ add these to `requirements.txt` instead of `requirements-dev.txt`, this hypothetical user would still not be able to run the tests. we need more than just these dependencies: we need `geth`, we need `npm`, etc.

JONAF2103: Yes but that's not in the Readme.md of lumino, it's on lumino docs and that's not accesible from lumino, we should reference that somewhere in the Readme since we are an open source project and we need to know that people outside the project will try this eventually right?

mortelli: > Yes but that's not in the Readme.md of lumino, it's on lumino docs and that's not accesible from lumino, we should reference that somewhere in the Readme since we are an open source project and we need to know that people outside the project will try this eventually right?

this is a separate issue. it's true, but it doesn't mean we should add these dependencies to `requirements.txt`.

JONAF2103: I'm just saying what any outsider will say if they try the readme, that's all, the change was applied anyway

mortelli (approved):
i've verified the tests are successful using these dependencies.

thanks for this

requirements.txt
mortelli: > I'm just saying what any outsider will say if they try the readme, that's all, the change was applied anyway

i agree 100%, i just don't think it's the issue that @mauricioirace originally raised.

thanks for applying the change

PR #221, opened 2021-03-04
RLP-1037: invoice minimum amount

mortelli:
This PR implements [RLP-1037](https://jirainfuy.atlassian.net/browse/RLP-1037), which is about not allowing the generation of an invoice with more precision than a 1p BTC or its equivalent in ETH/token, as defined per [BOLT #11](https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md).

### DIFF
--- a/raiden/billing/invoices/encoder/invoice_encoder.py
+++ b/raiden/billing/invoices/encoder/invoice_encoder.py
@@ -61,9 +61,10 @@ def encode_invoice(addr, privkey):
 
     if addr.amount:
         amount = Decimal(str(addr.amount))
-        # We can only send down to millisatoshi.
-        if amount * 10 ** 12 % 10:
-            raise ValueError("Cannot encode {}: too many decimal places".format(
+        # the minimum amount for an invoice is the equivalent of 1 millisatoshi, in wei
+        # this is done for compatibility reasons with Lightning Invoices
+        if amount < 10000000:
+            raise ValueError("cannot generate invoice, amount {} is too low to comply with BOLT #11".format(
                 addr.amount))
 
         amount = addr.currency + shorten_amount(amount)
--- a/raiden/billing/invoices/encoder/invoice_encoder.py
+++ b/raiden/billing/invoices/encoder/invoice_encoder.py
@@ -61,9 +61,8 @@ def encode_invoice(addr, privkey):
 
     if addr.amount:
         amount = Decimal(str(addr.amount))
-        # the minimum amount for an invoice is the equivalent of 1 millisatoshi, in wei
-        # this is done for compatibility reasons with Lightning Invoices
-        if amount < 10000000:
+        # We can only send down to millisatoshi.
+        if amount * 10 ** 12 % 10:
             raise ValueError("cannot generate invoice, amount {} is too low to comply with BOLT #11".format(
                 addr.amount))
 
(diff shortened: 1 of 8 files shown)


raiden/billing/invoices/encoder/invoice_encoder.py
```diff
@@ -61,9 +61,9 @@ def encode_invoice(addr, privkey):
 
     if addr.amount:
         amount = Decimal(str(addr.amount))
-        # We can only send down to millisatoshi.
-        if amount * 10 ** 12 % 10:
-            raise ValueError("Cannot encode {}: too many decimal places".format(
+        # per BOLT #11: smallest amount allowed is 0.000000000001 units
+        if amount * 10 ** 13 % 10:
+            raise ValueError("cannot generate invoice, amount {} is too low to comply with BOLT #11".format(
```
mortelli: my reasoning for this change:

the RFC specifies that the `p` multiplier is the smallest unit that can be used for the invoice amount.

based on this i assume that `1p` is the lowest amount allowed. this translates to `0.000000000001` BTC for the lightning network—for our case it would be ETH or token units.

therefore, if `amount = 0.000000000001` (i.e. `1p`):

1. `bool(amount * 10 ** 12 % 10) == True`, which means the minimum amount will raise a `ValueError`
2. `bool(amount * 10 ** 13 % 10) == False`, which means the minimum amount will be allowed

moreover, if `amount = 0.0000000000001` (i.e. more precision than `1p`), then: 

```
bool(amount * 10 ** 13 % 10) == True
```

which i believe is the correct behavior.

mortelli:
@JONAF2103 @marcosmartinez7 @mauricioirace 

please help me out in testing the API manually to make sure all scenarios are covered before moving this to QA (specifically: not allowing invalid amounts based on [BOLT #11](https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md))

thanks :pray:

raiden/billing/invoices/encoder/invoice_encoder.py
marcosmartinez7: There is a problem with this: 

```
100000.000000001 *10 ** 13 % 10
4.0

```
I dont think the mod approach will work, or at least the falsy/true evaluation. 

I think we need, assuming BTC/ETH amounts, parse the decimal part and compare against the max presition of bitcoin

In a future, if we use satoshis, milisatoshis, etc, into the encoded amount (like 1000000p) that must be refactored.

marcosmartinez7 (changes requested):

raiden/billing/invoices/encoder/invoice_encoder.py
mortelli: you're right, this case will be a problem as well.

i'll take your suggestion into consideration and propose an alternative

mortelli: @marcosmartinez7 

found and implemented [this solution](https://stackoverflow.com/a/6190291), lmk what you think

JONAF2103 (commented):
LGTM i will test this to approve it later

JONAF2103 (approved):
It works ok for me, approved!

raiden/billing/invoices/encoder/invoice_encoder.py
mauricioirace: @marcosmartinez7 Agree with that approach. 
@mortelli I think it will work

marcosmartinez7 (approved):

mauricioirace (approved):
LGTM! just take into accounts the issues that can happen due to floating point limited precission.

PR #222, opened 2021-03-09
Bugfix/rlp 714/fix tests

JONAF2103:
This PR aims to fix the current failing tests, right now we have at least 20 failing tests, we need to fix them and run a baseline to update lumino docs with that.

[Here](https://github.com/rsksmart/lumino/files/6116959/tests.log) are the test results.


mortelli (commented):
most of this appears to make sense at a first glance, my only doubts are with the `test_restapi.py` tests:

1. why are the `with` statements removed? do they no longer raise an exception? if so, why not?
2. i assume `BAD_REQUEST` might have been changed to `CONFLICT` on the rest API, and this seems fine. however, it's not so clear for me why `NOT_FOUND` now returns an `OK` when using `get` instead of `patch`.

if you could please explain these 2 points i would gladly approve :+1: 

thanks

JONAF2103:
@mortelli i answer your questions here:

1. They don't raise any exceptions anymore since we added the checksum conversion now, so if you send a valid address without the 0x prefix it will work just fine
2. BAD_REQUEST are CONFLICT now since the token address is an unknown token address so when it try to get that it throws a CONFLICT error instead of a BAD_REQUEST. 

About the NOT_FOUND, it was a bad call in the past, i search over all the resources and the call:
`    url_without_prefix = (
        "http://localhost:{port}/api/v1/" "channels/ea674fdde714fd979de3edf0f56aa9716b898ec8"
    ).format(port=rest_api_port_number)`
it's a get call and not a patch call, so the call was wrong, and actually that was the root cause of the error, so i changed this to the right get call and because it find the data it returns 200 instead of 404. 

**Note: the patch call exists but it has other parameters that don't apply for the purpose of that test.**

mortelli (approved):
thanks for the explanation :+1:

mortelli:
oh and also: please remember to upload the new pytest results to `lumino-docs` as this is merged

JONAF2103:
I'm going to upload the baseline tomorrow after i let this running over the night and add that result to here, the ticket and the lumino-docs repo

PR #225, opened 2021-03-18
Fix issue with fails transactions in db

mauricioirace:
This PR fixes an issue in which the backend would perform wrong filters, when requesting the payment history

### DIFF
--- a/raiden/storage/sqlite.py
+++ b/raiden/storage/sqlite.py
@@ -927,7 +927,6 @@ class SQLiteStorage:
                                               event_type,
                                               limit,
                                               offset)
-        print(str(len(entries)))
         result = [
             TimestampedEvent(self.serializer.deserialize(entry[0]), entry[1])
             for entry in entries
@@ -989,8 +988,6 @@ class SQLiteStorage:
                                             to_date,
                                             limit,
                                             offset)
-        print(query[0])
-        print(str(query[1]))
         cursor = self.conn.cursor()
 
         cursor.execute(
--- a/raiden/storage/sqlite.py
+++ b/raiden/storage/sqlite.py
@@ -927,7 +927,7 @@ class SQLiteStorage:
                                               event_type,
                                               limit,
                                               offset)
-
+        print(str(len(entries)))
         result = [
             TimestampedEvent(self.serializer.deserialize(entry[0]), entry[1])
             for entry in entries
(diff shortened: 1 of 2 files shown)


raiden/storage/sqlite.py
```diff
@@ -1019,36 +1017,22 @@ def _get_query_with_values(self,
                 state_events
             WHERE
                 json_extract(state_events.data,
-                        '$._type') IN ({}) {} {} {} {}
+                        '$._type') IN ({}) {} {} {}
```
JONAF2103, line 1020: Are we sure about this removal? i'm just concern about what could happen if the number of parameters differ

mauricioirace: It used to have two params, one for target, and another one for initator. Now they are one composed param (using AND operator), an I find it cleaner, as there are situations where we can have an initiator and no target, and viceversa. 

If you look to the place where the string is formated, a param was also removed (actually, I merged two params into one)

mortelli:
[the issue in question: RLP-1046](https://jirainfuy.atlassian.net/browse/RLP-1046)

raiden/storage/sqlite.py
mortelli: @JONAF2103 check out line `1035` on this file

raiden/storage/sqlite.py
```diff
@@ -1057,23 +1041,27 @@ def _get_query_with_values(self,
...
+            nodes_query_parts.append(cls._get_query_for_node_address("target", target_address, our_address))
+        if initiator_address:
+            nodes_query_parts.append(cls._get_query_for_node_address("initiator", initiator_address, our_address))
+        return " AND " + " AND ".join(nodes_query_parts)
 
-        result = " AND json_extract(state_events.data, '$.{}') = ? "
-
-        if or_contiional:
-            result = " OR json_extract(state_events.data, '$.{}') = ?  " \
-                     " AND json_extract(state_events.data, '$.token_network_identifier') = ?"
```
mortelli, line 1072: i'm not sure why or if the token network ID mattered for this query. do you know? maybe we should reflect on this to make sure we aren't making any mistakes

mortelli (approved):

JONAF2103 (approved):
LGTM

raiden/storage/sqlite.py
mauricioirace: The token network is two differentes places.

The code conditionally used an OR, because initator/target field could be missing, and they had to set the token network again to prevent using another token network with the same partner.

We fixed it in another way (using a null check), so it's no longer needed

PR #230, opened 2021-03-30
RLP-1061 Fixing leave network to be a safe operation

JONAF2103:
This PR fixes the leave network functionallity to be safe, now the node will always handle the errors as a response instead of stop working.


mauricioirace (approved):
LGTM

PR #231, opened 2021-03-30
Fix wrong filter in 'channel by  partneraddress' view

mauricioirace:
This PR should fix an issue which causes to node to hang out when requested to leave a network,


JONAF2103 (approved):
LGTM

PR #236, opened 2021-08-19
PP-69: timezone startup crash

mortelli:
Lumino is crashing with the following error at startup:

``` 
TypeError: Only timezones from the pytz library are supported
```

This PR implements a fix based on adding a new constraint.

[Source](https://github.com/Yelp/elastalert/issues/2968)


JONAF2103 (approved):
I tested this and it works fine, LGTM

